PR #6, opened 2020-04-29
Supervision set implementation

pzelasko:
I renamed `TextSupervisionSet` to simpler `SupervisionSet`, since it can support other tasks than ASR. It made more sense to me to make it more dict-like with the `.get()` method and query particular segments for their attributes, unlike `AudioSet` where you can actually do both... I guess it's not very important at this point but we might want to make it a bit more consistent.


danpovey:
Fantastic!  Will look over this soon.  It's great that you are pushing
forward with this.  I have been a bit distracted by other things.

pzelasko:
Cool! I will open a separate issue for discussion on the dataset thing and paste your comments.

PR #10, opened 2020-04-30
Feature extraction for whole recordings

pzelasko:
I have a rough sketch of how `FeatureSet` would be used and obtained via `FeatureExtractor`. I get a feeling this one will take a bit longer to get right, so before I deep dive into coding, I'd like to get some feedback on that first.


lhotse/features.py
```diff
@@ -0,0 +1,116 @@
...
+
+@dataclass
+class FeatureSegment:
+    """
+    PZ: Now that I've written it, seems redundant with SupervisionSegment... maybe it should be a base class for it
+    """
+    recording_id: str
+    channel: int
+    start: Seconds
+    duration: Seconds
```
danpovey, line 81: I think we also should have the array dimensions in the manifest file.

lhotse/features.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class FeatureExtractor:
+    def with_audio_set(self, audio_set: AudioSet) -> 'FeatureExtractor':
```
danpovey, line 85: I'm not so sure about this FeatureExtractor thing.  It looks to me like something that maybe should be a script, or a collection of scripts, rather than a datatype.

test/test_feature_set.py
```diff
@@ -0,0 +1,81 @@
...
+        FeatureExtractor()
+            .with_audio_set(audio_set)
+            .with_augmentation(some_augmentation)
+            .with_algorithm(method='mfcc', frame_size=0.025, frame_shift=0.01, **other_params)
+            .with_segmentation(segmentation)
+            .extract()
+    )
+
+
+@mark.parametrize(
```
danpovey, line 66: this confuses me (and likely others)... please explain in comment.

danpovey:
@pzelaso I'm not going to try to sketch out a design because you would do it better, but let me give you some typical scenarios:

   - You have a huge dataset, you have already prepared the manifest file for the audio (and there is no segmentation), and you want to dump to disk a FeatureSet for this dataset, possibly with some kind of parallelization.  To avoid the dumped data being too huge we use lilcom to compress the log-mel features to approximately one byte per element.  Probably one file per audio recording.

   - You have a scenario like the above, but the log-mels will fit in memory without compression.

   - You have a scenario like the above, but the log-mels will fit in memory only if you use lilcom compression; the FeatureSet object will decompress them when the user asks for them (lilcom won't be exposed to the user level).

  - You have one of the scenarios above, but in addition you want to add background noise dynamically.   There are two ways to do this: by simply adding up the log-mel energies, and by adding it at the signal level.  Presumably if it's added dynamically at the signal level you'd have to either be doing the audio loading and feature extraction while training, or dump to disk several versions of the dataset.  This might be slightly difficult with the current interfaces as there is no allowance made for having multiple versions of the same data (?).  Or is there?  Perhaps if there are alternatives available in the manifest a random one could be returned? 
  
 - The above would also apply with reverberation, but that would probably have to be done at the signal level.  The same choices apply.

pzelasko:
> let me give you some typical scenarios:

Thanks, that's exactly what I needed! Let me digest and reflect on that, I'll come back with some codes.

>     * It might be useful to have some kind of object that's the same as TextSupervision but with no text.  What I'm thinking is you might want to extend the audio boundaries of the supervision by some margins on the left and right, and then extract those segments.  Having that as a separate object might make things easier.

I was thinking along these lines when I renamed `TextSupervisionSet` to `SupervisionSet` and made the text optional.

test/test_feature_set.py
pzelasko: that's a `pytest` utility, which allows you to use the same test with different arguments (parametrized tests); simple example:

```
@pytest.mark.parametrize(
  ['a', 'b', 'c'],
  [(1, 1, 2), (1, 1, 3)]
)
def test_addition(a, b, c):
  assert a + b == c
```

the first argument to parametrize describes the names of inputs (`['a', 'b', 'c']`, and the second argument is a list of tuples or arguments (`[(1, 1, 2), (1, 1, 3)]`)

the second case would fail, as you might expect it to, so `pytest` allows you to mark it as "expected fail", in short, `xfail`, hence we should instead write `param(1, 1, 3, mark=pytest.mark.xfail)`

lhotse/features.py
pzelasko: It's just a very rough sketch that helped me put my thoughts in code. My personal experience is that it's often convenient when you can have it both ways, with CLI being just a wrapper for the class, because then you can import that in a jupyter notebook or a one-off script and have the same functionality. But I won't insist - let's see how it plays out?

danpovey:
BTW, regarding the feature files: it might make sense to include some kind of string that says what the format is: e.g. lilcom, numpy?  If you think so.

pzelasko:
Yeah definitely, I'd like the manifest to specify "how to read them", so we
can extend to other formats in the future if needed.

wt., 5 maj 2020, 22:22 użytkownik Daniel Povey <notifications@github.com>
napisał:

pzelasko:
While there is still some work to finish in this PR, it is now possible to extract the features using CLI utility like this:

    lhotse make-feats test/fixtures/audio.yaml lilcom_feats
    lhotse make-feats --not-compressed test/fixtures/audio.yaml numpy_feats

The created directory has a `storage` subdirectory with one feature file per recording, and `feature_manifest.yml` which looks like this:

```
feature_extractor:
  fbank_config:
    use_log_fbank: true
  mfcc_config:
    cepstral_lifter: 22.0
  mfcc_fbank_common_config:
    high_freq: 0.0
    low_freq: 20.0
    num_mel_bins: 23
    use_energy: true
    vtln_high: -500.0
    vtln_low: 100.0
    vtln_warp: 1.0
  spectrogram_config:
    dither: 0.0
    energy_floor: 1.0
    frame_length: 25.0
    frame_shift: 10.0
    min_duration: 0.0
    preemphasis_coefficient: 0.97
    raw_energy: true
    remove_dc_offset: false
    round_to_power_of_two: false
    snip_edges: true
    window_type: povey
  type: mfcc
features:
- channel_id: 0
  duration: 0.5
  recording_id: recording-1
  start: 0.0
  storage_path: test_lhotse/storage/6582916e-e4a1-4459-a93b-e30e6ffa70c0.llc
  storage_type: lilcom
- channel_id: 1
  duration: 0.5
  recording_id: recording-1
  start: 0.0
  storage_path: test_lhotse/storage/907dd31f-860e-4d1c-876b-a0aa26678bb2.llc
  storage_type: lilcom
- channel_id: 0
  duration: 1.0
  recording_id: recording-2
  start: 0.0
  storage_path: test_lhotse/storage/204d3a36-da89-4427-a6be-4d0b96d208c0.llc
  storage_type: lilcom
- channel_id: 1
  duration: 1.0
  recording_id: recording-2
  start: 0.0
  storage_path: test_lhotse/storage/e8390a78-862a-4de4-9393-edc6848a4627.llc
  storage_type: lilcom
```

It basically contains the config of the feature extractor that was used to obtain this, as well as a list of extracted features. It is possible to pass just the feature extractor config to this script to override the defaults. Note that it actually contains all the possible parameters (but all of them can be omitted as well) and the type of parametrization actually used is stored in the `feature_extractor.type` field. It supports fbank, mfcc and spectrogram, basically reflecting the full scope of Kaldi compatibility functions in torchaudio.

The next steps that are IMO minimum to mege are adding more tests to cover most supported cases as of now and making the `FeatureSet` return a smaller chunk of features if the best matching "feature segment" it found is larger than the requested one... I think other stuff like adding support for custom segmentation (for now it only works on full recordings) or data augmentation, etc. can be done in future PRs.

pzelasko:
Oh - it works best when installed in virtualenv/conda env (`pip install -e .`), then `lhotse` script is automatically visible in $PATH, selects the right python interpreter, etc.

danpovey:
Fantastic!
Would you mind checking that it works (when run from scratch) with lilcom
1.1, and requiring at least that version in requirements.txt?
I added a small header and versioning to the compressed strings, for more
informative error messages later on.

pzelasko:
Yeah it seems to work alright.

danpovey:
Thanks!

pzelasko:
I started checking the compatibility with Kaldi - there seem to be some differences in values, but not in the patterns of MFCC. I tried to set all the settings to the same values in torchaudio and Kaldi, for now I'm not sure where these differences come from. I will be looking into it further later, but maybe you'll have any ideas in the meanwhile. The spectrograms look like this:

image

Kaldi mfcc.conf:
```
--use-energy=false   # only non-default option.
--sample-frequency=16000
--dither=0
```

Lhotse feature conf:
```
feature_extractor:
  type: mfcc
  fbank_config:
    use_log_fbank: true
  mfcc_config:
    cepstral_lifter: 22.0
    num_ceps: 13
  mfcc_fbank_common_config:
    high_freq: 0.0
    low_freq: 20.0
    num_mel_bins: 23
    use_energy: false
    vtln_high: -500.0
    vtln_low: 100.0
    vtln_warp: 1.0
  spectrogram_config:
    blackman_coeff: 0.42
    dither: 0.0
    energy_floor: 0.0
    frame_length: 25.0
    frame_shift: 10.0
    min_duration: 0.0
    preemphasis_coefficient: 0.97
    raw_energy: true
    remove_dc_offset: true
    round_to_power_of_two: true
    snip_edges: true
    window_type: povey
```

pzelasko:
Actually looking at the values themselves the differences are not large - it's mostly the 0th MFCC bin...

image

pzelasko:
@danpovey I think you added `import numpy` in lilcom's setup.py, which makes it crash in CI during installation when being installed along with numpy. I will make a two-stage install to overcome this, but consider making a fix in lilcom too.

pzelasko:
I think this PR is ready -- it introduces a CLI, supports the case of extracting the features for whole recordings and querying the `FeatureSet` for either everything available or sub-chunks of features.

There are still lots of things to improve - the querying might be slow on large datasets, it doesn't support extraction for just chunks of audio (although I think it's very close), the tests are not ideal, and I haven't looked for the feature differences vs Kaldi yet, but I don't want to hold this PR back too long, and we can build it up as we go.

I think the next steps are building a simple pytorch Dataset and creating an actual recipe for some dataset like librispeech in pure `lhotse`, so that I can start testing things more end-to-end and onboarding users, while working on the other features and improvements.

danpovey:
> @danpovey I think you added `import numpy` in lilcom's setup.py, which makes it crash in CI during installation when being installed along with numpy. I will make a two-stage install to overcome this, but consider making a fix in lilcom too.

The setup.py needs it to get the locations of some headers, and it has:
  `  install_requires=['numpy'],`
.. that's not enough?  I don't understand much about these issues.

danpovey:
Regarding the features:
  - Firstly, I was thinking we'd mostly be using log-mel features.  Of course that's just a configuration change.  Lilcom will interact well with that feature type.
 - Regarding the `mfcc_kaldi`: the value of the 0th MFCC coefficient is not very plausible to me.  Extremely positive means the energies of the mel bins are much greater than 1.  How did you get those exactly?  Now, for compatibility with HTK I believe Kaldi uses the integer values of the 16-bit encoded stream, rather than normalizing, but still, it shouldn't be *that* positive.
   Compatibility with Kaldi features is not a priority for me.  I'd like to make it clear what the default features are and mostly stick to that.   (But probably log-mel).

pzelasko:
I changed the default features to fbank. You might want to review the `*Config` classes in `features.py` to check if the default settings make sense - I copied them from Kaldi `src/feat/*.h` files.

As to the large 0th MFCC coefficient, the plot thickens. When I compared the Kaldi and lhotse (torchaudio) output for the same file, again - the output has the same shape, but Kaldi energies are way larger, see the image:

image

The Kaldi result can be recreated by: `steps/make_fbank.sh --fbank-config librifbank.conf --nj 1 libridata fbank_libri fbank_libri`

where `librifbank.conf` has:
```
--dither=0
--sample-frequency=16000
```

and the `wav.scp` in `libridata` is:
```
rec1 /path/to/lhotse/test/fixtures/libri-1088-134315-0000.wav
```

Then, to analyse in Python, I read it in python like:
```
import kaldi_io
fbank_kaldi = list(kaldi_io.read_mat_scp('/path/to/fbank_libri/raw_fbank_libridata.1.scp'))[0][1]
```

Paranoid check with `copy-feats ark:fbank_libri/raw_fbank_libridata.1.ark ark,t:-` yields the same output:
```
rec1  [
  7.189558 9.037251 10.99918 11.80182 13.34635 14.90742 14.93212 16.49876 16.68227 16.07505 17.88986 17.97858 18.57075 17.19065 16.88927 17.6966 17.28112 17.44429 16.39793 15.73682 15.18271 14.46821 13.25515
  8.695339 10.86374 11.10041 11.28711 12.6426 13.6858 15.10104 14.55601 15.04671 15.78543 17.72975 17.83109 17.31431 16.33819 16.76282 17.25342 17.55506 17.83777 16.99071 15.81893 15.01264 13.81689 13.02168
  8.378333 9.902431 10.18937 10.38635 12.07959 13.96713 15.26997 15.65272 15.58296 15.4958 18.21008 17.68359 17.81999 16.76442 16.32026 17.28751 17.46375 17.26385 16.55256 15.46311 14.84257 13.89831 13.44193
(...)
```

Any ideas? Anyway, I think we can merge this one - let me know once you review.

danpovey:
I understand why, I think.  The additive term of (log(32768)*2) probably gets multiplied by either num-ceps=40, or sqrt(40) depending on the type of sinusoidal transform (symmetric or not).

I think we should always have snip-edges=false.  This simplifies the relationship between the num-samples and the num-frames.

lhotse/features.py
```diff
@@ -0,0 +1,314 @@
...
+    dither: float = 0.0
+    window_type: str = "povey"
+    frame_length: Milliseconds = 25.0
+    frame_shift: Milliseconds = 10.0
+    remove_dc_offset: bool = True
+    round_to_power_of_two: bool = True
+    energy_floor: float = 0.0
+    min_duration: float = 0.0
+    preemphasis_coefficient: float = 0.97
+    blackman_coeff: float = 0.42
```
danpovey, line 35: blackman_coeff makes no diff with povey window.  could remove, maybe.  How many of these are the defaults?  I wonder whether it's helpful to put them all, if we don't expect them to be changed?
BTW I think we should standardize on snip_edges = False, for simplicity of num-frames formula.

lhotse/bin/lhotse
```diff
@@ -0,0 +1,98 @@
...
+                   'Supervision manifest can be used here.')
+@click.option('-a', '--augmentation-manifest', type=click.Path(exists=True, dir_okay=False),
+              help='Optional manifest specifying augmentation transforms that can be applied to recordings.')
+@click.option('-f', '--feature-manifest', type=click.Path(exists=True, dir_okay=False),
+              help='Optional manifest specifying feature extractor configuration.')
+@click.option('--compressed/--not-compressed', default=True, help='Enable/disable lilcom for features compression.')
+@click.option('-t', '--lilcom-tick-power', type=int, default=-8,
+              help='Determines the compression accuracy; '
+                   'the input will be compressed to integer multiples of 2^tick_power')
+@click.option('-j', '--num-jobs', type=int, default=1, help='Number of parallel processes.')
```
danpovey, line 52: Do we have a mechanism for splitting up manifest files into pieces?  That could be another way of splitting it into jobs, in case we want them to be on multiple machines (e.g. a queue).

pzelasko: I'll add it in a follow-up PR as two CLI modes: `lhotse split` and `lhotse combine`

lhotse/features.py
pzelasko: I think a few of them differ from `torchaudio` defaults, but not a lot. My own preference is to have them explicitly defined so that if a default value changes in future torchaudio release and we update, we don't get surprised.

pzelasko:
> I understand why, I think. The additive term of (log(32768)*2) probably gets multiplied by either num-ceps=40, or sqrt(40) depending on the type of sinusoidal transform (symmetric or not).
> 
> I think we should always have snip-edges=false. This simplifies the relationship between the num-samples and the num-frames.

I tried extracting both lhotse and Kaldi feats with snip-edges=false - the values change a bit, but it still doesn't explain the energy differences. I think to get to the bottom of this we might need to go through Kaldi and torchaudio with a debugger and see where the values start diverging... Anyway, do you think it's a blocking issue? If not, I will add an issue here so that we don't forget about this, but we might resolve it later.

I'll also change snip-edges to false by default.

PR #15, opened 2020-05-20
CLI extensions: split, combine, convert-kaldi

### LINKED ISSUE #12: Split and combine manifests
Introduce python functions and CLI modes: `lhotse split` and `lhotse combine`. Similar to Kaldi `split_data_dir.sh` and `combine_data_dirs.sh`.

pzelasko:
Main highlights:
- split manifests (Audio, Supervision and Feature) into equal parts (python method + CLI)
- combine manifests (Audio, Supervision and Feature) (python method + CLI)
- CLI for Kaldi data dir conversion into audio + supervision manifests
- fixed some tests and bugs (TL;DR `pytest.mark.xfail` is for tests which fail because of bugs and are temporarily silenced; testing whether the code throws exception as expected is done with `pytest.raises`)

Resolves #12


lhotse/bin/lhotse
```diff
@@ -4,9 +4,11 @@ Use this script like:
 
 $ lhotse --help
 $ lhotse make-feats --help
-$ lhotse make-feats --compressed audio_manifest.yml mfcc_dir
-
+$ lhotse make-feats --compressed audio_manifest.yml mfcc_dir/
 $ lhotse default-feature-config feat-conf.yml
```
danpovey, line 8: maybe make default-feature-config `write-default-feature-config` ?  Otherwise it looks as if you might be attempting to change a default somehow, like git config.

Issue #17, opened 2020-06-09
Source separation dataset improvements


pzelasko:
- [x] return a mask in `SourceSeparationDataset`
- [x] a variant of LibriMix that uses pre-mixed audio files
- [x] add option for on-the-fly noise augmentation in `SourceSeparationDataset`
- [x] implement the correct method for overlaying spectrogram features
- [x] `max` speech separation mode - pad the shorter source to match the longer one
- [x] `min` speech separation mode - trim the mix to the length of the shorter recording

pzelasko:
The `min` speech separation mode can be achieved with `truncate` introduced in #24

PR #19, opened 2020-06-11
Cuts: dynamic and pre-mixed dataset, mixing multiple cuts, example scripts

pzelasko:
Goals of this PR:
- allow mixing more than two Cuts (for noise augmentation in source separation)
- return masks in source separation dataset
- create manifests for precomputed LibriMix mixes to enable the comparison of feature-domain mixing and time-domain mixing

I'm setting `feature/cut` as the base branch for this, so that GH renders only the actual changes made in this PR. We should merge #16 before going forward with this.

Resolves a few bullet points from #17


danpovey:
I wonder if it might make sense to put recipes in a separate repo, or
multiple repos?  The recipes may tend to get very large eventually.
My plan for next-gen Kaldi is to have a repo for recipes (or perhaps one
repo per recipe) that is separate from the scripts or the dependencies

pzelasko:
Separate recipe repo should be ok. I'm not sure about repo per recipe though, that could become difficult to maintain - imagine a new recipe requires some API change in one of the dependencies, now you have to update all the repos independently. Will probably happen less as the project matures, but at this stage it's still in flux. Also, a single repo makes things easier to discover by the users.

BTW I might have overused the word "recipe" in lhotse - I meant it more like "audio and supervision manifest preparation" (think equivalent of Kaldi local/CORPUS_data_prep.sh). It makes sense to me to have the scripts for that in lhotse, so they can be shared by multiple recipes.

danpovey:
Hm I guess you're right, that manifest creation for datasets might belong in the main repo.

By "recipe" I guess I was thinking about things that depend on those manifests, and that have lots of other dependencies on other software, e.g. speaker separation recipes.
But maybe having a small number of relatively easy-to-run recipes in lhotse itself would be a good thing, for testing etc.

pzelasko:
Yeah. We can actually start with having a small number of recipes here for convenience, and then move them to another repo when their number starts growing.

pzelasko:
That's the kind of outputs we're now getting from the SourceSeparationDataset classes:

image

There are two types: "dynamically mixed" and "pre-mixed" datasets, the first one adds the features on-the-fly, the other one computes features on waves that were already mixed (the latter was shown on the figures). Now that I can dynamically mix more than two tracks together, I will dynamically mix noises too and will compare how differently the features look from both datasets.

danpovey:
Fantastic!

pzelasko:
Some new developments: I have a comparison of time-domain mixed signals and fbank-domain mixed signals, overall the feature matrix has similar shapes, but there are some differences. Especially, it looks like the low-energy bins are over-emphasized when added as Mel-energies. Perhaps it makes sense to revisit the formula for mixing the signals in the feature domain (@danpovey any ideas are welcome, the place I'm talking about is in the `FbankMixer` class in `lhotse/features.py`).

image

danpovey:
@pzelasko where is the core code where the addition is done, i.e. what is the formula?  It is log-sum?

danpovey:
I suspect any difference comes from how you compute the desired SNR and the scaling factor.
Are you specifying the SNR here, or just summing the signals/energies with no scaling?

pzelasko:
Oops. You're right. I used 0dB SNR as a default for mix instead of `None`, so the mix rescaled the signals to have the same total energies (as expected). The fixed version looks like this:

image

It should be interesting to see a model trained on that. I'll add a bit more test coverage and then I'll move on to providing a DataLoader than can construct batches.

pzelasko:
@danpovey I think this is ready for review and merging, I will notify Manuel too (in case he's not following this PR). 

In particular, I've added a bash script and a notebook showing how to prepare the LibriMix recipe (both do exactly the same, just one uses CLI and the other the Python library). The notebook additionally visualizes some samples from the dataset. It's best to open the notebook via [this link](https://github.com/pzelasko/lhotse/blob/c9fe617e50b8b0ddaf990e927583c6c4fe620f27/examples/librimix/librimix.ipynb), as PR diffs do not render it well.

I think this will be ready for real use after one more PR with a DataLoader that can pad these samples.

lhotse/dataset/source_separation.py
```diff
@@ -1,41 +1,158 @@
...
-        mixture_cut = self.mixtures_set.cuts[cut_id].with_cut_set(self.sources_set)
-        source_cuts = [self.sources_set.cuts[id_] for id_ in [mixture_cut.left_cut_id, mixture_cut.right_cut_id]]
+        mixture_cut, source_cuts = self._obtain_mixture(cut_id=cut_id)
+
         mixture = torch.from_numpy(mixture_cut.load_features(root_dir=self.root_dir))
         sources = torch.stack(
             [torch.from_numpy(source_cut.load_features(root_dir=self.root_dir)) for source_cut in source_cuts],
             dim=0
         )
+
```
mpariente, line 63: I'm not completely sure how it works here when there is noise. Would there be a real_mask for the noise as well? And a source idx in the binary mask for it as well?

lhotse/features.py
```diff
@@ -365,59 +364,113 @@ def _process_and_store_recording(
...
+            #    (assumes no spectral leakage...) across time and frequencies
+            #  - achieve the final SNR by determining the signal energy ratio and then scaling
+            #    the signals to satisfy requested SNR
+
+            # Compute the added signal energy before it was padded
+            added_feats_energy = fbank_energy(feats)
+            target_energy = self.reference_energy * (10.0 ** (-snr / 10))
+            gain = target_energy / added_feats_energy
+
+        self.mixed_feats = np.log(np.exp(existing_feats) + gain * np.exp(feats_to_add))
```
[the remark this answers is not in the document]
danpovey: I think this assumes the alignment is random and uncorrelated.  If they were in phase, we'd have to add in the magnitude domain, there would be an extra factor of two/0.5 in there.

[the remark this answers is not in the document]
mpariente: Oh yes, that's the opposite of what I said, you're right !

lhotse/bin/modes/cut.py
```diff
@@ -89,26 +95,43 @@ def random_overlayed(
...
+        output_cut_manifest: Pathlike
+):
+    """
+    Create a CutSet stored in OUTPUT_CUT_MANIFEST by matching the Cuts from CUT_MANIFESTS by their recording IDs
+    and mixing them together.
+    """
+    all_cuts = combine(*[CutSet.from_yaml(path) for path in cut_manifests])
+    recording_id_to_cut = groupby(lambda cut: cut.recording_id, all_cuts)
+    mixed_cuts = (
+        reduce(lambda left_cut, right_cut: left_cut.overlay(right_cut), cuts[1:], cuts[0])
```
danpovey, line 133: Your python can be on the advanced side, making it harder to understand than if it were simple... having trouble with this.

danpovey: .. and I don't understand why you do 
 cuts[1:], cuts[0]

danpovey: Would be nice to clarify that it's a shallow operation.

If I have any concerns about all this, it's that I feel the code might be becoming a little opaque.

Not sure how to fix that though.

lhotse/cut.py
```diff
@@ -36,6 +35,14 @@ class Cut:
     # cut duration. They also might overlap.
     supervisions: List[SupervisionSegment]
```
danpovey, line 36: maybe document how you would get the feature num-frames and dimension, i.e. the shape of the features?
Would be useful to know at this point, for a typical user.

danpovey:
.. will github display the notebooks?
was slightly concerned about the images bulking up the git repo, but its up to you.

jtrmal:
IIRC, github will display notebooks (but in non-interactive way)
y.

examples/librimix/librimix.ipynb
```diff
@@ -0,0 +1,771 @@
...
+      ]
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "plt.matshow(sample['sources'][0].T);"
```
danpovey, line 200: dunno whether there is a way to show them upside down so they will look more familiar?
The .T (transpose?) is not a common notation I've seen; I think .transpose() or .transpose(0,1) if necessary, would be easier to read, no?

lhotse/bin/modes/cut.py
pzelasko: Alright, I'll either add a plenty of comments or re-write a bit.

lhotse/cut.py
pzelasko: Great idea! While writing the example script and notebook, I noticed that the CLI is actually nicer to use at this point than the Python lib, which is quite low-level for now. I have some ideas on how to build easier-to-use higher-level APIs on top of what's there, but I want to add a bit more recipes and task types (e.g. ASR) to get a better understanding first.

examples/librimix/librimix.ipynb
pzelasko: Yeah, I'll flip them. Yes, .T stands for "transpose". I don't mind using the longer variant.

lhotse/dataset/source_separation.py
pzelasko: Hmm, good point. I'm returning the masks as if the noise was not there, which is probably not the best supervision. Would it make sense if I made `real_mask` first dimension equal to the number of sources (+ noise), and changed the `binary_mask` to `source_mask` (not sure about the name)` which uses int values to indicate which bin belongs to which source (e.g. 2 sources - {0, 1} values, 3 sources - {0, 1, 2} values, etc.)? This would also enable this dataset to use source separation for multiple sources.

lhotse/cut.py
pzelasko: I'll actually need to make some changes to the `Features` object address this cleanly. I updated the issue #20 with details, and I'll take care of this soon after this PR.

lhotse/bin/modes/cut.py
pzelasko: Done, I refactored this into a `mix_cuts` function, and inside I explained what this code does.

lhotse/features.py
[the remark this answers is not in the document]
pzelasko: Do you guys think we should expose that assumption as a boolean parameter in the function (assume_aligned: bool = False)?

lhotse/features.py
```diff
@@ -365,59 +364,113 @@ def _process_and_store_recording(
...
+        if snr is not None:
+            # TODO: The SNR stuff here might be tricky. For prototyping assume the following:
+            #  - always dealing with log mel filterbank feats
+            #  - total signal energy can be approximated by summing individual bands energies
+            #    (assumes no spectral leakage...) across time and frequencies
+            #  - achieve the final SNR by determining the signal energy ratio and then scaling
+            #    the signals to satisfy requested SNR
+
+            # Compute the added signal energy before it was padded
+            added_feats_energy = fbank_energy(feats)
```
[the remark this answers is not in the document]
pzelasko: That's an interesting thought. Isn't Parseval's theorem defined in terms of DFT and not STFT? I mean, is there an easy way to reason about the time vs frequency domain energy relationship for a spectrogram as opposed to just DFT?

lhotse/dataset/source_separation.py
pzelasko: Huh, that's funny, but it actually works like that right now, depending on whether you pass the noise cuts into `sources_set` or `nonsources_set`. When you pass it to `sources_set` (e.g. by doing `source_cut_set + noise_cut_set`), the noise will get treated as a regular source, and it will be involved in the computation of real_mask and "binary"_mask. If you pass it to `nonsources_set` (which was how I intended it for usage) then it will not be returned as one of the sources and won't participate in mask computation - basically, in that case, it only serves to load its features and add them to the mixture.

Which usage would you say makes more sense? BTW I thought the code won't work for more than 2 sources, but it seems like it will, so I'll remove the exception that I'm raising in such cases.

pzelasko: Also, as a sanity check, what is a reasonable range of values for `real_mask`? It seems the values I'm getting are quite low (e.g. [-0.002, 0.007]), I'm thinking it's the log that's messing up the mask. We might need to exp these energies first for source separation.

lhotse/features.py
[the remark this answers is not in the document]
danpovey: No, the only scenario where they would be in phase is where you are adding the exact same signal, in which case you could just add something to the energy.
It's normally understood that the phases would be uncorrelated.

lhotse/features.py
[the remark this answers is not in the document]
danpovey: You don't need Parseval's Theorem if you deal with expectations, I think.  Expected energy comes from just summing the energies.

lhotse/dataset/source_separation.py
danpovey: I think normally these kinds of masks would be in the range zero to one.
I'd try to make it as flexible as possible and not bake it into the system deeply.  Different kinds of source separation task may need different types of masks, and we should be leaving this open I think.  When someone builds a speech separation setup from this, they can deal with this.

pzelasko: Sounds good to me

pzelasko: Sounds good to me

mpariente: Yes, these masks should have values between zero and one. They are defined in function of the magnitudes, so yes the log might be making things weird. 
I agree with Dan, different people will want different masks. Maybe having a method to obtain the masks from the mixture, the sources and the noise would make this choice easier when subclassing?

danpovey:
Sure, I think so..
I think the proportion of power can be used, or just zero-ones.

lhotse/features.py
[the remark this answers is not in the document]
mpariente: > That's an interesting thought. Isn't Parseval's theorem defined in terms of DFT and not STFT? I mean, is there an easy way to reason about the time vs frequency domain energy relationship for a spectrogram as opposed to just DFT?

I think you can group waveform slices in the log, apply jensen inequality and then apply Parseval theorem to each DFT frame independently? I'm not 100% sure. 

> You don't need Parseval's Theorem if you deal with expectations, I think. Expected energy comes from just summing the energies.

Ok. But the SNR range on fbank would be a bit different, no? 
Anyway, it might not be a very important remark from me.

lhotse/dataset/source_separation.py
pzelasko: I don't mind, but as Dan suggested in another comment, let's sort this out when building an actual source separation recipe and extend as we go. I changed the mask computation so that we first `exp` the features. You can check out how the masks look here (they're 0 - 1 range now). 

https://github.com/pzelasko/lhotse/blob/f2200fa7fe60fa134d0f256dd947aea3819f50de/examples/librimix/librimix.ipynb

PR #22, opened 2020-06-18
Consistency between duration and num_frames in features; compressed serialization

### LINKED ISSUE #20: Consistency between duration and number of frames
Original comment from Dan:

It would probably make sense to have `duration` always be `n * frame_shift`, but understand that this is kind of approximate and actually it may use slightly more context than that?

I'd rather make it so that downstream, these cuts behave quite simply w.r.t. the relationship of duration and num-frames, and the user doesn't have to worry about the frame-length and how it may impact the num-frames.  That was a rabbit-hole in Kaldi itself.

_Originally posted by @danpovey in https://github.com/pzelasko/lhotse/pull/16_

pzelasko:
@danpovey LMK if this is what you had in mind

Resolves #20


lhotse/features.py
```diff
@@ -198,27 +209,37 @@ def __post_init__(self):
...
+
+        :param recording_id: str, requested recording ID.
+        :param channel_id: int, requested channel.
+        :param start: float, requested start time in seconds for the feature chunk.
+        :param duration: optional float, requested duration in seconds for the feature chunk.
+            By default, return everything from the start.
+        :param leeway: float, controls how strictly we have to match the requested start and duration criteria.
+            It is necessary to keep a small positive value here (default 0.05s), as there might be differneces between
+            the duration of recording/supervision segment, and the duration of features. The latter one is constrained
+            to be a multiply of frame_shift, while the former can be arbitrary.
```
danpovey, line 241: multiple, not multiply

test/fixtures/dummy_feats/feature_manifest.yml
```diff
@@ -15,46 +15,49 @@ feature_extractor:
   spectrogram_config:
     dither: 0.0
     energy_floor: 0.0
-    frame_length: 25.0
-    frame_shift: 10.0
+    frame_length: 0.025
```
danpovey, line 18: can we comment that these will be converted to milliseonds before Kaldi sees them?  not sure if yaml has comments?

danpovey:
Great!

PR #24, opened 2020-06-21
New CLI modes: cut truncate, append; yaml filter; updated LibriMix example

pzelasko:
Highlights:
- `.truncate()` method for `MixedCut` that truncates the underlying `Cut`s and figures out the offsets and durations for each of them;
- `.truncate()` method for `CutSet` which batch truncates the cuts inside with different strategies
- new CLI modes for cut: `truncate` and `append`
- new CLI mode `lhotse yaml filter` that allows generic filtering of manifests based on their item attributes, e.g. `lhotse yaml filter 'duration>4.0' cut.yml cut_4s.yml` will select only the cuts that have duration greater than 4s
- updated LibriMix example to reflect some of these changes; now all cuts are of the same length and can be batched together

 I think it's time to move on to building an ASR example where we'll build the cuts from long recordings, e.g. conversations, which will allow me to test some of these capabilities better. At the same time, I think this should be enough to try building a recipe in Asteroid with Lhotse.


examples/librimix/librimix.ipynb
```diff
@@ -124,10 +125,43 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "clean_cut_set = mix_by_recording_id(cut_sets=[librimix_manifests['sources']['cuts']])\n",
```
danpovey, line 127: a few lines above here, I don't understand why you are doing `cuts[1:], cuts[0]`... it's like you are moving the 1st cut to the end?

danpovey: also on that line, a I think a line break before cuts[1:] would make it more readable.

examples/librimix/librimix.sh
```diff
@@ -37,4 +39,27 @@ lhotse cut mix-by-recording-id librimix/cuts_sources.yml.gz librimix/cuts_mix_dy
...
+# The next step is truncation - it makes sure that the cuts all have the same duration and makes them easily batchable
+
+# Truncate the pre-mixed cuts
+lhotse cut truncate \
+  --max-duration $DURATION \
+  --offset-type random \
+  --preserve-id \
+  librimix/cuts_mix.yml.gz librimix/cuts_mix_${DURATION}s.yml.gz
+
+# Truncate the dynamically-mixed clean cuts
```
danpovey, line 51: what do you mean by "clean cuts" if they are mixed or noisy?

lhotse/cut.py
```diff
@@ -78,31 +83,34 @@ def truncate(
...
-        changes happen (a subset of features is loaded).
-
-        Example:
-        >>> from math import isclose
-        >>> cut = Cut(id='x', channel=0, start=3.0, duration=8.0, features='dummy', supervisions=[])
-        >>> trimmed_cut = cut.truncate(offset=5.0, until=7.0)
-        >>> trimmed_cut.start == 5.0 and isclose(trimmed_cut.duration, 2.0) and isclose(cut.end, 7.0)
+        Returns a new Cut that is a sub-region of the current Cut.
+
+        Note that no operation is done on the actual features - it's only during the call to load_features()
```
danpovey: Just curious about something... what would it take to re-dump the features from such a manifest?  E.g. suppose you were planning to train for many epochs and did not want to incur repeated extra I/O?

lhotse/cut.py
```diff
@@ -78,31 +83,34 @@ def truncate(
...
+        when the actual changes happen (a subset of features is loaded).
+
+        :param offset: float (seconds), controls the start of the new cut relative to the current Cut's start.
+            E.g., if the current Cut starts at 10.0, and offset is 2.0, the new start is 12.0.
+        :param duration: optional float (seconds), controls the duration of the resulting Cut.
+            By default, the duration is (end of the cut before truncation) - (offset).
+        :param keep_excessive_supervisions: bool. Since trimming may happen inside a SupervisionSegment, the caller has
```
danpovey: line seems a bit long

examples/librimix/librimix.ipynb
pzelasko: Yeah, it's unnecessary. I thought that `reduce` requires an explicit starting value, but it turns out it doesn't. I simplified the usage in most places but missed this one.

pzelasko: Now I think that's because of using C++ `std::accumulate` too much in the past...

examples/librimix/librimix.sh
pzelasko: clean => non-noisy

lhotse/cut.py
pzelasko: It should be fairly simple, one method + one CLI mode. I'll add an issue.

examples/librimix/librimix.sh
pzelasko: Oh, I just noticed that it was a typo

examples/librimix/librimix.ipynb
danpovey: ha ha, OK

PR #37, opened 2020-06-27
bug when frames_to_trim == 0

popcornell:
Hey Piotr, I finally found some time to do a small recipe on Asteroid. 
I think I found a bug when loading cuts.  If frames_to_trim is zero the returned features tensor has zero frames. 
I issued this pull_request here... but I am not sure it is the proper place. Maybe directly on master ?


pzelasko:
Thanks @popcornell ! I changed the base to master. I'll merge after tests pass and add a unit test for this case later.

PR #40, opened 2020-06-29
recipe for mini_librispeech

jimbozhang:
I wrote a recipe (no bash version yet, I'll add it soon) for the mini_librispeech corpus. Not sure whether this implement is correct, so this PR is just for discussion for now.

This implement didn't do any padding or cutting. It just extracts the MFCC features and fetches the transcriptions.


examples/librimix/librimix.ipynb
```diff
@@ -84,7 +84,7 @@
     "for partition, manifests in librimix_manifests.items():\n",
     "    feature_set_builder = FeatureSetBuilder(\n",
     "        feature_extractor=FeatureExtractor(type='fbank'),\n",
-    "        output_dir=f'librimix_nb/feats_{partition}',\n",
+    "        output_dir=root_dir/f'librimix_nb/feats_{partition}',\n",
```
danpovey, line 87: this doesn't look right to me.

lhotse/utils.py
```diff
@@ -64,6 +67,16 @@ def non_null_dict_factory(collection):
     return asdict(dclass, dict_factory=non_null_dict_factory)
 
 
+def find_files_in_directory(path: Pathlike = '.', pattern: str = '*') -> List[str]:
```
danpovey, line 70: Perhaps should mention that '*' is supposed to be a UNIX-style shell wildcard, not a python regex,
and give a few examples of values that you think might be useful?  It's not even clear to me, as UNIX
has various types of pattern matching e.g. globs.

examples/librimix/librimix.ipynb
jimbozhang: fixed.

examples/librimix/librimix.ipynb
```diff
@@ -84,7 +84,7 @@
     "for partition, manifests in librimix_manifests.items():\n",
     "    feature_set_builder = FeatureSetBuilder(\n",
     "        feature_extractor=FeatureExtractor(type='fbank'),\n",
-    "        output_dir=f'librimix_nb/feats_{partition}',\n",
+    "        output_dir='{}/librimix_nb/feats_{}'.format(root_dir, partition),\n",
```
danpovey, line 87: Mm, it might be worthwhile considering the case where root_dir is in a shared place not owned by this user, and they want to dump feats to somewhere else.

lhotse/cut.py
```diff
@@ -467,7 +467,10 @@ def __add__(self, other: 'CutSet') -> 'CutSet':
         return CutSet(cuts={**self.cuts, **other.cuts})
 
 
-def make_cuts_from_features(feature_set: FeatureSet) -> CutSet:
+def make_cuts_from_features(
```
pzelasko, line 470: There is another function just below called `make_cuts_from_supervisions` - was it not sufficient or not visible? Feedback on this is welcome, I'm still trying to figure out what is the best way to make different cutting strategies discoverable.

examples/librimix/librimix.ipynb
pzelasko: I think it should be already possible.

.gitignore
```diff
@@ -2,10 +2,7 @@
 .idea/
 .vscode/
 deps/
-examples/librimix/MiniLibriMix/
-examples/librimix/MiniLibriMix.zip
-examples/librimix/librimix_nb/
-examples/librimix/librimix/
+examples/*/env/
```
pzelasko, line 5: I'm not sure about the `env` name - it collides with the `virtualenv` directory naming convention, suggesting that we are having a separate Python env for each recipe. How about an old-fashioned `data` instead?

lhotse/cut.py
jimbozhang: Sorry, I didn't notice it. The function `make_cuts_from_supervisions` is sufficient.

.gitignore
jimbozhang: Since the main purpose of  lhotse is dealing with data, I think the name `data` would be okay.

lhotse/dataset/speech_recognition.py
```diff
@@ -18,23 +18,23 @@ class SpeechRecognitionDataset(Dataset):
...
         cut_id = self.cut_ids[idx]
-        cut = self.data_set.cuts[cut_id]
+        cut = self.cuts[cut_id]
 
         feature = torch.from_numpy(cut.load_features(root_dir=self.root_dir))
 
         return {
             'feature': feature,
-            'text': cut.supervisions.text
+            'text': cut.supervisions[0].text
```
pzelasko, line 37: what if there's more than one supervision? can you leave a TODO comment here saying that we'll have to work this out for multi-supervision cuts?

lhotse/recipes/mini_librispeech.py
```diff
@@ -48,17 +53,17 @@ def prepare_mini_librispeech(
...
         for trans_path in part_path.rglob('*.txt'):
             with open(trans_path) as f:
                 for line in f:
                     idx, text = line.split(maxsplit=1)
                     audio_path = part_path / Path(idx.replace('-', '/')).parent / f'{idx}.flac'
                     if audio_path.is_file():
-                        audio_path = str(audio_path)
-                        metadata[idx] = MetaDataType(audio_path=audio_path, audio_info=torchaudio.info(audio_path), text=text)
+                        metadata[idx] = LibriSpeechMetaData(audio_path=audio_path,
+                                                            audio_info=torchaudio.info(str(audio_path))[0],
```
pzelasko, line 65: I don't know `torchaudio.info` very well; what's in index 0, and then in later indices?

.gitignore
pzelasko: Great! Could you rename? I recall it's not the only place where this name is used.

lhotse/dataset/speech_recognition.py
jimbozhang: I think we can assume that there should be only one supervision for each cut in ASR tasks. I put a warning when more supervisions detected.

examples/librimix/librimix.ipynb
jimbozhang: The bash scripts embedded in Librimix's ipynb was rewrote in Python. And it allowed to set output directory like MiniLibriSpeech's.

lhotse/dataset/speech_recognition.py
pzelasko: For LibriSpeech, there will be single supervision; however, that won't be true for all ASR tasks. Later on, for corpora like conversations (e.g., SWBD), we'll want to create cuts that might span more than a single utterance, and that's where multiple supervisions will come from. But we can handle it once the need arises, it's fine for now.

pzelasko:
@jimbozhang is it ready to merge? Looks good to me.

jimbozhang:
> @jimbozhang is it ready to merge? Looks good to me.

Yes, I think it's okay to merge 😃

PR #42, opened 2020-07-01
PaddingCut for dynamic sequence padding to common length

pzelasko:


pzelasko:
We'll need padding to construct evenly-sized batches for datasets such as LibriSpeech where no acoustic context is available.

I've made it easy to filter and pad the whole CutSet, both in Python and the CLI, so we can make recipes with bucketing - e.g. 5 second cuts bucket, 10 seconds cuts bucket, 15 seconds cut bucket, and so on.


Ready for review and merge.

pzelasko:
@danpovey can you take a look at this PR? I already included these changes in the documentation.

lhotse/dataset/source_separation.py
```diff
@@ -5,9 +5,7 @@
 from torch.utils.data import Dataset
 
 from lhotse.cut import CutSet, Cut, AnyCut
-from lhotse.utils import Pathlike
-
-EPS = 1e-8
```
danpovey, line 10: It's not very clear on reading  that EPS refers to epsilon.  I wonder if EPSILON or epsilon might be clearer.

lhotse/cut.py
```diff
@@ -181,20 +193,107 @@ def from_dict(data: dict) -> 'Cut':
...
+    def supervisions(self):
+        return []
+
+    @property
+    def frame_shift(self):
+        return round(self.duration / self.num_frames, ndigits=3)
+
+    # noinspection PyUnusedLocal
+    def load_features(self, *args, **kwargs) -> np.ndarray:
+        value = np.log(EPS) if self.use_log_energy else EPS
```
danpovey, line 216: Just checking: are we normalizing the signal to [-1,1] before doing FFT, or doing the old HTK/Kaldi compatibility thing in [-32k,32k]?

lhotse/cut.py
```diff
@@ -141,26 +150,10 @@ def truncate(
...
-            [MixTrack(cut=other, offset=offset_other_by, snr=snr)]
-            if isinstance(other, Cut)
-            else other.tracks
-        )
-        return MixedCut(
-            id=str(uuid4()),
-            tracks=[MixTrack(cut=self)] + new_tracks
-        )
+        """Refer to mix() documentation."""
+        return mix(self, other, offset_right_by=offset_other_by, snr=snr)
```
danpovey, line 154: offset_right_by is a little confusing... I get that it refers to the right-most arg to the function but it looks like it might be something to do with the RHS (end) of the recording.  I wonder if offset_a_by and offset_b_by might be clearer than left/right?  Or cut1,cut2 instead of a,b?

lhotse/cut.py
```diff
@@ -170,6 +163,25 @@ def append(self, other: 'Cut', snr: Optional[Decibels] = None) -> 'MixedCut':
         """
         return self.overlay(other=other, offset_other_by=self.duration, snr=snr)
 
+    def pad(self, desired_duration: Seconds) -> AnyCut:
+        """
+        Return a new MixedCut, padded to `desired_seconds` duration.
+
+        :param desired_duration: The cuts minimal duration after padding.
```
danpovey, line 170: cut's, not cuts, I think

danpovey: Also perhaps should clarify that it's padding with low-energy (energy = EPS in each bin)?

lhotse/cut.py
pzelasko: Yeah, we're in the [-1, 1] range. I'm open to extending the padding with more options. I was thinking specifically about accepting a specified energy value from the user and random noise padding. Lowest seen energy sounds good too.

lhotse/cut.py
pzelasko: Alright, we'll need to rename it. Currently, it's not possible to specify the offset for the "left/a/1" track - to create the mix, we have to assume that one of the tracks is a reference point to which all the offsets refer. So I assume the left-most track in the mix will always have a zero offset.

We could rename the params like this: 

    def mix(reference_cut, mixed_in_cut, offset, snr): ...

would that make it clearer?

danpovey:
Sure, that seems clearer.

pzelasko:
@danpovey should be better now.

PR #45, opened 2020-07-02
Documentation stub

pzelasko:
I've created an absolutely minimal stub for docs; it's a listing of all classes and functions in the codebase. As we go forward, we'll start putting some descriptions about the aims of this library and how to use it.


pzelasko:
I'm merging so that read the docs can render this, I'll fill out next chapters in next PRs.

PR #47, opened 2020-07-06
An utility for code style checking based on flake8

jimbozhang:


pzelasko:
Thanks! LGTM. 

BTW In the future, we might want to use https://github.com/psf/black as autoformatter - I'm using PyCharm and it's reformatting the code before each commit automatically, but once there are more contributors it'll probably be good to have an easily accessible CLI utility.

PR #48, opened 2020-07-07
Cuts documentation

pzelasko:


pzelasko:
Note that once the documentation is built, you can preview how it will look like by clicking on "Details" next to it:

image

PR #50, opened 2020-07-07
Add features documentation

pzelasko:


docs/features.rst
```diff
@@ -1,3 +1,134 @@
...
+and the frame shift.
+You only need to know two of those values to compute the third one, regardless of the frame length.
+This is equivalent of having Kaldi's _snip_edges_ parameter set to False.
+
+
+Storing features
+****************
+
+Features in Lhotse are stored as numpy matrices with shape ``(num_frames, num_features)``.
+By default, we use `lilcom`_ for lossy compression and reduce the size on the disk approximately by half.
```
danpovey, line 21: confused by underscore

danpovey: i just noticed that we have energy_floor 0.0 and dither 0.0 as default configs.  Seems like a dangerous combination as 0-energy signal might lead to NaN's.

danpovey: ... also this file when compiled says `© Copyright 2020, Kaldi development team Revision 01f48c8d.` ... should be Lhotse development team?

docs/requirements.txt
```diff
@@ -1,4 +1,5 @@
 numpy>=1.18.1
 sphinx_rtd_theme
-sphinx==3.1.0
-sphinx-click==2.3.2
\ No newline at end of file
+sphinx==3.1.2
```
danpovey, line 3: dunno if exact requirements might cause problems?  Hope this is not built when users build their code.

docs/features.rst
pzelasko: Underscore means its a hyperlink; the actual link is at the bottom of the page ("bibliography style").

I'll fix the energy_floor at 0.1.

Kaldi/Lhotse dev team - it's your call. I was under the impression we're still considering K2/Lhotse a part of the Kaldi effort.

docs/requirements.txt
pzelasko: No, these are only used for building the documentation. Actually it's the lack of exact versions that would have caused problems, as some environments like read-the-docs use older, incompatible sphinx by default.

docs/features.rst
danpovey: Thanks!  Let's call it Lhotse dev team for now.  It's all part of next-gen Kaldi but it's not clear that that will actually have the name "Kaldi" (since the project will be mostly independent code-wise, it might lead to confusion.)

danpovey: BTW the underscore didn't seem to display correctly, it displayed as  just an underscore.

pzelasko: Oh, I can see it now. Yeah I'll fix that, thanks.

PR #51, opened 2020-07-10
Dataset documentation + fix most sphinx warnings

### LINKED ISSUE #29: Create documentation
I'm thinking about a "read-the-docs" style doc page and adding some rationale, main features, and some examples in the README.md.

pzelasko:
Resolves #29

### DIFF
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -14,7 +14,16 @@ EPS = 1e-8
 class SpeechRecognitionDataset(Dataset):
     """
     The PyTorch Dataset for the speech recognition task.
-    Contains acoustic features and the corresponding text.
+    Returns a dict of:
+
+    .. code-block::
+
+        {
+            'features': (T x F) tensor
+            'text': string
+        }
+
+    In the future, will be extended by graph supervisions.
     """
 
     def __init__(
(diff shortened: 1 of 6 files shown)


lhotse/dataset/unsupervised.py
```diff
@@ -10,7 +10,8 @@
 class UnsupervisedDataset(Dataset):
     """
     Dataset that contains no supervision - it only provides the features extracted from recordings.
-    The returned features are a torch Tensor of shape (T x F).
+    The returned features are a :class:`torch.Tensor` of shape ``(T x F)``, where T is the number of frames,
+    and F is the feature dimension.
```
danpovey, line 14: How will you implement collate_fn?

lhotse/dataset/speech_recognition.py
```diff
@@ -14,7 +14,16 @@
 class SpeechRecognitionDataset(Dataset):
     """
     The PyTorch Dataset for the speech recognition task.
-    Contains acoustic features and the corresponding text.
+    Returns a dict of:
```
danpovey, line 17: When naming this, we might want to bear in mind that in future there may be other versions of this.
E.g. we may want to do padding while collating, to ensure even minibatch sizes.  
(I'm thinking that may be easier to do by implementing the data-loader with no underlying Dataset,
just reading the CutSet directly).
Up to you though.

lhotse/dataset/unsupervised.py
pzelasko: I'm considering a scenario where a custom collate_fn is not necessary; we'd instead truncate/pad all the samples into a number of duration bins, and implement a [Sampler](https://pytorch.org/docs/stable/data.html?highlight=sampler#torch.utils.data.Sampler) that constructs batches only from same duration samples. Then we'd only need to implement collate_fn if we opt into returning custom type objects from dataset (like graphs); even then, the only modification to the default collate_fn would be "informing it" how to batch these objects together.

pzelasko: "truncate/pad all the samples into a number of duration bins" -> I meant in the CutSet

lhotse/dataset/speech_recognition.py
pzelasko: Yeah, there'll definitely be more versions; we should rename as necessary. Re collating, check out my previous comment.

PR #55, opened 2020-07-16
Extensible feature extraction

pzelasko:
I'm looking for some feedback here before I go on and implement the tests, fix bugs, and sort out the details. Please check out the features documentation in this PR for a detailed description.


pzelasko:
@danpovey @jimbozhang could you have a look?

docs/features.rst
```diff
@@ -23,66 +22,106 @@ For now, we extract the features for the whole recordings, and store them in sep
...
-- a list of documents describing each feature matrix.
-
-The feature extractor config is a detailed list of arguments used to extract the features in a given manifest.
-Currently, these arguments are the sum of all possible arguments for all feature types - we will likely refactor that
-in the future, and we might either move this section into a separate manifest or drop it entirely and store
-only the non-default values in each feature document.
-
-The feature documents contain the information necessary to tie them to a particular recording - ``start``, ``duration``,
+    dither: 0.0
+    energy_floor: 0.1
```
danpovey, line 35: Mm... if our signals are normalized to between -1 and 1, then I think energy-floor=0.1 is way too big.  This is appropriate in a Kaldi setting but not if we normalize the energy.

docs/features.rst
```diff
@@ -23,66 +22,106 @@ For now, we extract the features for the whole recordings, and store them in sep
...
+        name = 'example-feature-extractor'
+        config_type = ExampleFeatureExtractorConfig
+
+        def extract(self, samples: np.ndarray, sampling_rate: int) -> np.ndarray:
+            f, t, Zxx = stft(samples, sampling_rate, noverlap=round(self.frame_shift * sampling_rate))
+            return np.abs(Zxx)
+
+        @property
+        def frame_shift(self) -> Seconds:
+            return self.config.frame_shift
```
danpovey, line 97: It might be a good idea to mention at some point that the lilcom compression method uses a fixed precision that doesn't depend on the magnitude of the thing being compressed, so it's better suited to log-energy features than energy features. (Of course it can be disabled).

docs/features.rst
```diff
@@ -23,66 +22,106 @@ For now, we extract the features for the whole recordings, and store them in sep
 We retrieve them by loading the whole feature matrix and selecting the relevant region (e.g. specified by a cut).
 Eventually we will look into optimizing the storage to further reduce the I/O.
 
-The feature matrices are accompanied by a YAML feature manifest. An example:
+There are two types of manifests:
```
danpovey, line 25: I notice above it says 'lilcom' is reducing the size by about half.  It takes a num_bits which defaults to -8, I think, which defines the 'tick size' or precision (-8 -> 1/256).  That may be a little too precise for log-energy features.   Perhaps -5 would be more appropriate as a default?

docs/features.rst
pzelasko: I'll change to 1e-5

docs/features.rst
pzelasko: Sure!

docs/features.rst
pzelasko: Added to docs.

pzelasko:
@danpovey this is ready for review and merge

pzelasko:
@danpovey any chance you'll find the time to review this one, and then #61?

docs/features.rst
danpovey: should it be changed in this example (documentation) too?

danpovey: .. also I think 1e-10 would be more appropriate since it's the square of a magnitude.

docs/features.rst
```diff
@@ -19,70 +18,111 @@ Storing features
...
+        raise ValueError(f'The feature extractor\'s "mix" operation is undefined.')
+
+    @staticmethod
+    def compute_energy(features: np.ndarray) -> float:
+        raise ValueError(f'The feature extractor\'s "compute_energy" is undefined.')
+
+They are:
+
+- ``mix()`` which specifies how to mix two feature matrices to obtain a new feature matrix representing the sum of signals;
+- ``compute_energy()`` which specifies how to obtain a total energy of the feature matrix, which is needed to mix two signals with a specified SNR.
```
danpovey, line 122: Can you be more specific about 'gain' and the return value of compute_energy, e.g. must they be in dB (and can you give an example of how dB is computed, because I think there are energy and magnitude versions).

danpovey, line 122: Can you be more specific about 'gain' and the return value of compute_energy, e.g. must they be in dB (and can you give an example of how dB is computed, because I think there are energy and magnitude versions).

lhotse/dataset/unsupervised.py
```diff
@@ -28,11 +30,42 @@ def __getitem__(self, item: int) -> torch.Tensor:
...
     def __getitem__(self, item: int) -> torch.Tensor:
         cut = self.cuts[self.cut_ids[item]]
         audio = cut.load_audio(root_dir=self.root_dir)
         return torch.from_numpy(audio)
+
+    def _validate(self):
+        assert all(cut.has_recording for cut in self.cuts)
+
+
+class OnTheFlyFeatureExtractionUnsupervisedDataset(UnsupervisedDataset):
```
danpovey, line 53: Maybe could shorten to DynamicUnsupervisedDataset?

lhotse/features/mixer.py
```diff
@@ -0,0 +1,106 @@
...
+        # When there is an offset, we need to pad before the start of the features we're adding.
+        if offset > 0:
+            feats_to_add = np.vstack([
+                self.log_energy_floor * np.ones((num_frames_offset, self.num_features)),
+                feats_to_add
+            ])
+
+        # When the features we're mixing in are shorter that the anticipated mix length,
+        # we need to pad after their end.
+        # Note: we're doing that non-efficiently, as it we potentially re-allocate numpy arrays twice,
```
danpovey, line 86: typo

lhotse/features/mixer.py
```diff
@@ -0,0 +1,106 @@
...
+        :param frame_shift: Required to correctly compute offset and padding during the mix.
+        :param log_energy_floor: The value used to pad the shorter features during the mix.
+        """
+        self.feature_extractor = feature_extractor
+        # The mixing output will be available in self.mixed_feats
+        self.mixed_feats = base_feats
+        # Keep a pre-computed energy value of the features that we initialize the Mixer with;
+        # it is required to compute gain ratios that satisfy SNR during the mix.
+        self.frame_shift = frame_shift
+        self.reference_energy = feature_extractor.compute_energy(base_feats)
```
danpovey, line 37: is there a requirement that energy never be zero?  Please clarify where compute_energy is documented if so!

test/dataset/test_unsupervised_dataset.py
```diff
@@ -21,3 +24,26 @@ def test_unsupervised_waveform_dataset(libri_cut_set):
...
+        cuts=libri_cut_set
+    )
+    ref_feats = ref_dataset[0]
+    tested_feats = tested_dataset[0]
+    # Note: comparison to 1 decimal fails.
+    #       I'm assuming this is due to lilcom's compression.
+    #       Pytest outputs looks like the following:
+    # E       Mismatched elements: 4 / 23000 (0.0174%)
+    # E       Max absolute difference: 0.46469784
+    # E       Max relative difference: 0.6171043
```
danpovey, line 42: Mm, lilcom's compression should not create this much difference, it should be rounding exactly to the closest 2^-5 given the current default, so this shouldn't be a lilcom issue.  Can you look more into this?

pzelasko:
I'm addressing the comments. In the meanwhile, I investigated the decimal in mixing spectrograms, it seems like an artifact of feature-domain mixing. Take a look at the differences in two scenarios. Each picture is ordered left -> right or top -> down and shows:
- signal1, 
- signal2, 
- mix in time domain, 
- mix in feature domain, 
- absolute difference between mixes

1) A toy example, mixing of two sine waves (550Hz and 4400Hz)

image

2) A more realistic example, mixing of two 3 second cuts from mini_librispeech

image


It seems we're observing the biggest differences between feature-domain and time-domain mixing in the time-frequency regions that are non-speech/silence in both recordings. I can't think of any obvious mistake done in the processing... 
I'm wondering though if maybe it makes more sense to mix amplitude-spectrum instead of power-spectrum, by doing additional root and pow2 operations during the mix, but it did not affect the results in a significant way.

Since I initially thought that the reference features extracted in tests could have been outdated, I refactored the code a little bit to make it easy to recreate reference cuts/features for unit tests - I'll be pushing the changes soon, but wait with the review - I've still got other comments to address.

pzelasko:
The sine waves look suspicious to me though - I've assumed it's windowing artifacts so far, but maybe that assumption is wrong?

docs/features.rst
pzelasko: Check out if the new description makes it clearer.

pzelasko:
This should be ok to review now. Note that I also refactored (simplified) the way the `CutSet` is created from other manifests and made the randomly generated manifest IDs replicable if the user specifies a random seed. Now we can easily re-create the reference data in our unit tests.

lhotse/dataset/unsupervised.py
```diff
@@ -28,11 +30,45 @@ def __getitem__(self, item: int) -> torch.Tensor:
...
 
+    def _validate(self):
+        assert all(cut.has_features for cut in self.cuts)
+
 
 class UnsupervisedWaveformDataset(UnsupervisedDataset):
-    """A special UnsupervisedDataset which provides waveform instead of features."""
+    """
+    A variant of UnsupervisedDataset that provides waveform samples instead of features.
+    The output is a tensor of shape (C, T), with C being the number of channels and T the number of audio samples.
```
danpovey, line 40: you really want such long lines?

lhotse/features/base.py
```diff
@@ -0,0 +1,411 @@
...
+
+    @abstractmethod
+    def extract(self, samples: np.ndarray, sampling_rate: int) -> np.ndarray: ...
+
+    @property
+    @abstractmethod
+    def frame_shift(self) -> Seconds: ...
+
+    @staticmethod
+    def mix(features_a: np.ndarray, features_b: np.ndarray, gain_b: float) -> np.ndarray:
```
danpovey, line 54: would be nice to have a doc-string saying e.g. what gain_b means.  E.g. is it in db?  (Clearer variable name, if so)...

lhotse/features/base.py
```diff
@@ -0,0 +1,411 @@
...
+        elif self.storage_type == 'numpy':
+            features = np.load(storage_path, allow_pickle=False)
+        else:
+            raise ValueError(f"Unknown storage_type: {self.storage_type}")
+
+        # In case the caller requested only a subset of features, trim them
+
+        # Left trim
+        if not isclose(start, self.start):
+            frames_to_trim = round((start - self.start) / self.frame_shift)
```
danpovey, line 191: may not give informative error message if start < self.start?

lhotse/features/example.py
```diff
@@ -0,0 +1,31 @@
...
+    A minimal class example, showing how to implement a custom feature extractor in Lhotse.
+    Note that the "@register_extractor" decorator is commented out so that this example
+    is not visible beyond this file.
+    """
+    name = 'example-feature-extractor'
+    config_type = ExampleFeatureExtractorConfig
+
+    def extract(self, samples: np.ndarray, sampling_rate: int) -> np.ndarray:
+        f, t, Zxx = stft(samples, sampling_rate, noverlap=round(self.frame_shift * sampling_rate))
+        return np.abs(Zxx)
```
danpovey, line 27: might want to clarify that this would interact badly with lilcom compression?

lhotse/utils.py
```diff
@@ -17,11 +18,24 @@
 INT16MAX = 32768
 EPSILON = 1e-8
 
+_lhotse_uuid: Optional[Callable] = None
+
 
 def fix_random_seed(random_seed: int):
```
danpovey, line 24: what is _lhotse_uuid for?

lhotse/dataset/unsupervised.py
pzelasko: I've always used 120 character limit, but I'm open to change. What's the reason to have the limit shorter and what number are you suggesting?

lhotse/features/base.py
pzelasko: Good catch

lhotse/features/base.py
pzelasko: While writing this doc-string, I noticed that when mixing audio, we were using a gain value estimated for energy/power mixing, and not amplitude mixing. I added an `sqrt()` call and a comment in that place.

lhotse/utils.py
pzelasko: I've added the following comment here:
```
# This is a utility that generates uuid4's and is set when the user calls
# the ``fix_random_seed`` function.
# Python's uuid module is not affected by the ``random.seed(value)`` call,
# so we work around it to provide deterministic ID generation when requested.
```

danpovey:
Great!  No problems.  120char limit is fine.  I would have used 100 but up to you.

Issue #44, opened 2020-07-02
Support for working with raw audio


pzelasko:
Currently, Lhotse is somewhat based on the assumption that a user will mostly work with features such as log-mel energies. It'll be useful to have PyTorch Datasets providing raw audio as well. I'm thinking we should make it a convention to put `Waveform` in the class name of datasets that provide raw audio (e.g. `UnsupervisedWaveformDataset`). I'll try to list the steps that are needed to get there. 

- [ ] Create `Cut`s directly from `RecordingSet` (with an empty `Features` field - currently, you have to extract _some_ features to make cuts, just to ignore them later)
- [x] load_audio() in `Cut`
- [ ] load_audio() in `MixedCut`
- [x] load_audio() in `PaddingCut`
- [ ] Implement `UnsupervisedWaveformDataset`
- [ ] Implement other waveform datasets

jimbozhang:
If nobody is working on this issue, I'll implement this.

danpovey:
Great as far as I'm concerned!

pzelasko:
Awesome! 

One suggestion for a starting point: it's probably the simplest if we add a `recording: Optional[Recording] = None` field to the `Cut` class, and make `features` optional as well. If a user wants a cut with audio loading capability, let them make sure they created it this way. This will avoid the manifests growing even bigger for users who do not need this capability.

jimbozhang:
> One suggestion for a starting point: it's probably the simplest if we add a `recording: Optional[Recording] = None` field to the `Cut` class, and make `features` optional as well. If a user wants a cut with audio loading capability, let them make sure they created it this way. This will avoid the manifests growing even bigger for users who do not need this capability.

I wonder is it appropriate to take the `Recording` as a special feature:

`GeneralFeatures = Union['Features', 'Recording']`, and then define non-optional `features: GeneralFeatures` in the class `Cut`, instead of `features: Features = None; recording: Recording = None`

The motivation is that a `Cut` is defined as an example for training, which should contains one type of features and the corresponding label. The features here should be allowed in any form, not limited in spectrum. So it is a little redundancy to contain both raw audio and cepstral features in `Cut`.

danpovey:
There are situations where you might want to access both the features and the raw recording, for different reasons.  I think it makes sense to keep the concepts separate, especially since they have different structure.

[...]

Issue #13, opened 2020-05-20
Validate manifest


pzelasko:
We could use a python function `validate_manifest` and a corresponding CLI mode `lhotse validate` at some point to check data integrity (like Kaldi `validate_data_dir.sh`).

danpovey:
Cool.

PR #56, opened 2020-07-17
Support for working with raw audio #44

jimbozhang:
For #44. In this PR, there are some points that I'm not sure if those are appropriate ...
- Let `MixedCut.load_audio()` returns a list of `ndarray` over the tracks. 
- Renamed `make_cuts_from_supervisions` to `make_cuts_from_supervisions_features` and added a function `make_cuts_from_supervisions_audio`.

I'll update the codes following your review comments. @pzelasko @danpovey 

About the array cuts for array-mic, I think we may discuss that after merging this PR. :relaxed:


lhotse/bin/modes/cut.py
```diff
@@ -45,7 +46,7 @@ def simple(
         cut_set = make_cuts_from_features(feature_set)
     else:
         supervision_set = SupervisionSet.from_yaml(supervision_manifest)
-        cut_set = make_cuts_from_supervisions(feature_set=feature_set, supervision_set=supervision_set)
+        cut_set = make_cuts_from_supervisions_features(feature_set=feature_set, supervision_set=supervision_set)
```
pzelasko, line 49: that's a step in a good direction - I think we'll eventually group these "make_cuts_" functions into some sort of "Editor"/"Cutter" class that makes them in different ways (not in this PR).

lhotse/cut.py
```diff
@@ -42,56 +42,68 @@ class Cut:
...
 
+    # Supervisions that will be used as targets for model training later on. They don't have to cover the whole
+    # cut duration. They also might overlap.
+    supervisions: List[SupervisionSegment]
+
     # The features can span longer than the actual cut - the Features object "knows" its start and end time
     # within the underlying recording. We can expect the interval [begin, begin + duration] to be a subset of the
     # interval represented in features.
-    features: Features
+    features: Features = None
```
pzelasko, line 56:
```suggestion
    features: Optional[Features] = None
```

lhotse/cut.py
```diff
@@ -42,56 +42,68 @@ class Cut:
...
 
-    # Supervisions that will be used as targets for model training later on. They don't have to cover the whole
-    # cut duration. They also might overlap.
-    supervisions: List[SupervisionSegment]
+    # For the cases that the model was trained by raw audio instead of features
+    recording: Recording = None
```
pzelasko, line 59:
```suggestion
    recording: Optional[Recording] = None
```

pzelasko: I'm wondering if it makes sense for a Cut to have neither Recording nor Features; probably not? But I think I'm ok with not enforcing it for now, we can add it to the list of things to validate when working on #13

lhotse/cut.py
```diff
@@ -42,56 +42,68 @@ class Cut:
...
+        return self.features.recording_id if self.features is not None else self.recording.id
 
     @property
     def end(self) -> Seconds:
         return self.start + self.duration
 
     @property
     def frame_shift(self) -> Seconds:
-        return self.features.frame_shift
+        return self.features.frame_shift if self.features is not None else 0
```
pzelasko, line 76: when there's no features I'd rather return `None` and change the type to `Optional[Seconds]`; returning 0 might lead to silent bugs in some cases

lhotse/cut.py
```diff
@@ -42,56 +42,68 @@ class Cut:
...
 
     @property
     def num_frames(self) -> int:
```
pzelasko, line 79: I'd rather make this `Optional[int]` and add a property `num_samples` (also returning `Optional[int]`)

lhotse/cut.py
```diff
@@ -42,56 +42,68 @@ class Cut:
...
-        return round(self.duration / self.frame_shift)
+        # Return samples number if this Cut only has raw audio
+        return round(self.duration / self.frame_shift) if self.features is not None else self.recording.num_samples
 
     @property
     def num_features(self) -> int:
-        return self.features.num_features
+        return self.features.num_features if self.features is not None else 0
```
pzelasko, line 85: same as with `frame_shift`

lhotse/cut.py
```diff
@@ -42,56 +42,68 @@ class Cut:
...
 
     @property
     def sampling_rate(self) -> int:
-        return self.features.sampling_rate
+        return self.features.sampling_rate if self.features is not None else self.recording.sampling_rate
 
     def load_features(self, root_dir: Optional[Pathlike] = None) -> np.ndarray:
```
pzelasko, line 91:
```suggestion
    def load_features(self, root_dir: Optional[Pathlike] = None) -> Optional[np.ndarray]:
```

lhotse/cut.py
```diff
@@ -102,13 +114,21 @@ def load_audio(self, recording_set: RecordingSet, root_dir: Optional[Pathlike] =
...
         :return: a numpy ndarray with audio samples, with shape (1 <channel>, N <samples>)
         """
-        return recording_set.load_audio(
-            self.recording_id,
-            channels=self.channel,
-            offset_seconds=self.start,
-            duration_seconds=self.duration,
-            root_dir=root_dir
-        )
+        if recording_set is None:
```
pzelasko, line 117: Can we remove the `recording_set` argument? At this point, if a user wants to be able to load audio from the Cut, they should just attach the recording to it. It will simplify this function.

lhotse/cut.py
```diff
@@ -172,6 +192,9 @@ def pad(self, desired_duration: Seconds) -> AnyCut:
         :param desired_duration: The cut's minimal duration after padding.
         :return: a padded MixedCut if desired_duration is greater than this cut's duration, otherwise self.
         """
+        if self.features is None:
```
pzelasko, line 195: I think this `if` is unnecessary - `PaddingCut` already supports padding audio with zeros IIRC.

pzelasko: That would actually make a great unit test.

lhotse/cut.py
```diff
@@ -186,11 +209,13 @@ def pad(self, desired_duration: Seconds) -> AnyCut:
 
     @staticmethod
     def from_dict(data: dict) -> 'Cut':
-        feature_info = data.pop('features')
-        supervision_infos = data.pop('supervisions')
+        features = Features.from_dict(data.pop('features')) if 'features' in data else None
+        recording = Recording.from_dict(data.pop('recording')) if 'recording' in data else None
```
pzelasko, line 213: I think this doesn't work when the dict has `{..., 'features': None, ...}`? Can you add a unit test for that case (same with `recording`?

lhotse/cut.py
```diff
@@ -299,7 +324,7 @@ def from_dict(data: dict):
         raw_cut = data.pop('cut')
         try:
             cut = Cut.from_dict(raw_cut)
-        except:
+        except TypeError:
```
pzelasko, line 327: Are you sure it's going to be a `TypeError`? I admit I was lazy and did not check when writing this, and the change is in the right direction, just double-checking that this does not break loading PaddingCuts

lhotse/cut.py
```diff
@@ -343,7 +368,7 @@ def num_features(self) -> int:
 
     @property
     def features_type(self) -> str:
```
pzelasko, line 370:
```suggestion
    def features_type(self) -> Optional[str]:
```

lhotse/cut.py
```diff
@@ -477,7 +502,16 @@ def load_features(self, root_dir: Optional[Pathlike] = None) -> np.ndarray:
             )
         return mixer.mixed_feats
 
-    # TODO: load_audio()
+    def load_audio(self,
+                   recording_set: Optional[RecordingSet] = None,
```
pzelasko, line 506: Let's remove `recording_set` here too as per previous comment

pzelasko: I think this function should be renamed to `load_unmixed_audio`, and there should be one next to it called `load_audio` that applies the offsets and SNRs encoded in `MixTrack`s similarly as in `load_features`

lhotse/cut.py
```diff
@@ -695,6 +745,23 @@ def make_cuts_from_supervisions(supervision_set: SupervisionSet, feature_set: Fe
...
+    """
+    return CutSet.from_cuts(
+        Cut(
+            id=str(uuid4()),
+            start=supervision.start,
+            duration=supervision.duration,
+            recording=recording_set[supervision.recording_id],
+            supervisions=[supervision]
+        )
+        for idx, supervision in enumerate(supervision_set)
```
pzelasko, line 761:
```suggestion
        for supervision in supervision_set
```

test/cut/test_cut.py
```diff
@@ -37,3 +61,26 @@ def test_load_features(libri_cut):
...
+def test_load_none_features(libri_cut):
+    libri_cut.features = None
+    feats = libri_cut.load_features()
+    assert feats is None
+
+
+def test_make_cuts_from_recordings(libri_recording_set):
+    expected_duration = 16.04
+    cutset = make_cuts_from_recordings(libri_recording_set)
+    duration = list(cutset.cuts.values())[0].duration
```
pzelasko, line 75: Note to self: we should probably make cutset work with both str (id) and int (index) based indexing

test/cut/test_cut.py
```diff
@@ -37,3 +61,26 @@ def test_load_features(libri_cut):
...
+    assert isclose(duration, expected_duration)
+
+
+def test_make_cuts_from_supervisions_features(supervision_set, libri_features_set):
+    cutset = make_cuts_from_supervisions_features(supervision_set, libri_features_set)
+    assert len(cutset) == 2
+
+
+def test_make_cuts_from_supervisions_recordingss(supervision_set, libri_recording_set):
```
pzelasko, line 84:
```suggestion
def test_make_cuts_from_supervisions_recordings(supervision_set, libri_recording_set):
```

test/cut/test_padding_cut.py
```diff
@@ -220,6 +220,14 @@ def test_pad_simple_cut(libri_cut):
     np.testing.assert_almost_equal(pre_mixed_feats, mixed_feats[:1604, :])
 
 
+def test_pad_cut_no_features(libri_cut):
+    libri_cut.features = None
+    padded = libri_cut.pad(desired_duration=20.0)
+
+    # TODO: implement padding for cut without features in libri_cut.pad()
+    assert padded.duration <= 20.0
```
pzelasko, line 228: It should be `==`, right?

lhotse/cut.py
```diff
@@ -477,7 +502,16 @@ def load_features(self, root_dir: Optional[Pathlike] = None) -> np.ndarray:
...
-    # TODO: load_audio()
+    def load_audio(self,
+                   recording_set: Optional[RecordingSet] = None,
+                   root_dir: Optional[Pathlike] = None,
+    ) -> List:
+        """Loads the audios of the source cuts and put them into a list."""
+        cuts = [track.cut for track in self.tracks]
+        audios = []
+        for cut in cuts:
+            audios.append(cut.load_audio(recording_set, root_dir))
```
pzelasko, line 513: I'm not sure if this is right - I think it does not respect the `offset` parameter encoded in the `MixTrack` object that holds the cut, so the returned audio could be in a wrong time span. I saw that you have a test for that but maybe you could extend it to explicitly check such case.

test/cut/test_padding_cut.py
jimbozhang: Because padding hasn't been implemented for raw audio, the duration here should be less than 20.0.
It will be implemented in future (in next PR, I think).

test/cut/test_cut.py
jimbozhang: Agree. The indexing by string(id) should not cost much.

lhotse/cut.py
jimbozhang: I made two unit tests (`test_load_none_feats_cut_set`, `test_load_none_recording_cut_set`) and those are passed ...

test/cut/test_padding_cut.py
jimbozhang: I remove this unit test because we assumes that `Cut.features` is not None when calling `Cut.pad()`.

lhotse/cut.py
jimbozhang: By analyzing `Cut.from_dict()`, I believe the expected exception type here should be `TypeError`. And according the following codes:
```
try:
    cut = Cut.from_dict(raw_cut)
except TypeError:
    cut = PaddingCut.from_dict(raw_cut)
```
It shows that once an exception raised, `PaddingCut.from_dict()` would be called instead of `Cut.from_dict()`, which implies the type of `raw_cut` does not match `Cut` but matches `PaddingCut`.

lhotse/cut.py
```diff
@@ -195,12 +183,10 @@ def pad(self, desired_duration: Seconds) -> AnyCut:
         :param desired_duration: The cut's minimal duration after padding.
         :return: a padded MixedCut if desired_duration is greater than this cut's duration, otherwise self.
         """
-        if self.features is None:
-            # TODO: fill zeros for raw audio
-            return self
         if desired_duration <= self.duration:
             return self
         padding_duration = desired_duration - self.duration
+        assert self.features is not None
```
pzelasko, line 189: I think this is not needed - PaddingCut implements `load_audio()` that returns a vector of zeros. Audio padding should work just fine (unless I'm missing something?)

lhotse/cut.py
```diff
@@ -505,17 +491,27 @@ def load_features(self, root_dir: Optional[Pathlike] = None) -> np.ndarray:
             )
         return mixer.mixed_feats
 
-    def load_audio(self,
-                   recording_set: Optional[RecordingSet] = None,
-                   root_dir: Optional[Pathlike] = None,
-    ) -> List:
-        """Loads the audios of the source cuts and put them into a list."""
+    def load_unmixed_audio(self, root_dir: Optional[Pathlike] = None) -> List:
```
pzelasko, line 494: For consistency, it could return a numpy array with shape (n_channels, n_samples)

jimbozhang:
Sure, I'll add a notebook example this weekend.

Piotr Żelasko <notifications@github.com> 于 2020年7月24日周五 上午9:55写道：

lhotse/cut.py
jimbozhang: The lengths of the audio are different, so we have to pad them to put them into a matrix. Returning as List avoids padding. So I prefer the current implement ...

danpovey: To make sense of that, wouldn't you need to know the offsets?   That would be complex for the user, no?
I kind of feel like padding them to the same length would be the cleanest/most useful solution.

danpovey: .. and please clarify that this is before applying the SNRs?

lhotse/cut.py
```diff
@@ -477,7 +491,26 @@ def load_features(self, root_dir: Optional[Pathlike] = None) -> np.ndarray:
...
+        """Loads the unmixed audios of the source cuts and put them into a list."""
+        cuts = [track.cut for track in self.tracks]
+        audios = []
+        for cut in cuts:
+            audios.append(cut.load_audio(root_dir))
+        return audios
+
+    def load_audio(self, root_dir: Optional[Pathlike] = None) -> np.ndarray:
+        """Loads the audios of the source cuts and mix them on-the-fly.."""
```
danpovey, line 503: Please clarify the dim of the returned array.

lhotse/cut.py
danpovey: Mm.. I think it might be useful to have an option to have the SNRs applied already, as they would be
in 'load_audio' except keeping things in separate channels?  Is this complex to implement?

jimbozhang: I think there is another solution: just remove this `load_unmixed_audio()`, only retain the `load_unmixed_audio()` which returns the mixed audio in an `np.ndarray`, just like what `load_features()` does. This way can avoid users being confused. Not sure it this solution appropriate ...

danpovey:
Sure!

lhotse/audio.py
```diff
@@ -14,7 +14,7 @@
         import librosa
 import numpy as np
 
-from lhotse.utils import Pathlike, SetContainingAnything, Seconds, load_yaml, save_to_yaml
+from lhotse.utils import Pathlike, Decibels, SetContainingAnything, Seconds, load_yaml, save_to_yaml
```
csukuangfj, line 17: Should the import be sorted alphabetically?

danpovey:
Thanks for looking at it, Fangjun!

jimbozhang:
> One more question - could you run `AudioMixer` with different values of offset and snr in a notebook, visualizing the results (or listening to them since it's audio) just to double-check that it works as expected? I remember this helped me find some edge cases when implementing `FbankMixer` before, and we really have no way to make automated tests for these operations.

Ah, indeed there is a bug which I caught it when drawing the waveform. :joy: : 
Now this bug has been fixed, and the waveform looks correct:

https://colab.research.google.com/drive/1Mslc1jEKX6_kWwKu8F2fN3gXSme8OUTF?usp=sharing

lhotse/audio.py
jimbozhang: Is sorting by alphabetic really needed? @pzelasko

jtrmal:
Guys, I realize I'm a bit late to the game, but are you sure you absolutely
need this feature instead of just using same approach as in Kaldi, i.e.
input is a command/filter generating an audio format that has a header
well-defining what the input is?
y.

lhotse/cut.py
jimbozhang: Although `PaddingCut` implemented `load_audio()`, its properties {`num_frames`, `frame_shift`} are only for features. I think maybe we have to modify `PaddingCut` to support raw audio.

lhotse/audio.py
pzelasko: I don't have such strict requirements, as I actually almost never write the import lines myself - I use PyCharm's auto-import to import the relevant symbols, and then PyCharm's "optimize imports" so that it re-arranges them whatever way is appropriate. In general, I prefer not to burn too much energy on these kinds of things if it can be automated.

pzelasko:
@jimbozhang Looks good to me. You can actually easily listen to the audio in a notebook with this snippet, which I think works in colab as well:

```
from IPython.display import Audio
Audio(audio_array, rate=sampling_rate)
```

Other than that, could you also modify the `PaddingCut` as necessary so that the working-with-raw-audio feature is complete? Thanks!

@jtrmal thanks for chiming in, nice to see you here :) I'm not sure I understood what particular feature are you referring to. Could you elaborate?

lhotse/audio.py
pzelasko: Actually, feel free to reformat the code with https://github.com/psf/black if you like, I think it does re-arrange the imports too?

jimbozhang:
> @jimbozhang Looks good to me. You can actually easily listen to the audio in a notebook with this snippet, which I think works in colab as well:
> 
> ```
> from IPython.display import Audio
> Audio(audio_array, rate=sampling_rate)
> ```
> 
> Other than that, could you also modify the `PaddingCut` as necessary so that the working-with-raw-audio feature is complete? Thanks!

I updated the [notebook](https://colab.research.google.com/drive/1Mslc1jEKX6_kWwKu8F2fN3gXSme8OUTF?usp=sharing) on colab as you suggested. Very cool! 👍 
And I'll complete the modifications on `PaddingCut` in one or two days. 😃

lhotse/audio.py
csukuangfj: Consistency matters.

If @pzelasko  uses PyCharm to sort the import automatically, which follows the alphabetic order, then the remaining
code should also follow this style.

The are various style guides recommending an alphabetic import order.

Please refer to https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement

pzelasko: There is a nice link there for a utility that does this automatically: https://pypi.org/project/isort/

I'm considering writing a dev-helper script at some point that runs stuff like this + black + mypy to address consistency issues.

jimbozhang: PR #58.

jtrmal:
Well, my comment might not make sense, given you didn't understand :)
But in Kaldi, we do not support any other formats than just 16bit PCM wav.
That doesn't mean that we cannot work with other formats, though -- thanks
to the  ability to specify a command that should be executed and it's
output read as the input file, the range of formats we can work with is
virtually limitless...
So my suggestion would be do it that way instead of implementing your own
raw audio stuff -- it can become a complex task when you move to 16 bit and
up, due to endianness, signedness, multiple microphones (perhaps)...

pzelasko:
@jtrmal OK I think I get it now. This PR is not about audio loading - we actually do support a superset of what Kaldi does, i.e. WAV files + shell command pipes + many other formats that librosa/libsoundfile provide support for out of the box. This PR is mostly about mixing audio from multiple cuts on the fly and introducing an example PyTorch Dataset class that shows how you can work with audio samples (instead of extracted features). I can imagine the name of the PR could be misleading in that regard ;)

jtrmal:
thanks for the explanation, that makes sense as a feature...
y.

PR #57, opened 2020-07-22
add pychain support (WIP)

YiwenShaoStephen:
@pzelasko I drafted some codes to show how lhotse would probably support pychain. 
Things are not fixed yet and this PR is meant to show some basic ideas in my mind that might help you decide how/whether to make pychain support.
Something that is not included in this PR is how denomintor fst path can be added as an attribute to the dataset.

### DIFF
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -8,9 +8,6 @@ from torch.utils.data import Dataset
 from lhotse.cut import CutSet
 from lhotse.utils import Pathlike
 
-# TODO: make them optional
-import simplefst
-from pychain import ChainGraph, ChainGraphBatch
 EPS = 1e-8
 
 
@@ -38,16 +35,6 @@ class SpeechRecognitionDataset(Dataset):
         self.cuts = cuts.cuts
         self.root_dir = Path(root_dir) if root_dir else None
         self.cut_ids = list(self.cuts.keys())
-        if cuts[0].supervisions[0].graph_archive is not None:
-            self.train = True
-        else:
-            self.train = False
-
-    def parse_rxfile(file):
-        # separate offset from filename
-        if re.search(':[0-9]+$', file):
-            (file, offset) = file.rsplit(':', 1)
-        return file, int(offset)
 
     def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
         cut_id = self.cut_ids[idx]
@@ -58,17 +45,14 @@ class SpeechRecognitionDataset(Dataset):
         if len(cut.supervisions) > 1:
             logging.warning(
                 "More than one supervision in ASR task! Selected the first one and ignored others.")
-
-        if self.train:
-            graph_archive = cut.supervisions[0]
-            filename, offset = parse_rxfile(graph_archive)
-            fst = simplefst.StdVectorFst.read_ark(filename, offset)
-            graph = ChainGraph(fst)
+        supervision = cut.supervisions[0]
+        if supervision.pychain_graph is not None:
+            pychain_graph = supervision.pychain_graph.load()
 
         return {
             'features': features,
-            'text': cut.supervisions[0].text
-            'graph': graph if self.train else None
+            'text': supervision.text
+            'pychain_graph': pychain_graph if supervision.pychain_graph is not None
         }
 
     def __len__(self) -> int:
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -8,6 +8,9 @@ from torch.utils.data import Dataset
 from lhotse.cut import CutSet
 from lhotse.utils import Pathlike
 
+# TODO: make them optional
+import simplefst
+from pychain import ChainGraph, ChainGraphBatch
 EPS = 1e-8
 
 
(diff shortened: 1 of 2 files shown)


lhotse/supervision.py
```diff
@@ -15,6 +15,8 @@ class SupervisionSegment:
     language: Optional[str] = None
     speaker: Optional[str] = None
     gender: Optional[str] = None
+    # exmaple: "data/graph/train_si284/num.1.ark:9"
+    graph_archive: Optional[str] = None
```
pzelasko, line 19: I think we should keep the graph loading logic more self-contained. How about: 

```
pychain_graph: Optional[PyChainGraph] 
```

and:

```
@dataclass
class PyChainGraph:
    archive_path: Pathlike
    archive_offset: int

    def from_dict(data: dict) -> 'PyChainGraph':
        ...  # look up other from_dict implementations across the codebase
    
    def load():
        import simplefst
        from pychain import ChainGraph
        # loads and returns a chain graph; 
        # local import ensures that only the users who actually need it will need to have pychain/simplefst installed
```

you'll need to update `SupervisionSegment.from_dict` to make sure that dict is converted to this class during deserialization

lhotse/dataset/speech_recognition.py
```diff
@@ -8,6 +8,9 @@
 from lhotse.cut import CutSet
 from lhotse.utils import Pathlike
 
+# TODO: make them optional
+import simplefst
+from pychain import ChainGraph, ChainGraphBatch
```
pzelasko, line 13: the easiest way to make this optional is to do a local import (i.e. import in a function with a very limited scope)

lhotse/dataset/speech_recognition.py
```diff
@@ -35,21 +38,134 @@ def __init__(
         self.cuts = cuts.cuts
         self.root_dir = Path(root_dir) if root_dir else None
         self.cut_ids = list(self.cuts.keys())
+        if cuts[0].supervisions[0].graph_archive is not None:
+            self.train = True
+        else:
+            self.train = False
+
+    def parse_rxfile(file):
```
pzelasko, line 46: this won't be needed with `PyChainGraph` class suggested earlier

lhotse/dataset/speech_recognition.py
```diff
@@ -35,21 +38,134 @@ def __init__(
...
+    As the default pytorch dataloader can only batchify tensor-like 
+    (e.g. numpy array or pytorch tensor) items from the __getitem__ function but we have a dictionary output in
+    SpeechRecognitionDataset so that we need to give customized dataloader here (another option is to leave it to
+    the users themselves to decide how to load data). All these usages below are quite basic and flexible and can
+    be replaced/re-writtern by other functions. The only special thing here is to use ChainGraphBatch class to 
+    batchify a list of ChainGraph.
+"""
+
+
+def _collate_fn_train(batch):
```
pzelasko, line 90: I am not very familiar with PyChain graphs design so I will need some more input from you to understand it better. First, let me share what my original plan is, and let's figure out if it makes sense to follow that and if that's possible to do with PyChain. 

I'd like the `CutSet` to be the only data structure where we "mutate" the data - i.e. trim, mix, pad, and so on. These operations should lead the user to one of the following scenarios:
1) all "cuts" are of the same duration (e.g. all cuts are 10 seconds long);
2) the "cuts" are in buckets of the same duration (e.g. cuts are exactly either 5s, 10s, or 20s long).

Then, the user passes that `CutSet` to a `*Dataset` class that only needs to call `load_features`/`load_audio` to get the inputs of the requested duration. It's also the `*Dataset` class that needs to take the supervisions present in a given cut and create a meaningful representation suitable for training.

I'd like the CutSet to be able to cut not only the features/audio, but also the supervisions, like the `ChainGraph`. It already adjusts the supervision time bounds. I'm wondering whether (and how) we can trim/pad PyChain graph supervisions. I think you're already doing that in the `ChainGraphBatch` class though. I'm thinking, it would be ideal if, during the padding/trimming of the `Cut`, we could insert the information about the graph offset and the number of graph states to be loaded in the supervision object. An example:

1) an utterance (`SupervisionSegment`) originally has 50 seconds (super long utterance)
2) we decided to cut a 20 seconds long cut starting at second 10 (creating `Cut`s with `SupervisionSegment` that is only 20s long)
3) while cutting, we put the information in the `PyChainGraph` inside the `SupervisionSegment` object to load the graph only from state 1000 until the state 3000 (assuming 1 state == 10ms)
4) in the `SpeechRecognitionDataset`, we call `Cut.load_features()` to get 20s of features, and `supervision.pychain_graph.load()` to get 20s of graph supervision

The collate_fn also gets simpler, as the only thing it needs to do is put all the graphs into a `ChainGraphBatch` (no padding, no trimming), and maybe it can delegate all the other work into the default collate_fn in PyTorch.

lhotse/supervision.py
pzelasko: We might also need fields like `state_offset: int` and `num_states_to_load: int` (somewhat equivalent to `begin: Seconds` and `duration: Seconds`)

danpovey:
I don't think the supervision objects are really cuttable.. at least not if we're using the e2e style supervision objects,
that don't have alignment info, which I think Yiwen is (?)
If it's using the e2e supervision objects, I think it would make most sense to pad things to one a set of standard lengths.
Perhaps could pad with silence but add some kind of very quiet noise, e.g. from what David released
(forget its name) to make the discontinuity less noticeable?

danpovey:
PyChain/Kaldi isn't set up to use different durations in the same minibatch, that functionality was for k2.  So for this case
we'd just assume the supervision covers the entire cut, including any padding.

lhotse/supervision.py
YiwenShaoStephen: Sounds good to me!

YiwenShaoStephen:
Yes. As @danpovey pointed out, PyChain is designed for end-to-end LF-MMI where each sample is an entire utterance. To form a minibatch, I do zero-padding on the features and also record their original lengths before padding. These padded features won't be calculated in the loss function (also not calculated in networks if they are RNNs) so I think we don't need to bother too much on how we pad the inputs.

pzelasko:
Alright, got it. I thought the PyChain graphs represent alignments and could be cuttable. If they are e2e-style graphs, I think the current `Cut` padding functionality is sufficient. So the typical PyChain ASR workflow would look like this:

0. (external step?) Create PyChain graphs archive.
1. Create RecordingSet and SupervisionSet manifests.
2. Extract features, creating FeatureSet.
3. Convert supervision segments to cuts.
4. Pad cuts (either using `cut_set.pad()` to do low-energy padding or mix in noises to achieve the desired duration).
5. Load the padded CutSet into `SpeechRecognitionDataset` (as we go forward we'll likely create a hierarchy of ASR dataset classes: for PyChain, K2, and maybe other frameworks).

For padding, you can refer to this short example in the docs: https://lhotse.readthedocs.io/en/latest/cuts.html#python  
I think maybe we should create an explicit section about the padding going forward.

I think there remains a question of whether we should return a mask or a tensor of sequence lengths (number of non-padded frames) for each sample in the batch. That'd have to be done in the `Dataset` class. It's probably useful if we pad with zeros, and not useful if we pad with noise.

lhotse/supervision.py
pzelasko: Given our discussion, the `state_offset` and `num_states_to_load` shouldn't be need after all.

YiwenShaoStephen:
Hi, I made a new commit to make pychain_graph self contained as you suggested. Here are something that I'm not sure where to put so I'd like to know how you think about it:

1. Supervision Preparation. Take minilibrispeech as an example, my current thought is to add some rows near https://github.com/pzelasko/lhotse/blob/master/lhotse/recipes/mini_librispeech.py#L99. The use case will be that, after running some graph preparation bash codes outside lhotse, I will have a scp file with each line of format "uttid archive_path:offset". I'm wondering what would be the best way to load this file within ```prepare_mini_librispeech``` function.
2. Dataloader. In current lhotse's codebase, it seems there isn't anything about dataloader. I guess probably you decide to leave this part to users themselves as it will be quite different from case to case. Personally I would also suggest to do so. Just to double check this point with you in case you have other plans.

pzelasko:
Hey Yiwen. 

Regarding supervisions, I think it might be better to have a separation between preparing the manifests for a corpus, and preparing/adding the pychain graph information for them. For example if you were writing a script, I could see it going like that:

```
# create recording and supervision manifets
lhotse prepare mini_librispeech path/to/libri/ data/

# create pychain graphs
<some external script(s) creating the FSTs>

# add pychain graph supervisions - "supervisions.yml" is the input produced by the prev script, 
# and "supervision_pychain.yml" is the output containing with added paths to 
# pychain graphs in each supervision.
lhotse pychain make-graphs data/supervisions.yml path/to/pychain/fsts data/supervisions_pychain.yml
```

Ideally, there would be a function in Lhotse that can convert any text supervision to a pychain graph, but I don't know how much effort is needed to do that.

Regarding dataloaders - I'm striving to make it possible for the users to truncate/pad/otherwise transform the data using CutSet so that a custom DataLoader is not needed. I think this works currently for the source separation task and should easily work for VAD. However, I can see that PyChain would require some custom handling of the graphs. I think it should be possible by just providing a custom collate_fn though? I'm fine with providing a custom DataLoader if needed. 

Eventually I'd like to have a Dataset/DataLoader with a custom Sampler that assumes the Dataset is bucketed into a number of duration bins (5s, 10s, 15s) and samples equal duration samples to construct a batch, making it somewhat more flexible.

danpovey:
Mm, this integration with PyChain is non-ideal because PyChain, at least as
Yiwen is using it, assumes you've run Kaldi scripts first,
and the data prep isn't available from python.
For now I'd be OK with anything that runs though, especially if it were
segregated somehow, e.g. in a separate file.
Perhaps Yiwen was right that there isn't much value in using Lhotse given
how PyChain currently works.  What would be better is if
we could make it so PyChain could prepare the supervision objects for
arbitrary data taken from Lhotse itself.  Yiwen, what would be
necessary in order to make that possible?

YiwenShaoStephen:
Thanks for the comment! 
@pzelasko I'm a little confused about this line: ```lhotse pychain add-graphs data/supervisions.yml path/to/pychain/fsts data/supervisions_pychain.yml``` Do I need to add any codes to enable this line? Or it has been available already?
@danpovey Yes, that's why I hesitated a lot about what to expect for this work. As for non-kaldi users, the biggest challenge of data preparation is in graph generation. I'm currently using kaldi scripts and binaries to do so. 
If we want to make it possible in python/lhotse, we will need FST-alike operations that help us build numerator/denominator graphs. Others like preparing lang/dictionary/lexicon will also be needed but easier than FST stuff. Personally, l think these things won't be a small deal but still would not necessarily reduce the complexity to users (and we will still go back to kaldi for decoding), I just put these parts remained in kaldi.
Anyway, for now, I think this work can be currently used as a test case for lhotse. TBH, I think probably it would be even better if this PR won't be merged into lhotse if it breaks some design patterns/motivation and confuses the users.

pzelasko:
Yes, the code for that line to work would have to be written. It's just a design proposal.

I think handling lexicons and graph preparation should be a part of Lhotse, as it's essentially a part of data preparation for ASR. I can see that being a separate module that modifies supervisions/cuts. I don't mind having optional dependencies on OpenFST or other graph libraries like K2. 

I leave the scope of integration between Lhotse and PyChain up to you and Dan - I don't know how much work you'd like to spend on this. Both deep and shallow integrations are fine with me, and we can figure out how to merge that in later. Probably a shallow integration (pre made graphs) is a good starting point anyway and a reference point for deeper integrations later on if we want to pursue that.

danpovey:
For the k2 integration we would definitely need some tools for handling
graphs, etc.  I wasn't necessarily
planning to make those part of lhotse itself, but we can see what makes
sense.
For now I'm OK with shallow integration I guess (it's just a shame because
it doesn't really allow us to exercise many of the
features of lhotse).  But we may be able to do extra data augmentation with
it.

pzelasko:
> For now I'm OK with shallow integration I guess (it's just a shame because it doesn't really allow us to exercise many of the features of lhotse).

Same thoughts here.

pzelasko:
I'm closing this PR since we're not actively working on that anymore. Feel free to pick it up in the future if you'd like.

PR #61, opened 2020-08-01
Time-domain data augmentation using WavAugment

pzelasko:
Building and applying the data augmentation effect chain in Python is quite straightforward. I am more concerned about the CLI - I don't think I want to end up having an "augmentation manifest" that's basically a copy-paste of all of WavAugment's effects. IIRC in Kaldi, it was sufficient to have a small number of augmentation setups with parameters constant across recipes (i.e. MUSAN/reverb/speed-perturbation/volume or any combination of those). @danpovey is it fine if we have a number of pre-defined effect chains, that will be named and selectable from the CLI? Python API users will get extra flexibility if they want.


danpovey:
Yeah I think what you propose is reasonable.

pzelasko:
For some reason, the CI actions disappeared, but I can't see any obvious error in the GitHub Actions YAML. I'll try to have a look later.

I think this is ready to review and merge. Note that #55 should get merged first - I based this PR on #55 since I'm enabling the augmentations in feature extraction script.

pzelasko:
Oh I know - the CI checks are not being run because the target branch is not `master` - I temporarily changed it so that the diff doesn't include changes merged from #55 into this branch.

examples/mini_librispeech/mini_librispeech.ipynb
```diff
@@ -68,6 +69,26 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "use_data_augmentation = False\n",
+    "augmenter = WavAugmenter.create_predefined('pitch_reverb_tdrop', sampling_rate=16000) if use_data_augmentation else None\n",
+    "# It seems when spawning multiple Python subprocesses with the same sox handle it raises \"std::runtime_error: Couldn't close file\"\n",
+    "# For now, work around \n",
```
danpovey, line 88: perhaps could raise a bug with the sox package maintainer?

lhotse/augmentation.py
```diff
@@ -0,0 +1,153 @@
...
+    https://github.com/facebookresearch/WavAugment/blob/master/examples/python/librispeech_selfsupervised.py#L152
+    https://arxiv.org/abs/2007.00991
+
+    :param sampling_rate: a sampling rate value for which the effect will be created (resampling is needed for pitch).
+    """
+
+    import augment
+    effect_chain = augment.EffectChain()
+    # The pitch effect changes the sampling ratio; we have to compensate for that.
+    # Here, we specify 'quick' options on both pitch and rate effects, to speed up things
```
danpovey, line 138: would be nice to make clear the meaning of these params, e.g. what is the -300:300?

test/dataset/test_unsupervised_dataset.py
```diff
@@ -47,3 +48,14 @@ def test_on_the_fly_feature_extraction_unsupervised_dataset(libri_cut_set):
     # E              [-13.2, -11.2,  -9.6, ...,  -5.6,  -6.5,  -7.6],
     # E              [-12. , -10.1, -10.1, ...,  -5.8,  -7. ,  -7.8],...
     np.testing.assert_array_almost_equal(ref_feats, tested_feats, decimal=0)
```
danpovey, line 50: again, not so sure about this decimal=0 thing.

danpovey:
Small comments, but generally looks great!  Same for the other one...

examples/mini_librispeech/mini_librispeech.ipynb
pzelasko: Actually, it seems to be specific to my jupyter lab setup; it works well in the command line, and in Google colab using a modified official WavAugment example notebook... If I come up with a way to make this reproducible, I'll raise it with them.

test/dataset/test_unsupervised_dataset.py
pzelasko: In case somebody stumbles upon this in the future - we discussed and analyzed this in #55

Issue #60, opened 2020-08-01
Voice activity detection (VAD) dataset


pzelasko:

pzelasko:
@jimbozhang would you like to take this one?

jimbozhang:
Sure, I'd be glad to do this :smile_cat: 
If I didn't misunderstand it, this issue requires to add an example for VAD task.
Is there any recommended opensource data-set? I didn't find any on OpenSLR ...

pzelasko:
Hmm, I think any ASR dataset that has speech and silence in it would be sufficient. Maybe AMI would be ok? I never worked with it, but these are meetings, so it should have continuous recordings. For VAD dataset, you'd discard the text in the supervisions and output a tensor of 0s for silence frames and 1s for speech frames. I think some setups in Kaldi also had 2 for noise frames, but IDK if that's necessary/useful? Maybe @danpovey or @jtrmal would know more.

AMI download: http://groups.inf.ed.ac.uk/ami/download/

danpovey:
MM, my tendency would be to just have no supervision when training the
noise frames, i.e. to make it a don't-care.

pzelasko:
But in that case it's also undefined what the system predicts during
inference in these frames, right? Maybe it's better to have only
speech/nonspeech?

sob., 1 sie 2020, 05:33 użytkownik Daniel Povey <notifications@github.com>
napisał:

danpovey:
Well, yes, its undefined, but those frames could in practice contain either
speech or nonspeech, which is why you don't really want to train on them
as it would confuse the model.

pzelasko:
Fair enough. I guess how you wish to treat the noisy/uncertain frames is dependent on a particular application.

jimbozhang:
So we have decided to choose AMI. l trend to illustrate both ASR and VAD in the `ami` task.

danpovey:
Great!

PR #64, opened 2020-08-13
ASR and VAD recipes for the AMI corpus

jimbozhang:
For issue #60.
I'll add the corresponding bash examples after this PR be ready to merge.

A drawback is that it do not support "cross-cut history", that means a sequence model (e.g. RNN, TDNN) can not obtain the features of the previous segment as its history. I haven't found an elegant way to solve this problem ...


lhotse/dataset/vad.py
```diff
@@ -0,0 +1,55 @@
...
+class VadDataset(Dataset):
+    """
+    The PyTorch Dataset for the voice activity detection task.
+    Returns a dict of:
+
+    .. code-block::
+
+        {
+            'features': (T x F) tensor
+            'is_voice': bool
```
pzelasko, line 23: I think this should be a 1-dim tensor of 0/1 values for each feature frame. Let's say you have a `Cut` of duration 10 seconds, with a supervision between second 3 and 7 - then you would return sth like `torch.tensor(300 * [0] + 400 * [1] + 300 * [0])` with frame_shift = 10ms

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,55 @@
...
+from pathlib import Path
+from typing import Dict, Optional
+
+import torch
+from torch.utils.data import Dataset
+
+from lhotse.cut import CutSet
+from lhotse.utils import Pathlike
+
+EPS = 1e-8
```
pzelasko, line 11: unused variable

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,55 @@
...
+        self.root_dir = Path(root_dir) if root_dir else None
+        self.cut_ids = list(self.cuts.keys())
+
+    def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
+        cut_id = self.cut_ids[idx]
+        cut = self.cuts[cut_id]
+
+        features = torch.from_numpy(cut.load_features(root_dir=self.root_dir))
+
+        # We assume there is only one supervision for each cut in VAD tasks (for now)
```
pzelasko, line 45: I think we don't need that assumption if we output a 1-d supervision tensor

requirements.txt
```diff
@@ -1,10 +1,14 @@
-pyyaml>=5.3.1
-numpy>=1.18.1
-numba==0.48.0  # Fix to librosa not respecting something deprecated
-librosa==0.7.2
 click>=7.1.1
+cytoolz>=0.10.1
+lhotse~=0.1
+librosa==0.7.2
 lilcom>=1.1.0
+matplotlib~=3.2.2
```
pzelasko, line 6: The new dependencies added here shouldn't be necessary - you can do `pip install -e .[dev]` to install both lhotse and all of its dependencies. Pip supports installing different configurations, e.g. `pip install lhotse` or `pip install lhotse[dev]` (the dot `.` in the previous cmd corresponds to `lhotse`, it just tells pip to install from local sources). I recently ordered the dependencies to clearly separate what is needed by the end users and what is needed by the devs. If you install the `dev` version, nothing will be missing.

lhotse/recipes/ami.py
```diff
@@ -0,0 +1,256 @@
...
+    import torchaudio
+
+from lhotse.audio import AudioSource, Recording, RecordingSet
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike
+
+# TODO: support other microphone settings like "sdm1" or "mdm8"
+mic = 'ihm'
+
+# The splits are same with Kaldi's AMI recipe
```
pzelasko, line 22: 👍

lhotse/recipes/ami.py
```diff
@@ -0,0 +1,256 @@
...
+    text_path = target_dir / text_name
+    if force_download or not text_path.is_file():
+        text_url = f'{url}/AMICorpusAnnotations/ami_manual_annotations_v1.6.1_export.gzip'
+        urllib.request.urlretrieve(text_url, filename=text_path)
+
+
+class AmiSegmentAnnotation(NamedTuple):
+    text: str
+    begin_time: float
+    end_time: float
```
pzelasko, line 82: Can you use the `Seconds` type from `lhotse.utils` (assuming it's in seconds)? Also, for consistency with other classes like `Cut`, could you name these `start` and `end`?

pzelasko: Actually I think you don't need this class. You can create `lhotse.supervision.SupervisionSegment` directly, since you have all of: meeting id (can be same as recording id), channel, start, end and text. Other fields, like speaker, can be updated later if this information is not available at this stage.

lhotse/recipes/ami.py
```diff
@@ -0,0 +1,256 @@
...
+    manifests = defaultdict(dict)
+
+    data_dir = Path(data_dir)
+    output_dir = Path(output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    metadata = make_metadata(data_dir, asr_task)
+    for part in dataset_parts:
+        # Audio
+        audio = RecordingSet.from_recordings(
```
pzelasko, line 214: Hmm, this doesn't seem right to me. The idea behind `Recording` is that even if a wave file is 10 hours long, it's a single recording. It just doesn't care about the segmentation (like `wav.scp` in Kaldi). So for each audio file, you should have one `Recording`, and many different `SupervisionSegment`s, that refer to this recording via `recording_id`.

Then, you don't need the `offset_seconds` field that you added to `Recording`. The `FeatureSetBuilder` is going to extract the features for the whole duration of the recording. After that, you can create the `CutSet` with the method named like `make_windowed_cuts_...()` with an even duration of e.g. 10 seconds. I don't remember if it handles supervisions, but if not, then you'd make it preserve the supervisions (just cut their time boundaries to not exceed those of the cut, I think this is already implemented when you call `cut.truncate()`). As a result, you'll get cuts that have some frames with speech and some frames without, with that information encoded by the presence of `SupervisionSegment`s. 

If you do it this way, you also won't need the `asr_task` flag.

pzelasko: Actually the best way to implement the windowing would be to add a new method that creates a cut set, that takes a feature set and a supervision set, and creates one cut per each feature (not each supervision, like the current one) ((that method could actually be instead a flag to the `from_manifests()`, maybe called `cut_on_supervisions: bool = True`)). Then, to create evenly sized cuts, you'd just iteratively call `cut.truncate(offset, offset + duration, keep_excessive_supervisions=True)`, and it's going to do the right thing with the supervisions for you. This could maybe be another method on `CutSet`, named e.g. `split_cuts`?

lhotse/dataset/vad.py
pzelasko: Essentially, instead of a "sequence classification" task, we want a "sequence tagging" or "frame classification" task, which is how VAD would usually be used.

pzelasko:
Regarding the features of a previous cut - I think if we convert this to "frame classification", then we can just make large cuts, say one minute or more, that have a lot of context. Otherwise, if cross-cut information passing is needed, that would require another dataset class that's prepared to support that (kind of like special data preparation was needed for Keras' `RNN(..., stateful=True)`).

lhotse/recipes/ami.py
danpovey: Piotr, I don't quite understand what you are saying here, e.g. "creates one cut per each feature".

danpovey: ... also, my recollection is that AMI is quite complicated w.r.t. microphones/channels, e.g. we have headset microphones and multiple distant microphones, and in Kaldi we have several different ways of dealing with that.  Can someone please clarify the approach that is being taken here?  This should probably be documented, e.g. in a docstring at the top of the file or wherever is appropriate.

lhotse/recipes/ami.py
```diff
@@ -0,0 +1,256 @@
...
+
+    return metadata
+
+
+def prepare_ami(
+        data_dir: Pathlike,
+        output_dir: Pathlike,
+        asr_task: Optional[bool] = True,
+        write_yaml: Optional[bool] = False,
+) -> Dict[str, Dict[str, Union[RecordingSet, SupervisionSet]]]:
```
danpovey, line 204: I think this needs a bit of documentation since it's not clear what the structure of the output is, e.g. what do the string-values keys correspond to?

lhotse/recipes/librimix.py
```diff
@@ -1,6 +1,6 @@
 import shutil
```
danpovey, line 1: This is a pre-existing issues, but I think all of these recipes should have some kind of file-level docstring to give some context.  The reader may not know, for instance, that librimix is a dataset based on Librispeech for speech separation.

lhotse/recipes/ami.py
pzelasko: I agree with the comment about AMI. Regarding the "cut per each feature", please check #65 - I drafted the changes that I previously described and explained a bit more.

lhotse/recipes/librimix.py
pzelasko: Great point!

pzelasko:
@jimbozhang check out PR #65 that I just merged - I added new features that will allow you to easily create cuts as windows over the full recording/feature matrix.

jimbozhang:
> @jimbozhang check out PR #65 that I just merged - I added new features that will allow you to easily create cuts as windows over the full recording/feature matrix.

Thanks @pzelasko , I'll check them and resolve the reviews of this PR soon.

jimbozhang:
The work of @pzelasko  makes writing the AMI recipe quite easier :thumbsup:
I made a commit to use the new method of `Cutset` from PR #65.

danpovey:
Fantastic!!

examples/ami/ami.ipynb
```diff
@@ -0,0 +1,201 @@
+{
```
pzelasko, line 1: Can you commit the notebook after executing "Run all"? That way, it will serve as extra documentation.

lhotse/cut.py
```diff
@@ -169,7 +169,8 @@ def truncate(
             # causing trouble for users expecting matching audio tensor shapes for cuts truncated to a single duration.
             new_duration = duration
         assert new_duration > 0.0
-        assert new_start + new_duration <= self.start + self.duration + 1e-5
+        if new_start + new_duration > self.start + self.duration:
+            new_duration = self.start + self.duration - new_start
```
pzelasko, line 173: Hmm, I'm not sure about this one. I originally intended to "force" the user to provide the offset and duration arguments that "physically make sense", i.e. won't ask `truncate` to exceed the Cut's bounds. You're proposing to let it pass by "correcting" the duration value that the user had provided. It seems a bit dangerous and error-prone. @danpovey what do you think? 

At the very least I would move the `assert new_duration > 0.0` after this `if` statement and add a unit test for it that demonstrates the condition when it occurs and the expected behavior.

lhotse/dataset/speech_recognition.py
```diff
@@ -32,7 +32,7 @@ def __init__(
             root_dir: Optional[Pathlike] = None
     ):
         super().__init__()
-        self.cuts = cuts.cuts
+        self.cuts = cuts.trim_to_supervisions().cuts
```
pzelasko, line 35:
```suggestion
        self.cuts = cuts.trim_to_supervisions()
```

I think the `.cuts` at the end is not needed - `CutSet` works almost like a dict already.

lhotse/dataset/speech_recognition.py
```diff
@@ -42,9 +42,8 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
 
         features = torch.from_numpy(cut.load_features(root_dir=self.root_dir))
 
-        # We assume there is only one supervision for each cut in ASR tasks (for now)
-        if len(cut.supervisions) > 1:
-            logging.warning("More than one supervision in ASR task! Selected the first one and ignored others.")
+        # There must be only one supervision because we have had trim_to_supervisions() processed
+        assert len(cut.supervisions) == 1
```
pzelasko, line 46:
```suggestion
        assert len(cut.supervisions) == 1, "SpeechRecognitionDataset does not support overlapping supervisions yet."
```

lhotse/dataset/vad.py
pzelasko: still unused :)

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,61 @@
...
+import torch
+from torch.utils.data import Dataset
+
+from lhotse.cut import CutSet
+from lhotse.utils import Pathlike, Seconds
+
+EPS = 1e-8
+
+
+class VadDataset(Dataset):
```
pzelasko, line 15: Can we have a unit test that builds the `VadDataset` on a single-element `CutSet`, obtains `dataset[0]` and checks that the `is_voice` field contains correct supervision (e.g. 0-2s silence, 2-5s speech)?

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+
+    def __init__(
+            self,
+            cuts: CutSet,
+            duration: Optional[Seconds] = 5.0,
+            root_dir: Optional[Pathlike] = None
+    ):
+        super().__init__()
+        self.cuts = cuts.cut_into_windows(duration, keep_excessive_supervisions=True).cuts
```
pzelasko, line 37:
```suggestion
        self.cuts = cuts.cut_into_windows(duration, keep_excessive_supervisions=True).filter(lambda cut: cut.duration == duration)
```

We need the filter step at the end to remove the last window for each recording which might be shorter (maybe `cut_into_windows` should do it automatically? I didn't want to impose my own preference on the users...)

pzelasko: BTW another option is to call `.pad(duration)` instead to retain these windows.

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,61 @@
...
+        self.root_dir = Path(root_dir) if root_dir else None
+        self.cut_ids = list(self.cuts.keys())
```
pzelasko, line 39:
```suggestion
        self.cut_ids = list(self.cuts.cuts.keys())
```

(if we retain a CutSet instead of dict in `self.cuts` - although it reads a bit awkward, we probably need to add `.ids` property to the CutSet)

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,61 @@
...
+    def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
+        cut_id = self.cut_ids[idx]
+        cut = self.cuts[cut_id]
+
+        features = torch.from_numpy(cut.load_features(root_dir=self.root_dir))
+        assert features.shape[0] == cut.num_frames
+        assert isclose(cut.num_frames * cut.frame_shift, cut.duration)
+
+        is_voice = np.zeros(cut.num_frames)
+        for subversion in cut.supervisions:
```
pzelasko, line 50: subversion -> supervision

lhotse/dataset/vad.py
pzelasko: (actually, it's best if that cut has two supervisions and silence between them - that tests the supervision iteration in __getitem__)

lhotse/dataset/vad.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
+        cut_id = self.cut_ids[idx]
+        cut = self.cuts[cut_id]
+
+        features = torch.from_numpy(cut.load_features(root_dir=self.root_dir))
+        assert features.shape[0] == cut.num_frames
+        assert isclose(cut.num_frames * cut.frame_shift, cut.duration)
+
+        is_voice = np.zeros(cut.num_frames)
```
pzelasko, line 49: Probably a technicality but at this point it makes sense to use `torch.zeros` and skip `from_numpy` later on.

lhotse/recipes/ami.py
pzelasko: Could you add a docstring that describes what is exactly returned by this function?

requirements.txt
```diff
@@ -1,10 +1,13 @@
...
+cytoolz>=0.10.1
+librosa==0.7.2
 lilcom>=1.1.0
+matplotlib~=3.2.2
+numba==0.48.0  # Fix to librosa not respecting something deprecated
+numpy>=1.18.1
+pandas>=1.0.3
+pytest~=5.4.3
+pyyaml>=5.3.1
+setuptools~=47.3.1
```
pzelasko, line 11: I see there's still a couple of extra dependencies that shouldn't be here - can you revert the changes to this file altogether?

lhotse/dataset/vad.py
jimbozhang: :sweat_smile:

requirements.txt
jimbozhang: Ah sorry, I just wanted to sort them and failed to check them carefully. I'll remove the unnecessary items.

lhotse/dataset/vad.py
jimbozhang: :thumbsup:

lhotse/dataset/speech_recognition.py
jimbozhang: I think this message will never be printed because `trim_to_supervisions()` makes sure that there would be only one supervision. So can we set this message as: "Unexpected multi-supervisions cut. Maybe cuts.trim_to_supervisions() was not work correctly."

lhotse/recipes/ami.py
jimbozhang: I'll add the docstring here.

danpovey:
Thanks!  (was not work -> did not work)

jimbozhang:
Now I'm sure there are some serious bugs in this PR. I'll fix them and update in one or two days.

danpovey:
Thanks!!

lhotse/cut.py
jimbozhang: Maybe we can just return `None` when the current Cut was not long enough? Like what the commit https://github.com/pzelasko/lhotse/pull/64/commits/f5c166020d92ced2a0928fcd271067ddd73b4e57 does.

lhotse/dataset/vad.py
jimbozhang: We can exclude them in `lhotse/cut.py`: https://github.com/pzelasko/lhotse/pull/64/commits/f5c166020d92ced2a0928fcd271067ddd73b4e57

test/dataset/test_vad_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+from lhotse.cut import CutSet
+from lhotse.dataset.vad import VadDataset
+
+
+@pytest.fixture
+def cut_set():
+    return CutSet.from_yaml('test/fixtures/ami/cuts.yml')
+
+
+def test_unsupervised_dataset(cut_set):
```
pzelasko, line 16:
```suggestion
def test_vad_dataset(cut_set):
```

test/dataset/test_vad_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+    duration = 5.0
+    feature_threhold = -8.0
+
+    cuts = list(cut_set.cuts.values())
+    assert duration < cuts[0].duration < 2 * duration
```
pzelasko, line 21: I don't understand the intention of this assertion - why is it testing for duration of range [5, 10]?

test/dataset/test_vad_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.fixture
+def cut_set():
+    return CutSet.from_yaml('test/fixtures/ami/cuts.yml')
+
+
+def test_unsupervised_dataset(cut_set):
+    duration = 5.0
+    feature_threhold = -8.0
+
+    cuts = list(cut_set.cuts.values())
```
pzelasko, line 20:
```suggestion
    cuts = list(cut_set)
```
Just syntactic sugar - it's possible because `CutSet` defines `__iter__` that returns `.cuts.values()`

lhotse/cut.py
```diff
@@ -158,7 +158,7 @@ def truncate(
         :param keep_excessive_supervisions: bool. Since trimming may happen inside a SupervisionSegment,
             the caller has an option to either keep or discard such supervisions.
         :param preserve_id: bool. Should the truncated cut keep the same ID or get a new, random one.
-        :return: a new Cut instance.
+        :return: a new Cut instance. If the current Cut is shorter than the duration, return None.
```
pzelasko, line 161: Ah, I understand the intent now. Looks good 👍

lhotse/recipes/ami.py
```diff
@@ -0,0 +1,206 @@
+import logging
```
pzelasko, line 1: Could you add a top-level docstring explaining what's the AMI corpus, as Dan suggested? It's fine to copy some existing description (e.g. from Kaldi or AMI download website). I'll make a separate PR adding these descriptions for LibriMix and LibriSpeech.

examples/ami/ami.ipynb
```diff
@@ -0,0 +1,264 @@
...
+ "cells": [
```
pzelasko, line 2: The last cell rendered in this notebook doesn't look right - the VAD label says the first 1/3rd is speech and the rest is silence, whereas the spectrogram suggests something else. 

See: https://github.com/pzelasko/lhotse/blob/b55e349d40ed88ecb2ad411ba0002405af4845d0/examples/ami/ami.ipynb

lhotse/recipes/ami.py
jimbozhang: Sure.

test/dataset/test_vad_dataset.py
jimbozhang: Eh, it was just for confirming the duration is 6s. Not quite useful indeed. I'll remove this line.

examples/ami/ami.ipynb
jimbozhang: I have noticed this problem last week, and I double checked it. I think the codes of lhotse is correct,  but the time point labeling of AMI corpus is not very precise. Manually I modified the labels of an example for unit test: `test/fixtures/ami/cuts.yml`, and the result from the unit test looks correct.

danpovey:
Mm, some of these datasets don't label *all* speech, only some of it.
Also please make it clear whether this is the `sdm`, `ihm` or `mdm`
microphone setup (single distant; individual headset; multiple distant
[i.e. beamforming]).

jimbozhang:
> Also please make it clear whether this is the `sdm`, `ihm` or `mdm` microphone setup (single distant; individual headset; multiple distant [i.e. beamforming]).

This recipe is for `ihm` setup for now.

danpovey:
Anyway it doesn't have to be perfect.  We can refind it in future.
I was thinking it might be nice to prepare at least one dataset for
wake-word detection at some point.
@freewym knows about these things, he might have suggestions.

PR #65, opened 2020-08-14
Create CutSet from duration windows of another CutSet

pzelasko:
@danpovey This is what I meant in #64 by "create one large cut from a single feature (matrix) and cut it into windows". Check out the `CutSet.windows()` method. I noticed there were minor bugs while writing that so I added unit tests for all the ways we can create a `CutSet` at this moment.

In `test/cut/test_cut_set.py` you'll see two test classes (groups) - one is called `TestCutOnSupervisions` which is the current behavior if you supply a SupervisionSet during the CutSet creation. It creates cuts to match the supervision segment boundaries and discards the rest. 

I'm considering changing that to the behavior observed in `TestNoCutOnSupervisions` (marked xfail for now because it's not implemented), where we'll create cuts spanning the whole Recording/Features (matrix) duration, and add all observed SupervisionSegments in that time span. It'll make it easier to create windowed CutSets (with the new `.window()` method), and we can add a method called e.g. `CutSet.trim_to_supervisions()` to retain the old behavior.

Does it make sense now?

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -102,7 +102,7 @@ class Cut:
         return self.features.num_features if self.has_features else None
 
     @property
-    def features_type(self) -> Optional[int]:
+    def features_type(self) -> Optional[str]:
         return self.features.type if self.has_features else None
 
     @property
@@ -759,7 +759,7 @@ class CutSet:
             ))
         return CutSet.from_cuts(truncated_cuts)
 
-    def windows(self, duration: Seconds, keep_excessive_supervisions: bool = True) -> 'CutSet':
+    def cut_into_windows(self, duration: Seconds, keep_excessive_supervisions: bool = True) -> 'CutSet':
         """
         Return a new ``CutSet``, made by traversing each ``Cut`` in windows of ``duration`` seconds and
         creating new ``Cut`` out of them.
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -39,6 +39,16 @@ class Cut:
     """
     A Cut is a single "segment" that we'll train on. It contains the features corresponding to
     a piece of a recording, with zero or more SupervisionSegments.
+
+    The SupervisionSegments indicate which time spans of the Cut contain some kind of supervision information:
+    e.g. transcript, speaker, language, etc. The regions without a corresponding SupervisionSegment may
+    contain anything - usually we assume it's either silence or some kind of noise.
+
+    Note: The SupervisionSegment time boundaries are relative to the beginning of the cut.
+    E.g. if the underlying Recording starts at 0s (always true), the Cut starts at 100s,
+    and the SupervisionSegment starts at 3s, it means in the Recording, the supervision actually started at 103s.
+    In some cases, the supervision might have a negative start, or a duration exceeding the duration of the Cut.
+    It means that the supervision in the recording extends beyond the Cut.
     """
     id: str
 
@@ -85,11 +95,15 @@ class Cut:
 
     @property
     def num_samples(self) -> Optional[int]:
-        return self.recording.num_samples if self.has_recording else None
+        return round(self.duration * self.sampling_rate) if self.has_recording else None
 
     @property
     def num_features(self) -> Optional[int]:
         return self.features.num_features if self.has_features else None
+
+    @property
+    def features_type(self) -> Optional[int]:
+        return self.features.type if self.has_features else None
 
     @property
     def sampling_rate(self) -> int:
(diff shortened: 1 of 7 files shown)


lhotse/cut.py
```diff
@@ -39,6 +39,16 @@ class Cut:
...
     A Cut is a single "segment" that we'll train on. It contains the features corresponding to
     a piece of a recording, with zero or more SupervisionSegments.
+
+    The SupervisionSegments indicate which time spans of the Cut contain some kind of supervision information:
+    e.g. transcript, speaker, language, etc. The regions without a corresponding SupervisionSegment may
+    contain anything - usually we assume it's either silence or some kind of noise.
+
+    Note: The SupervisionSegment time boundaries are relative to the beginning of the cut.
+    E.g. if the underlying Recording starts at 0s (always true), the Cut starts at 100s,
+    and the SupervisionSegment starts at 3s, it means in the Recording, the supervision actually started at 103s.
```
danpovey, line 49: I'd rephrase to "it means that in the Recording the..."
the sentence doesn't read well currently.

lhotse/cut.py
```diff
@@ -39,6 +39,16 @@ class Cut:
...
+    In some cases, the supervision might have a negative start, or a duration exceeding the duration of the Cut.
+    It means that the supervision in the recording extends beyond the Cut.
```
danpovey, line 51: `Cut. It -> Cut; this`

lhotse/cut.py
```diff
@@ -741,6 +755,30 @@ def compute_offset():
             ))
         return CutSet.from_cuts(truncated_cuts)
 
+    def windows(self, duration: Seconds, keep_excessive_supervisions: bool = True) -> 'CutSet':
```
danpovey, line 758: Mm, OK, but this seems a rather non-ideal way to create cuts, because it doesn't make an attempt to avoid breaking supervisions up.  And surely we'd want to discard parts of the utterance that had no supervisions (or is that a separate operation)?

pzelasko: It depends on the task, right? For VAD, every part of the recording can be useful as long as the supervisions tell us where the speech is. For ASR, we can have other ways of creating cuts - e.g. cut right at the supervision segment boundaries, or with a small leeway, or make equally sized cuts that don't break supervisions up. Each of those would be a method on the CutSet (to be added sometime soon).

It all assumes that the typical workflow would be: create RecordingSet and SupervisionSet (and maybe a FeatureSet for the whole recordings), then create a CutSet with Cuts that are the same duration as the Recordings, and then call a method (like `windows()`) on that CutSet to transform it in a specific way. The benefit of that approach is that the CutSet becomes the place where you can find all the necessary pieces of (meta)data, data dependencies, and the "standard" tools for building your dataset however you want.

pzelasko:
Check out the changes after the last commit - now it's more flexible, allowing to create Cuts both as windows or tightly spanned over supervisions. Do you think we should add some other methods of "cutting" at this point?

lhotse/cut.py
danpovey: OK. Just bear it in mind when naming that function "windows".. if it's going to be just one of many, maybe choose a more specific name?

lhotse/supervision.py
```diff
@@ -64,6 +66,34 @@ def filter(self, predicate: Callable[[SupervisionSegment], bool]) -> 'Supervisio
...
+        :param recording_id: Desired recording ID.
+        :param channel: Return supervisions in the specified channel.
+        :param start_after: When specified, return segments that start after the given value.
+        :param end_before: When specified, return segments that end before the given value.
+        :param adjust_offset: When true, return segments as if the recordings had started at ``start_after``.
+            In other words, :math:`segment.start = segment.start - start_after`.
+        :return: An iterator over supervision segments satisfying all criteria.
+        """
+        return (
+            segment.with_offset(-start_after) if adjust_offset else segment
```
danpovey, line 89: can you explain why it's not necessary to modify the duration here?

lhotse/supervision.py
```diff
@@ -64,6 +66,34 @@ def filter(self, predicate: Callable[[SupervisionSegment], bool]) -> 'Supervisio
...
+            for segment in self
+            if segment.recording_id == recording_id
```
danpovey, line 91: wouldn't this be extremely slow for big manifests?  And if start_after and end_before meant to be filters can you clarify?  start_after seems to have a side effect of modifying the start time.  And I'm not sure the logic is correct if start_after and end_before are both nonzero.

lhotse/supervision.py
pzelasko: I extended the comments and involved a rationale - hopefully it's clearer now

lhotse/supervision.py
pzelasko: Yes, that could get quite slow. There was another place in the code with the same issue - I believe it's in `FeatureSet.find()`. I'd leave it be for now until it hurts us, and once it does, we can optimize it. I'm thinking about constructing and caching an index: something like a `Dict[<recording-id>, IntervalTree[Segment]]`, where the interval tree uses (start, end) to make the search in the time axis efficient.

pzelasko: Re other comments - I updated the description, but the idea is that `start_after` *may* modify the start time, but only if you explicitly request it with `adjust_offset=True`. This is to make it easier to create Cuts. I want the supervisions in the Cuts to have start/end times relative to the start of the Cut, and not to the start of the underlying Recording. I expect it to be more useful that way.

pzelasko: I added a bunch of unit tests to check the logic - I think it's fine?

danpovey: My feeling is just a dict on the recording-id, initially, would be enough.

lhotse/supervision.py
danpovey: Ah I see.  Could you please add:  Note: in the anticipated use-case, ``start_after`` and ``end_before`` would be the beginning and end of a cut; this option converts the times to be relative to the start of the cut.
IDK why this wasn't clear to me originally, but hopefully this will prevent the same confusion.  Otherwise, LGTM.

PR #69, opened 2020-09-02
Switchboard data preparation

pzelasko:
This introduces a data preparation recipe for Switchboard. 

It also fixes an issue I observed after merging AMI recipe - the code was not ready to handle `None` returned from `cut.truncate()`, I played around with the API and figured it's better to return an unmodified Cut when the user requires a longer duration than available, rather than returning `None`.

Notably, there is no special text preprocessing, probably I should add it to avoid partial words (`ex[ample]-`) in SWBD transcripts (I might just mimic Kaldi data prep behavior - @danpovey any comments?).

I will also try to add dialog act and sentiment labels here (not sure yet whether as a part of this PR or a separate one)

### DIFF
--- a/docs/corpus.rst
+++ b/docs/corpus.rst
@@ -145,7 +145,8 @@ a CLI tool that create the manifests given a corpus directory.
 
 Currently supported corpora:
 
-- AMI :func:`lhotse.recipes.ami.prepare_ami`
-- Mini LibriMix :func:`lhotse.recipes.librimix.prepare_librimix`
-- Mini LibriSpeech :func:`lhotse.recipes.mini_librispeech.prepare_mini_librispeech`
-- 1997 English Broadcast News :func:`lhotse.recipes.broadcast_news.prepare_broadcast_news`
+- AMI :func:`lhotse.recipes.prepare_ami`
+- English Broadcast News 1997 :func:`lhotse.recipes.prepare_broadcast_news`
+- Mini LibriMix :func:`lhotse.recipes.prepare_librimix`
+- Mini LibriSpeech :func:`lhotse.recipes.prepare_mini_librispeech`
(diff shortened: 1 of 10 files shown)


danpovey:
Thanks!
Partial words are generally not a problem because they'll be mapped to <unk> when we see they are not in the vocab.
Regarding text cleanup: I think anything about that should be clealy separable from the rest of Lhotse, e.g. passed in as a function object or something, and not at all wired in deeply.  We might end up using external codebases for that in some circumstances. And if anything is specific to a language or dataset, we should name it as such- don't name things in a way that makes them seem 'canonical' because there is no canonical way to do these things.
Also, remember, notation-wise: this process is sometimes called `text normalization` and there is something called `inverse text normalization` (ITN) for going from ASR output to  human-readable output.

RE dialog acts, sentiment labels and so on, just make sure you don't do anything that makes the basic version  of the recipe too complicated or confusing or have too-heavy dependencies.

pzelasko:
> Partial words are generally not a problem because they'll be mapped to when we see they are not in the vocab.
> Regarding text cleanup: I think anything about that should be clealy separable from the rest of Lhotse, e.g. passed in as a function object or something, and not at all wired in deeply. We might end up using external codebases for that in some circumstances. And if anything is specific to a language or dataset, we should name it as such- don't name things in a way that makes them seem 'canonical' because there is no canonical way to do these things.

OK, not touching that for now. We'll do it when the time's due.

> Also, remember, notation-wise: this process is sometimes called `text normalization` and there is something called `inverse text normalization` (ITN) for going from ASR output to human-readable output.

Haha, I should know given the ungodly number of hours spent on coding and polishing these things in multiple systems and languages ;)

danpovey:
ha, sorry, teaching grandma to suck eggs...

PR #70, opened 2020-09-03
Overall Python API lifting

pzelasko:
# Changes:

[This notebook has a nice overview of the changes.](https://github.com/pzelasko/lhotse/blob/d533df5521fe0b63226198095657d945dfee1a2d/examples/cut-python-api.ipynb)
It doesn't show the audio play widget, but it's there when you run it yourself.

### Moved many classes/methods to `lhotse` namespace

Easier discoverability, e.g. this is possible: `from lhotse import CutSet, Mfcc, VadDataset, load_manifest`

### Speedups in processing bottlenecks

Creating `CutSet` is now 10x or more faster.

### Feature extraction from cuts

Previously we could only extract the features from recordings; I figured it's quite useful to extract them from cuts. 
Actually, maybe it's reasonable going forward to enable that only from cuts to make the codebase simpler.

### Convenient visualisation/playback methods

I added a number of functions to `Cut`, such as `play_audio`, `plot_audio`, `plot_features` - they make it easier to check if the data is being processed correctly. They import dependencies (matplotlib/jupyter) locally, so if somebody doesn't need that, they won't be affected or forced to install them.

### Some renaming for consistency


lhotse/cut.py
```diff
@@ -34,8 +35,60 @@
 AnyCut = Union['Cut', 'MixedCut', 'PaddingCut']
 
 
+class CutCommonsMixin:
```
danpovey, line 38: I find the `Commons` part of the name confusing.  IDK if this is a common pattern you  got from somewhere?
Maybe CutUtilsMixin or something?  And perhaps comment what it's intended for, i.e. what classes are supposed to use it an what general kind of functionality it adds?

lhotse/dataset/vad.py
```diff
@@ -27,13 +27,12 @@ class VadDataset(Dataset):
     def __init__(
             self,
             cuts: CutSet,
-            duration: Optional[Seconds] = 5.0,
```
danpovey, line 30: A few lines above here: instead of `Returns a dict of:`, I'd rather say: `each item in this dataset is a dict of:`... I got confused because I thought by "Returns" you meant this class itself, which is like a dict too.

test/cut/test_feature_extraction.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.fixture
+def cut(recording):
+    return Cut(id='cut', start=0, duration=1.0, channel=0, recording=recording)
+
+
+def test_extract_features(cut):
+    extractor = Fbank()
+    arr = cut.compute_features(extractor=extractor)
+    assert arr.shape[0] == 100
+    assert arr.shape[1] == extractor.config.num_mel_bins
```
danpovey, line 35: Is there a more generic way of querying the feature-dim of an extractor object?

danpovey (commented):
Made few small comments directly, but LGTM!

lhotse/cut.py
pzelasko: Eh, it's a mess really. I've just tried to turn it into an abstract base class for all Cut classes, but `dataclasses` module authors did not bother to support ABCs. For now, I've changed the name, moved `mix` and `append` in here (as their implementations are identical in all types of cuts), and added a comment explaining this somewhat weird design choice.

As to `Commons`, only after your comment I've realized it has a well-defined meaning in English and is a bad substitute for "common/shared parts", which is not as terse. `Utils` it is, at least for now.

test/cut/test_feature_extraction.py
pzelasko: Good idea. I'll add that as a method that accepts `sampling_rate` argument so that it can be computed for `Spectrogram`.

pzelasko:
It may seem like I did a lot of changes - I just renamed further:
- `duration_seconds` -> `duration`
- `channel_ids` -> `channels`
- `channel_id` -> `channel`
I should have done that waaay earlier, before we acummulated the current amount of code/manifests/notebooks....

pzelasko:
Also, FYI, since we have a bit larger manifests now that Switchboard is here, I found that after my optimizations the most time-consuming part of every script is... loading and parsing YAML. We'll have to watch out for that when we add larger datasets like Librispeech or Fisher and check if it still can be done in an acceptable time.

pzelasko:
Uh, I did a little investigation and it's worse than I thought.

image

I quickly coded up JSON serialization and it's 40x faster 😂. [After a bit of googling, I found out about this:](https://stackoverflow.com/a/27744056/5285891)

image

I remember somebody (either @danpovey or @jtrmal) was against JSON for reason but I can't recall it. Can we have that discussion again? I'm open to other options (I don't mind JSON though), but keep in mind that it probably won't be possible to use a table-like text format, since most of the classes (`Recording`, `Cut`) are structured.

danpovey:
Mm, does seem like a big time difference.  What do the JSON manifests look like?

danpovey:
... and I wonder whether there might be value in adding options so it could read/write in json or yaml... dont want to make it too complicated though.

pzelasko:
Supervisions look like this:

```json
[
  {
    "id": "sw2001A-ms98-a-0002",
    "recording_id": "sw02001",
    "start": 0.977625,
    "duration": 10.584,
    "channel": 0,
    "text": "hi um yeah i'd like to talk about how you dress for work and and um what do you normally what type of outfit do you normally have to wear",
    "language": "English",
    "speaker": "sw02001A"
  },
  {
    "id": "sw2001A-ms98-a-0004",
    "recording_id": "sw02001",
    "start": 19.804875,
    "duration": 1.508,
    "channel": 0,
    "text": "um-hum",
    "language": "English",
    "speaker": "sw02001A"
  },
  {
    "id": "sw2001A-ms98-a-0006",
    "recording_id": "sw02001",
    "start": 27.362,
    "duration": 1.571,
    "channel": 0,
    "text": "and is",
    "language": "English",
    "speaker": "sw02001A"
  }
]
```

This is a single JSON cut:

```json
{
  "id": "f46d2b1e-3bb7-495d-bf6c-44bddf6b6651",
  "start": 0,
  "duration": 30,
  "channel": 0,
  "supervisions": [
    {
      "id": "sw2001A-ms98-a-0002",
      "recording_id": "sw02001",
      "start": 0.978,
      "duration": 10.584,
      "channel": 0,
      "text": "hi um yeah i'd like to talk about how you dress for work and and um what do you normally what type of outfit do you normally have to wear",
      "language": "English",
      "speaker": "sw02001A"
    },
    {
      "id": "sw2001A-ms98-a-0004",
      "recording_id": "sw02001",
      "start": 19.805,
      "duration": 1.508,
      "channel": 0,
      "text": "um-hum",
      "language": "English",
      "speaker": "sw02001A"
    },
    {
      "id": "sw2001A-ms98-a-0006",
      "recording_id": "sw02001",
      "start": 27.362,
      "duration": 1.571,
      "channel": 0,
      "text": "and is",
      "language": "English",
      "speaker": "sw02001A"
    }
  ],
  "recording": {
    "id": "sw02001",
    "sources": [
      {
        "type": "file",
        "channels": [
          0,
          1
        ],
        "source": "LDC97S62/swb1/sw02001.sph"
      }
    ],
    "sampling_rate": 8000,
    "num_samples": 2018387,
    "duration": 252.298375
  }
}
```

We could also check out [JSONL](https://jsonlines.org/) which is perhaps more amenable to stream processing.

From the implementation point of view, it's easy to support both YAML and JSON - the real question is, should we really do that, i.e., what's the benefit. From a user perspective, it's surely more confusing and I can easily imagine people getting frustrated at the library once they run into these issues.

danpovey:
I guess I agree that yaml+json would be too much.
JSONL doesn't look very readable.  I wonder if it could be supported as an
option when writing?
When reading, it wouldn't be that hard to check whether the 1st line is a
valid JSON object and, if so, read as JSONL; else JSON.

pzelasko:
I believe JSONL data is typically stored in files with a .jsonl extension,
so we could just check that. However, I'm not very keen on the idea of
supporting both, for the same reason as with YAML+JSON.

I suggest to replace YAML with JSON for now to enjoy better performance
with a minor (?) readability loss. If it turns out we need more
flexibility, we can revisit this choice and add another supported format
after all.

niedz., 6 wrz 2020, 23:25 użytkownik Daniel Povey <notifications@github.com>
napisał:

danpovey:
OK, fine.

PR #71, opened 2020-09-03
Switchboard sentiment annotations

pzelasko:
I've added an option to provide the path to LDC2020T14 during SWBD data preparation - then, it adds the sentiment labels to the supervisions. I also fixed some small issues with the CLI.

With sentiment labels the supervisions look like this (there are multiple annotators and the dataset creators do not disambiguate the labels, leaving it up to the user - so did I):

```yaml
- channel_id: 0
  custom:
    sentiment0: Neutral-{Mixed emotion}
    sentiment1: Neutral-{Mixed  emotion}
    sentiment2: Positive-{Industrialize}
  duration: 9.474
  id: sw2005A-ms98-a-0051
  language: English
  recording_id: sw02005
  speaker: sw02005A
  start: 381.31075
  text: um i[n]- matter of fact in the United States we used to have extended families
    it wasn't but i guess as we become more industrialized and more you know less
    in a rural situation
- channel_id: 0
  custom:
    sentiment0: Positive-{Easier}
    sentiment1: Positive-{Easier}
    sentiment2: Positive-{Easier}
  duration: 10.446
  id: sw2005A-ms98-a-0052
  language: English
  recording_id: sw02005
  speaker: sw02005A
  start: 390.784625
  text: we we don't we we we choose not to deal with the extended family because we
    feel it's kind of cumbersome when in reality it makes things much much easier
```

### DIFF
--- a/lhotse/bin/modes/recipes/__init__.py
+++ b/lhotse/bin/modes/recipes/__init__.py
@@ -1,3 +1,4 @@
 from .broadcast_news import *
 from .librimix import *
 from .mini_librispeech import *
+from .switchboard import *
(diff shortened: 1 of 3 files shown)


danpovey (commented):
LGTM!!

PR #75, opened 2020-09-14
Feature extraction from mixed cuts

pzelasko:


pzelasko:
@danpovey ok to review

danpovey (commented):
LGTM!
Was this in response to a conversation with Paola?
She was concerned about what happens where the audio is extremely long and we don't want to load it all at once.

pzelasko:
Actually no, we haven't talked (yet). It's adding a missing feature - I
implemented feature extraction from Cuts before, but it didn't support
MixedCuts. I also saw an opportunity to get rid of librosa as a core
dependency - it installs ~20 other deps. The audio was already being read
partially before though.

śr., 16 wrz 2020, 00:21 użytkownik Daniel Povey <notifications@github.com>
napisał:

danpovey:
Great!

PR #80, opened 2020-09-18
A preliminary implementation of a DiarizationDataset

pzelasko:
I will consult with Paola (couldn't find her GitHub handle for some reason...) what would be the most useful formulation of a `DiarizationDataset` for her - this is a starting point implementation that made sense to me.

It would return elements like this:
image

where each row is effectively a VAD supervision for each speaker (allowing to train the models that deal with speech overlap).


danpovey:
Great!
I am hoping Paola can comment on this.

pzelasko:
I am going to merge this so it doesn't rot too much, and we can make changes in the future as needed.

pzelasko:
(we had some discussions with Paola and it seems it's going in the right direction)

Issue #67, opened 2020-09-01
TTS recipe


jimbozhang:
I'm not sure if it useful to let lhotse support TTS training. If needed, I can make a TTS recipe using the [LJSpeech](https://keithito.com/LJ-Speech-Dataset/) corpus.
@danpovey @pzelasko

danpovey:
That sounds like a great idea!  Yes, I intend for it to support all kinds
of speech-related tasks.

pzelasko:
Yes, it will be most useful. Thanks!

jimbozhang:
Just started working on this. A new recipe `ljspeech` will be added for end-to-end (e.g. Tacotron2) training. I'll make a PR next week.

pzelasko:
Sounds great! A heads up - I made a couple of general changes in the meanwhile, e.g. the preferred manifest format is now JSON because it’s much faster to read/write.

jimbozhang:
> Sounds great! A heads up - I made a couple of general changes in the meanwhile, e.g. the preferred manifest format is now JSON because it’s much faster to read/write.

I had a little worry: if the dataset is extremely huge, the JSON would be very large. To avoid out-of-memory, the manifest has to be splitted. I think splitting YAML is easier than JSON, because if you want to split a JSON, you have to load it into memory first.

danpovey:
We had also discussed JSONL format (JSON, one element per line).
It will take quite a big dataset to hit that memory limit though.

pzelasko:
Hmm, you're raising a good point. We can add JSONL quite easily - we're already supporting two formats anyway... 

I wonder if it makes sense to use a binary format for such super-large datasets. E.g. in https://github.com/huggingface/nlp/ they use Apache Arrow to traverse mmap-ed files, allowing them to iterate very large datasets with almost zero memory footprint (likely at the cost of storage size).

danpovey:
Lets solve that only once it becomes a problem.

Issue #79, opened 2020-09-17
torchaudio.sox_signalinfo_t is going to be deprecated.


mthrok:
Hi

torchaudio is perfoming overhaul on I/O and libsox binding. The detail is found [here](https://github.com/pytorch/audio/issues/903).
As part of this plan, in 0.9.0 release we plan to remove libsox class from API.

I looked at `lhoste` and it refers to `torchaudio. sox_signalinfo_t`. for example [here](https://github.com/pzelasko/lhotse/blob/3a34a9d34615a82913d24f9728cf33dae4a93032/lhotse/recipes/mini_librispeech.py#L42)
It is not immediately clear to me how this can be migrated, but torchaudio team is eager to help this project.
Please let us know what you think of torchaudio.

cc @dongreenberg @astaff @cpuhrsch @vincentqb

pzelasko:
Thanks for the heads up! 

AFAIK the code you pointed to is currently the only usage of torchaudio to read audio/obtain audio file metadata in our codebase. We mostly rely on libsndfile (and python `SoundFile` wrapper) for loading audio otherwise. We'll update this recipe to do so as well.

@jimbozhang since you contributed the recipe, could you look into it?

Otherwise, we're using torchaudio for Kaldi-style feature extraction and so far it's working well for us.

jimbozhang:
Thanks @mthrok's remainder. I think we need to update `torchaudio` to `>=0.9.0` and make `lhotse` compatible with it. I'll do this soon. @pzelasko

pzelasko:
I don't think we actually need to bump the version - let our users have their choice if they want to use the older one. If we replace this usage of `sox_signalinfo_t` with `SoundFile` to read the audio metadata, it should work with any torchaudio version.

mthrok:
> I don't think we actually need to bump the version - let our users have their choice if they want to use the older one. If we replace this usage of `sox_signalinfo_t` with `SoundFile` to read the audio metadata, it should work with any torchaudio version.

I looked at the code and sadly (for us), it seems that the right way is to replace `torchaudio` with `soundfile`. If you are going to create a recipe for a dataset with formats which `soundfile` does not support (such as `mp3` and `opus`), `torchaudio` supports it so let us know and we can help you.

pzelasko:
I replaced the `torchaudio.info` use with `soundfile.info`. Thanks again for letting us know @mthrok!

PR #81, opened 2020-09-20
a TTS recipe using the LJ-Speech dataset

jimbozhang:
For issue #67.


lhotse/dataset/speech_synthesis.py
```diff
@@ -0,0 +1,73 @@
...
+from lhotse.utils import Pathlike
+
+EPS = 1e-8
+
+
+def text_to_tokens(text: str) -> List[str]:
+    text = re.sub(r'[^\w !?]', '', text)
+    text = re.sub(r'^\s+', '', text)
+    text = re.sub(r'\s+$', '', text)
+    text = re.sub(' +', ' ', text)
```
pzelasko, line 18:
```suggestion
    text = re.sub('\s+', ' ', text)
```

Just to be on the safe side if there's any tabs or other whitespace characters. 

As a note to myself and for discussion - it seems like this is another place in the code (next to ASR dataset) where we could use some kind of standardized interface for customizable text normalization (as probably almost every corpus will require to do it a bit differently).

lhotse/dataset/speech_synthesis.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(
+            self,
+            cuts: CutSet,
+            root_dir: Optional[Pathlike] = None
+    ):
+        super().__init__()
+        self.cuts = cuts
+        self.root_dir = Path(root_dir) if root_dir else None
+        self.cut_ids = list(self.cuts.cuts.keys())
```
pzelasko, line 44:
```suggestion
        self.cut_ids = list(self.cuts.ids)
```

lhotse/dataset/speech_synthesis.py
```diff
@@ -0,0 +1,73 @@
...
+            self.token_set.update(set(tokens))
+            self.id_to_tokens[cut.id] = tokens
+
+    def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
+        cut_id = self.cut_ids[idx]
+        cut = self.cuts[cut_id]
+
+        features = torch.from_numpy(cut.load_features(root_dir=self.root_dir))
+        audio = torch.from_numpy(cut.load_audio(root_dir=self.root_dir))
+        assert len(cut.supervisions) == 1, "SpeechSynthesisDataset does not support multiple supervisions yet."
```
pzelasko, line 60: I'd rather move this assertion into constructor (in the loop over cuts) to warn the user early that he's breaking the pre-conditions.

lhotse/recipes/ljspeech.py
```diff
@@ -0,0 +1,129 @@
...
+    if not completed_detector.is_file():
+        shutil.rmtree(corpus_dir, ignore_errors=True)
+        with tarfile.open(tar_path) as tar:
+            tar.extractall(path=target_dir)
+            completed_detector.touch()
+
+
+class LJSpeechMetaData(NamedTuple):
+    audio_path: Pathlike
+    audio_info: torchaudio.sox_signalinfo_t
```
pzelasko, line 48: The torchaudio guys warned us about this, but I think it's okay to merge now and fix all usages in another PR

lhotse/dataset/speech_synthesis.py
```diff
@@ -0,0 +1,73 @@
...
+
+def text_to_tokens(text: str) -> List[str]:
+    text = re.sub(r'[^\w !?]', '', text)
+    text = re.sub(r'^\s+', '', text)
+    text = re.sub(r'\s+$', '', text)
+    text = re.sub(' +', ' ', text)
+    return list(text.upper())
+
+
+class SpeechSynthesisDataset(Dataset):
```
pzelasko, line 22: Can you add a unit test for this dataset?

lhotse/dataset/speech_synthesis.py
```diff
@@ -0,0 +1,73 @@
...
+            'audio': audio,
+            'features': features,
+            'tokens': self.id_to_tokens[cut.id]
+        }
+
+    def __len__(self) -> int:
+        return len(self.cut_ids)
+
+    @property
+    def targets(self) -> List[str]:
```
pzelasko, line 72: Isn't "vocabulary" a more appropriate name? TTS "targets" suggest spectrograms/waveforms

examples/ljspeech/ljspeech.ipynb
```diff
@@ -0,0 +1,281 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "['I', 'N', ' ', '1', '8', '1', '3']\n"
```
pzelasko, line 173: I assume this text is typically normalized before training (digits to words)?

jimbozhang: Text normalization was necessary for conventional TTS pipelines, but some end-to-end TTS systems do not need TN anymore. Anyway, it would be good to allow `lhotse` uses an external TN module.
I plan to allow the function `text_to_tokens()` be overwritten and as a optional parameter of `SpeechSynthesisDataset.__init__()`.

lhotse/dataset/speech_synthesis.py
jimbozhang: I forgot to write unit tests. I'll add them soon.

danpovey:
What did the torchaudio guys warn us about?

lhotse/recipes/ljspeech.py
```diff
@@ -0,0 +1,129 @@
...
+def prepare_ljspeech(
+        corpus_dir: Pathlike,
+        output_dir: Optional[Pathlike] = None
+) -> Dict[str, Union[RecordingSet, SupervisionSet]]:
+    """
+    Returns the manifests which consist of the Recordings and Supervisions
+
+    :param corpus_dir: Pathlike, the path of the data dir.
+    :param output_dir: Pathlike, the path where to write the manifests.
+    :return: a Dict whose key is the dataset part, and the value is Dicts with the keys 'audio' and 'supervisions'.
```
danpovey, line 61: Would be nice to say specifically what these keys are for the dataset part-- 'train', 'test'?

lhotse/dataset/speech_synthesis.py
danpovey: Mm.  My inclination would be to leave this in the more 'local' part of the code, in ljspeech.py.
I am cautious about putting anything like that centrally until we are more confident we know where we are going with it.  The danger is that we'll be tempted to add options, etc., creating interfaces that aren't needed.

lhotse/recipes/ljspeech.py
jimbozhang: Sorry for the incorrect comments. I'll fix them.

jimbozhang:
> What did the torchaudio guys warn us about?

torchaudio.sox_signalinfo_t is going to be deprecated. #79

lhotse/dataset/speech_synthesis.py
pzelasko: I was thinking about how to interact with text normalization in Lhotse for some time now, and I suggest we add a method called "transform_text" to `CutSet` and `SupervisionSet` that accepts a single function argument (`transform(text: str) -> str`), goes through all supervisions and applies that on their text. This is general enough that you can use it to split to characters, sub-word tokens, transcribe with G2P, whatever you can think of.

BTW Ideally we return new objects instead of modifying the existing ones, as that would be a consistent behaviour with other transform-like methods I implemented so far. I'm not 100% sure if we want immutability for Cuts/Supervisions etc. in the long run - if things get sluggish, we should rethink.

pzelasko:
@jimbozhang I forgot to mention documentation - could you add a mention about `SpeechSynthesisDataset` in `docs/datasets.rst` and `LJSpeech` in `docs/corpus.rst`?

lhotse/dataset/speech_synthesis.py
jimbozhang: I think we can let users modify supervisions by their own functions, and the ljspeech recipe provide a TN function as well, as  the commit https://github.com/pzelasko/lhotse/pull/81/commits/77d07e447090ea2a9b51285b1ec9956809952046 do.

lhotse/supervision.py
```diff
@@ -75,6 +75,15 @@ def filter(self, predicate: Callable[[SupervisionSegment], bool]) -> 'Supervisio
         """
         return SupervisionSet.from_segments(seg for seg in self if predicate(seg))
 
+    def modify(self, process: Callable[[SupervisionSegment], None]) -> None:
```
pzelasko, line 78: Suggestions:
- similar changes as in the `CutSet`
- rename `modify` to `map`
- rename `process` to `transform_fn` with signature `Callable[[SupervisionSegment], SupervisionSegment]`
- return a new `SupervisionSet` (`SupervisionSet.from_segments(...)`)

pzelasko (commented):
I'd like us to rethink some parts of this design. What I don't like about it, is that the supervision is mutated in-place - so as soon as the function is applied, you no longer have access to the unmodified version. The approach I took instead in some recent PRs is to assume that the SupervisionSegment/Cut is immutable - and return copies of the objects with the relevant fields updated. 

You can see examples of that in `SupervisionSegment` methods `.trim()` and `.with_offset()`, as well as a lot of `CutSet` methods: `.compute_and_store_features()`, `.cut_into_windows()`, `.truncate()`, etc. They all return copies of the original objects so that if the user wants to have multiple versions of some object, he does not become surprised by Python's mechanism of sharing references to the same object. To that end, I even benchmarked multiple ways of creating copies and implemented the fastest one as a `fastcopy` method that can override select fields during copying, see `.trim()` and `.with_offset()` for example.

I think the required modifications are quite simple, but I wanted to share the rationale first. I'd also like to get some feedback first - my main concern is performance (immutability may be costly for very large CutSets) but my current thinking is that it's worth it. It is also rather consistent with APIs of some commonly used data science libraries like `pandas` (notably they have an `inplace=False` argument for many operations, so both options are supported).

pzelasko:
Otherwise everything looks great! Just making sure we're consistent w.r.t. to mutability/immutability...

lhotse/cut.py
```diff
@@ -139,6 +139,17 @@ def supervisions_audio_mask(self) -> np.ndarray:
             mask[st:et] = 1
         return mask
 
+    def map_supervisions(self, transform_fn: Callable[[SupervisionSegment], SupervisionSegment]) -> 'CutUtilsMixin':
```
pzelasko, line 142:
```suggestion
    def map_supervisions(self, transform_fn: Callable[[SupervisionSegment], SupervisionSegment]) -> AnyCut:
```

lhotse/cut.py
```diff
@@ -139,6 +139,17 @@ def supervisions_audio_mask(self) -> np.ndarray:
...
+        """
+        Modify the SupervisionSegments by `transform_fn` of this Cut.
+
+        :param transform_fn: a function that modifies a supervision as an argument.
+        :return: a modified Cut.
+        """
+        new_cut = fastcopy(self)
```
pzelasko, line 149: This will be even faster:

    new_cut = fastcopy(self, supervisions=[s.map(transform_fn) for s in new_cut.supervisions])

lhotse/cut.py
```diff
@@ -1152,15 +1163,14 @@ def with_features_path_prefix(self, path: Pathlike) -> 'CutSet':
     def with_recording_path_prefix(self, path: Pathlike) -> 'CutSet':
         return CutSet.from_cuts(c.with_recording_path_prefix(path) for c in self)
 
-    def apply_supervision_modifications(self, process_supervision: Callable[[SupervisionSegment], None]) -> None:
+    def map_supervisions(self, transform_fn: Callable[[SupervisionSegment], SupervisionSegment]) -> 'CutSet':
```
pzelasko, line 1166: We'll need that method for `PaddingCut` and `MixedCut` too

pzelasko: (for `PaddingCut` it can just return self, for `MixedCut` it should apply this method on all tracks and return a new `MixedCut`)

lhotse/supervision.py
```diff
@@ -38,6 +38,17 @@ def trim(self, end: Seconds) -> 'SupervisionSegment':
         end_exceeds_by = max(0, self.end - end)
         return fastcopy(self, start=max(0, self.start), duration=self.duration - end_exceeds_by - start_exceeds_by)
 
+    def map(self, transform_fn: Callable[['SupervisionSegment'], None]) -> 'SupervisionSegment':
```
pzelasko, line 41: Now we have an incosistent API for Supervisions and Cuts -> one accepts a function that returns nothing, and the other a function that returns a new object. Can you change this `transform_fn` to return a SupervisionSegment?

lhotse/supervision.py
```diff
@@ -75,14 +86,14 @@ def filter(self, predicate: Callable[[SupervisionSegment], bool]) -> 'Supervisio
         """
         return SupervisionSet.from_segments(seg for seg in self if predicate(seg))
 
-    def modify(self, process: Callable[[SupervisionSegment], None]) -> None:
+    def map(self, transform_fn: Callable[[SupervisionSegment], None]) -> 'SupervisionSet':
```
pzelasko, line 89: Same comment as above about the signature consistency.

pzelasko (commented):
A couple of comments - almost there!

lhotse/cut.py
jimbozhang: `Cut`, `PaddingCut` and `MixedCut` are all derived from `CutUtilsMixin`, aren't them? So I implemented `map_supervisions` as a member of `CutUtilsMixin`.

lhotse/supervision.py
jimbozhang: What I thought is that `transform_fn` would be provided by users, so return `None` might be easier for users to write this function. :grin:
But I'm not sure about that. If needed, I can change that to return a `SupervisionSegment`.

lhotse/cut.py
jimbozhang: Emm.. yes, this method should perform different behaviors in these three classes. I'll add for them separately.

lhotse/supervision.py
pzelasko: You're raising a good point. I think if we want to be super-user-friendly we could expose more specific functions like `SupervisionSet.transform_text(transform_fn: Callable[[str], str)`. I think `map` is more general-purpose and might confuse inexperienced users anyway, so no harm is done. But I also think it's good enough for now if it behaves consistently with the other map-like functions in Lhotse.

Also consider this: you cannot mutate a string or a tuple in Python, you would always have to return a new instance (like in the `transform_text` example). I'd like to pretend Lhotse's classes are immutable too, which makes it easier to create multiple "versions" of cuts/supervisions. Maybe it makes sense to make them truly immutable - I'll have to think more about this, I'm not super sure ATM.

lhotse/cut.py
pzelasko: Ha, I missed that the method is in `CutUtilsMixin`. Yeah, they'll need different behaviours - the cuts have definitely grown over time... 😀

pzelasko (approved):
Looks great, thank you!

PR #82, opened 2020-09-21
Tune default feature configs

pzelasko:
I'm somewhat at a loss as to what should be the default settings. I assume we consider `fbank` as the go-to feature extractor. The options I see:
1) if we want `fbank` to resemble mfcc-hires from Kaldi (AFAIK the only difference is DCT for storage and IDCT as the first NN layer?), we can set `num_mel_bins` to 40 (which is what I committed now)
2) ESPnet uses 80-dim fbank for their models - but IDK if that makes a significant difference in results...

Any of these is probably better than the Kaldi default of 23 Mel bins which I don't think is used anywhere.

Also, I left Kaldi's default 23 `num_mel_bins` and 13 `num_ceps` for MFCCs.

@danpovey WDYT?


danpovey:
The defaults you are using seem OK to me.  I don't think the difference between 40 and 80 tends to be consistent.
I think people can override the default if they care.

PR #86, opened 2020-09-26
Refactor the "root_dir" argument in audio/feature loading APIs

pzelasko:


danpovey (commented):
yeah, I can think of situations where this would be necessary, like when a common file system is dynamically mounted.

PR #90, opened 2020-09-29
update CutSet.from_manifests parameters for ipynbs

jimbozhang:


pzelasko:
Thanks @jimbozhang, my bad - now I know that PyCharm doesn't scan the notebooks when refactoring 😅

PR #92, opened 2020-10-02
Refactoring of features storage I/O and an initial (probably inefficient) HDF storage implementation

pzelasko:
OK so I looked into HDF5 a little bit. This is a rough draft, failing tests, no comments/docs etc. just checking if this design is adequate and can work. It's mostly refactoring of the writing-to-individual-files storage method so that it's easy to plug in some other backend - this is done by introducing `FeaturesWriter` and `FeaturesReader`. They have a `name` which is written into feature manifests so during load the correct `FeaturesReader` can be instantiated (because it's decorated with `@registered_reader` decorator). Since that would lead to opening and closing handles on HDF file all the time, it is being held in an LRU cache (global program state) and can be cleared with a `close_cached_hdf_files()` call. It shouldn't lead to any problems with forking because I expect that the first call to `load_features()` will only happen after a fork (for multi-process training/dataloading), so each subprocess will open its own file handles.

The HDF5 raw-string handling documentation is not great ([they suggest putting binary data into metadata headers?!](https://docs.h5py.org/en/latest/strings.html#how-to-store-raw-binary-data)) and I might be doing something dumb. Maybe somebody more familiar with HDF can comment on that.

Below is a screenshot of usage in practice. A naive benchmark doesn't show much difference in terms of runtime on my MacBook, probably need to check on more data and on the CLSP grid which has slow I/O and will reveal any possible performance benefits. IDK when I'll have time to pick this up so feel free to comment on this in the meanwhile.

image

### DIFF
--- a/lhotse/features/base.py
+++ b/lhotse/features/base.py
@@ -7,12 +7,12 @@ from math import isclose
 from pathlib import Path
 from typing import Optional, Any, List, Iterable, Type, Union, Dict
 
-import lilcom
 import numpy as np
 import torch
 
 from lhotse.audio import Recording
 from lhotse.augmentation import WavAugmenter
+from lhotse.features.io import FeaturesWriter, get_reader
 from lhotse.supervision import SupervisionSegment
 from lhotse.utils import Seconds, Pathlike, load_yaml, save_to_yaml, uuid4, JsonMixin, YamlMixin, fastcopy
 
@@ -101,12 +101,10 @@ class FeatureExtractor(metaclass=ABCMeta):
     def extract_from_samples_and_store(
             self,
             samples: np.ndarray,
+            storage: FeaturesWriter,
             sampling_rate: int,
-            output_dir: Pathlike,
             offset: Seconds = 0,
             augmenter: Optional[WavAugmenter] = None,
-            compress: bool = True,
-            lilcom_tick_power: int = -5,
     ):
         """
         Extract the features from an array of audio samples in a full pipeline:
@@ -136,7 +134,7 @@ class FeatureExtractor(metaclass=ABCMeta):
         if augmenter is not None:
             samples = augmenter.apply(samples)
         feats = self.extract(samples=samples, sampling_rate=sampling_rate)
-        output_features_path = store_feature_array(feats, output_dir, compress, lilcom_tick_power)
+        storage_key = store_feature_array(feats, storage=storage)
         return Features(
             start=offset,
             # We simplify the relationship between num_frames and duration - we guarantee that
@@ -146,20 +144,19 @@ class FeatureExtractor(metaclass=ABCMeta):
             num_frames=feats.shape[0],
             num_features=feats.shape[1],
             sampling_rate=sampling_rate,
-            storage_type='lilcom' if compress else 'numpy',
-            storage_path=str(output_features_path)
+            storage_type=storage.name,
+            storage_path=storage.storage_path,
+            storage_key=storage_key
         )
 
     def extract_from_recording_and_store(
             self,
             recording: Recording,
-            output_dir: Pathlike,
+            storage: FeaturesWriter,
             offset: Seconds = 0,
             duration: Optional[Seconds] = None,
             channels: Union[int, List[int]] = None,
             augmenter: Optional[WavAugmenter] = None,
-            compress: bool = True,
-            lilcom_tick_power: int = -5,
     ):
         """
         Extract the features from a ``Recording`` in a full pipeline:
@@ -190,7 +187,7 @@ class FeatureExtractor(metaclass=ABCMeta):
         if augmenter is not None:
             samples = augmenter.apply(samples)
         feats = self.extract(samples=samples, sampling_rate=recording.sampling_rate)
-        output_features_path = store_feature_array(feats, output_dir, compress, lilcom_tick_power)
+        storage_key = store_feature_array(feats, storage=storage)
         return Features(
             recording_id=recording.id,
             channels=channels if channels is not None else recording.channel_ids,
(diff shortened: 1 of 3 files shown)


danpovey:
I wonder if int8_t might be a better way to store the raw character data?

lhotse/features/base.py
```diff
@@ -565,14 +564,6 @@ def store_feature_array(
         might be appropriate for non-log space features.
     :return: a path to the file containing the stored array.
```
danpovey, line 565: Now it seems to store a storage_key?  Is this concept defined anywhere?

lhotse/features/io.py
```diff
@@ -0,0 +1,177 @@
+from abc import ABCMeta, abstractmethod
+from functools import lru_cache
+from pathlib import Path
+from typing import Type
+
+import lilcom
+import numpy as np
```
danpovey, line 7: hard for me to comment on this file without documentation...

danpovey (commented):
looks OK but need some clarification

danpovey:
Have you ever seen someone use hdf5 for stuff like speech data that would naturally be indexed by, say, string and then be an uneven-sized array?  I don't know what the normal pattern is.  It's hard to be confident we're doing it right without knowing how someone would use hdf5 for our kind of data.

lhotse/features/base.py
pzelasko: Yeah, I will describe it later in the code, the short story is: a single HDF file has a `storage_path` on the filesystem (e.g. `feats.hdf`) and stores multiple utterances indexed by the `storage_key` (e.g. `utt1`, `utt2`, `utt3`). For arrays stored in separate files on the filesystem, `storage_path` would be the name of the directory, and `storage_key` would be the filename inside that directory.

pzelasko:
I multiplied `mini_librispeech` train set size by 20 and extracted the features with either files or hdf5 storage with lilcom compression. Then, I benchmarked the features loading sequentially with a single thread from a NFS node on the CLSP grid. 

Mean loading times for 7 runs:
Files + lilcom: 57.3 seconds
HDF5 + lilcom: 38.1 seconds

So HDF5 reader managed to load it in 66% of the time. I'll also later check the HDF5 without lilcom compression for comparison.

pzelasko:
> I wonder if int8_t might be a better way to store the raw character data?

I checked, there was no difference either in size or speed compared to void.

I have more news about the different storage strategies:
Files + lilcom: 57.3 seconds
HDF5 + lilcom: 38.1 seconds
pure HDF5: 15.8 seconds

In terms of storage size:
HDF5 + lilcom: 1.7GB
pure HDF5: 5.8GB

So `lilcom` enables more than 3x size reduction at the cost of 2.5x longer reads. The benefits are likely become more pronounced with parallel reads but I don't have the bandwidth to test it right now. Also, @danpovey maybe the decompression can be further optimized to reduce the read times? 

Another issue with the current implementation is doing parallel writes to HDF5 file (if you wanted to use multiprocessing/threading to parallelize). I know there is some support for that in HDF5 but you need to compile with MPI and add some extra arguments. We can always split the RecordingSet/CutSet into a few parts and launch individual processes to store the data in separate HDF5 files. I'm open to suggestions (and contributions).

I'll fix the tests sometime soon and make this PR ready to merge.

danpovey:
Great!
There is probably some scope for optimizing the decompression speed; but
the motivation for it was more about not overloading the file server,
especially for cases where there is parallel training.  We can look at that
once it's a limiting factor,
Dan

lhotse/features/io.py
pzelasko: The docs are here now 🔥

pzelasko:
@danpovey ready for final review if you'd like

docs/features.rst
```diff
@@ -17,11 +17,14 @@ Storing features
...
 Features in Lhotse are stored as numpy matrices with shape ``(num_frames, num_features)``.
-By default, we use `lilcom`_ for lossy compression and reduce the size on the disk approximately by half.
+By default, we use `lilcom`_ for lossy compression and reduce the size on the disk by about 3x.
 The lilcom compression method uses a fixed precision that doesn't depend on the magnitude of the thing being compressed, so it's better suited to log-energy features than energy features.
-For now, we extract the features for the whole recordings, and store them in separate files.
-We retrieve them by loading the whole feature matrix and selecting the relevant region (e.g. specified by a cut).
-Eventually we will look into optimizing the storage to further reduce the I/O.
+We currently support two kinds of storage:
+
+- HDF5 files with multiple feature matrices (recommended)
```
danpovey, line 24: Will this actually give a speed boost?  I would imagine it depends on the order of access?

docs/features.rst
```diff
@@ -17,11 +17,14 @@ Storing features
...
+- directory with feature matrix per file (not recommended, costly I/O)
+
+We retrieve the arrays by loading the whole feature matrix from disk and selecting the relevant region (e.g. specified by a cut). Therefore it makes sense to cut the recordings first, and then extract the features for them to avoid loading unnecessary data from disk (especially for very long recordings).
```
danpovey, line 27: We might at some point need to implement a storage method that avoids this decision... e.g. use lilcom to separately store features for 1 second at a time and figure out a transparent way to load this.

docs/features.rst
pzelasko: I've just ran some tests with random access, it does seem to take 50-70% longer time to pass over mini_librispeech features in HDF5 than sequential access. But then sometimes it's the same speed if you run it many times (caching kicks in) and (probably) nobody else is using that disk. So I'm not sure how much slowdown is due to random access and how much due to other users. But for sure there is an opportunity to optimize it further with some kind of chunked storage etc.

docs/features.rst
pzelasko: Sounds good. It's probably good to drive the decisions where to optimize based on the training speed/bottlenecks when we have some recipes.

docs/features.rst
pzelasko: I removed the recommendations - we obviously need to spend more time playing with this

PR #95, opened 2020-10-09
Updating usage examples, contributor's guide, fixing some bugs

pzelasko:
I added docs about adding new corpora and updated feature extraction examples in the notebooks to use the new storage mechanism and do in directly on the cuts.

I also made sure that the LibriSpeech 100h example works smoothly with a DataLoader and resolved a number of bugs I found on the way (mismatching dtypes, off-by-one frames in padding, and a bunch of others).


lhotse/audio.py
```diff
@@ -277,6 +280,7 @@ def __init__(self, base_audio: np.ndarray, sampling_rate: int):
         self.tracks = [base_audio]
         self.sampling_rate = sampling_rate
         self.reference_energy = audio_energy(base_audio)
+        self.dtype = self.tracks[0].dtype
```
danpovey, line 283: what would this be in practice?  would be nice to have a comment

lhotse/audio.py
```diff
@@ -277,6 +280,7 @@ def __init__(self, base_audio: np.ndarray, sampling_rate: int):
         self.tracks = [base_audio]
```
danpovey, line 280: RE "point of reference in terms of offset", is there a way to make that more precise?  It's not clear to me why users would want to specify the energy via a file, and whether it's a good idea for all the output to have the same average energy.  (Systems sometimes are not robust w.r.t. energy variations).

danpovey: oh, perhaps the AudioMIxer is to be instantiated for each piece of audio to be mixed.  Might be good to make that clear?

pzelasko: You are right in the latter comment. I've updated the description of both Audio and Feature mixers to clarify.

PR #96, opened 2020-10-14
Make split() a method of the manifest classes (and add CutSet support)

pzelasko:
Mostly a discoverability improvement + fixing lack of support for CutSet. I found the option in PyCharm to sort imports alphabetically, too...

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -4,11 +4,12 @@ from io import BytesIO
 from math import sqrt
 from pathlib import Path
 from subprocess import PIPE, run
-from typing import Callable, Dict, Iterable, List, Optional, Union, Tuple
+from typing import Callable, Dict, Iterable, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 
-from lhotse.utils import Decibels, Pathlike, Seconds, SetContainingAnything, JsonMixin, YamlMixin, fastcopy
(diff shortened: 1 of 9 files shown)


danpovey:
Looks good!  Sorry I started looking at it earlier but got interrupted and didnt get back to it.

pzelasko:
Haha, it's my bad - these itchy fingers love those shiny merge buttons...

lhotse/audio.py
```diff
@@ -218,6 +219,19 @@ def filter(self, predicate: Callable[[Recording], bool]) -> 'RecordingSet':
         """
         return RecordingSet.from_recordings(rec for rec in self if predicate(rec))
 
+    def split(self, num_splits: int, randomize: bool = False) -> List['RecordingSet']:
+        """
+        Split the ``RecordingSet`` into ``num_splits`` pieces of equal size.
```
csukuangfj, line 224: > into ``num_splits`` pieces of equal size

The comment is misleading and does not match the implementation.

It does not describe how it handles the case when `len(self) % num_splits != 0`. 

There is an option `drop_last` in PyTorch's dataloader and `drop_remainder` in Tensorflow's dataset.

The current implementation simply uses the semantic `drop_last=False` without
documenting it.

pzelasko: I’ll fix it (soon)

PR #99, opened 2020-10-28
[WIP] Sharding recordings across the filesystem

pzelasko:
I plan to extend this functionality to feature matrices, `FeatureSet` and `CutSet` as well. Features will be a bit more work as every backend will have to support it (maybe I'll implement one and open an issue for volunteers for the other one). Open to feedback on the design. 

I'm also thinking about another option to return N RecordingSets/CutSets, each representing a different shard, which might be more useful in some distributed training scenarios (each worker node loading data from its exclusive shard). I don't have any setup to test such scenario at the moment, so we might add it in the future instead.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1,3 +1,5 @@
+import math
+import random
 import warnings
 from dataclasses import asdict, dataclass
 from io import BytesIO
@@ -9,7 +11,7 @@ from typing import Callable, Dict, Iterable, List, Optional, Sequence, Tuple, Un
 import numpy as np
 
 from lhotse.utils import (Decibels, JsonMixin, Pathlike, Seconds, SetContainingAnything, YamlMixin, fastcopy,
-                          split_sequence)
+                          split_sequence, uuid4)
 
 Channels = Union[int, List[int]]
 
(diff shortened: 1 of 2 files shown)


lhotse/audio.py
```diff
@@ -232,6 +270,45 @@ def split(self, num_splits: int, randomize: bool = False) -> List['RecordingSet'
             split_sequence(self, num_splits=num_splits, randomize=randomize)
         ]
 
+    def shard(self, locations: Sequence[Pathlike], include_source_dir: bool = True) -> 'RecordingSet':
+        """
```
danpovey, line 274: This would be very slow if done from one machine for one dataset.
I wonder if there could be an interface that was more friendly to doing this task from multiple jobs?
I wonder whether doing this by splitting up the RecordingSet into pieces and doing operations on the 
individual pieces, and then recombining them, might be more elegant in the sense that it would separate
concerns more.

danpovey: ... my general bias is towards having Lhotse behave in a fairly passive way, so the internal code mostly acts as a manifest; and we can do things like generate those manifests and manipulate them even if the underlying data doesn't exist yet; and then manually write to that data [or at least: have fairly outer-level code write to that data.]

I'm not sure if that's really practical or doable though in the current context; just putting it out there because in the past I've designed things and later wished they were more 'passive' because passive objects tend to be easier to manipulate and modify later on.

pzelasko: > This would be very slow if done from one machine for one dataset.
> I wonder if there could be an interface that was more friendly to doing this task from multiple jobs?
> I wonder whether doing this by splitting up the RecordingSet into pieces and doing operations on the
> individual pieces, and then recombining them, might be more elegant in the sense that it would separate
> concerns more.

We can either have a `num_jobs` argument (IMO sufficient in this case) or an `executor` argument. The `executor` is an API exposed by Python's stdlib https://docs.python.org/3/library/concurrent.futures.html that we're already using in the feature extraction code. It is convenient because some other external libraries implement it too (like dask) which allows you to leverage distributed execution with the same API (I am actually able to extract the features this way with SGE on the CLSP grid).

> ... my general bias is towards having Lhotse behave in a fairly passive way, so the internal code mostly acts as a manifest; and we can do things like generate those manifests and manipulate them even if the underlying data doesn't exist yet; and then manually write to that data [or at least: have fairly outer-level code write to that data.]
> 
> I'm not sure if that's really practical or doable though in the current context; just putting it out there because in the past I've designed things and later wished they were more 'passive' because passive objects tend to be easier to manipulate and modify later on.

That makes sense. I'm striving for a mix of that and easy discoverability. As a user, you're able to use these classes exclusively as manifests and do your own manipulaton on them (write/read/modify) since practically nothing is private. On the other hand, a lot of convenience methods are already there and you don't have to search hard to find them - your IDE/notebook autocomplete will list most of them as soon as you instantiate the object (to some extent I'm inspired by pandas API). But because all these methods use public data of the class, you are free to implement your own way to do any of these things if the "standard" way is not adequate.

pzelasko: Oh, it's also possible to do split - shard - combine with this API, by providing a 1-element list of locations to the shard call... I just think it might be a pretty common thing to do and I'd rather have it take 1 line rather than 5-6 lines of code in the experiment script.

danpovey:
OK!  You decide-- was just bringing up the issue.

pzelasko:
Thanks, appreciate it :)

pzelasko:
I have no interest in this for now; if somebody needs it feel free to pick it up

PR #101, opened 2020-10-28
Initial K2SpeechRecognitionDataset implementation

pzelasko:


pzelasko:
Once the Dataset works (at least in some basic way) I can look into any potential issues with the collate_fn in the DataLoader (we may need to override the default to allow for k2.Fsa batching, probably not too complicated?)

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,122 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
...
+        features = torch.from_numpy(cut.load_features())
+        mask = torch.from_numpy(cut.supervisions_feature_mask())
+
+        sil_id = self.symbol_table.get('<sil>')
+        eps_id = self.symbol_table.get('<eps>')
+
+        supervision_fsas = []
+        for supervision in cut.supervisions:
+            # TODO(pzelasko): I am not sure how to attach the symbol table to the FSA object
+            #                 (or whether that's need at all?)
```
csukuangfj, line 132: Symbol tables are only for visualization. 
If you like, you could attach a symbol table to an FSA using the following code

```python
fsa.symbols = self.symbol_table
```

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,122 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
...
+            # Creating the core supervision FSA corresponding to the supervision text.
+            fsa = k2.linear_fsa([self.symbol_table.get(c) for c in supervision.text])
+
+            # TODO: Adding optional silence at beginning and end.
+
+            # TODO(pzelasko): Should be possible to create it with sth like:
+            #   silfsa = k2.union(
+            #     k2.linear_fsa([sil_id]), k2.linear_fsa([eps_id])
+            #   ).closure()
+            silfsa = k2.Fsa.from_str(f'''0 1 {sil_id} 0
```
csukuangfj, line 143: This string is in the OpenFST format. Please use `k2.Fsa.from_openfst`.

`k2.Fsa.from_str` accepts strings only in k2 format.

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,122 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
...
+                                         0 1 {eps_id} 0
+                                         1 0 {sil_id} 0
+                                         1''')
+            # TODO(pzelasko): is it possible to do:
```
csukuangfj, line 147: @danpovey 
Do we need to implement `concat` in k2?

danpovey: We'll definitely need `concat` at some point.  
If we use the L.fst that comes from kaldi, I don't think  we'd need the `silfsa` because L.fst provides the optional silence itself.

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,122 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
 
     def __len__(self) -> int:
         return len(self.cut_ids)
+
```
danpovey, line 61: Thanks...
What about collation?  How would one use this from PyTorch (I am not super familiar with such things).

pzelasko: We'll have to provide a custom collate_fn to the DataLoader that converts multiple `Fsa`s into `FsaVec` if I understood K2 approach correctly (or a list of `Fsa`? not sure what's more convenient). Fangjun helped me with K2 installation so I'll play around with that today.

lhotse/dataset/speech_recognition.py
pzelasko: Is K2 format documented somewhere? I thought it's `<from-state> <to-state> <label> <weight>`

csukuangfj: The documentation is at: https://github.com/k2-fsa/k2/blob/3363292e3b2d4554eee025d2950a4c4ac20b6c01/k2/python/k2/fsa.py#L147

CAUTION: Arcs entering the final state have label `-1`.

```
       The given string `s` consists of lines with the following format:
        (1) When it represents an acceptor:
                src_state dest_state label score
        (2) When it represents a transducer:
                src_state dest_state label aux_label score
        The line for the final state consists of only one field:
                final_state
        Note:
          Fields are separated by space(s), tab(s) or both. The `score`
          field is a float, while other fields are integers.
        Caution:
          The first column has to be non-decreasing.
        Caution:
          The final state has the largest state number. There is only
          one final state. All arcs that are connected to the final state
          have label -1.
```

lhotse/dataset/speech_recognition.py
pzelasko: In that case, do you want to use phones to create the supervision FSAs from utterances (i.e. create the word-level symbol table and compose with L.fst before returning from __getitem__)?

danpovey:
My preference is that you use a simple collation method that just turns it
into a simple format like vector of 4-tuple:
  (text, sequence_id, begin_frame, end_frame)
where sequence_id is just the first index of the tensor containing the
frames, i.e. 0, 1, 2 for the 0,1,2'th elements of the
minibatch.
Then we can process that using k2 into the FSAs we want, in a way that
depends on the setup.

pzelasko:
Sure, I can do that. I think a more "py-torchy" way would be to create dataset classes that depend on the setup (possibly a hierarchy of them) and hold vocabularies, lexicons, etc. and prepare the FSAs for you, but I don't mind doing it this way.

danpovey:
Ah I see what you mean.  Let's consider that when k2 is a bit more
finished, as the code necessary for that may not exist yet.

pzelasko:
I've changed the Dataset to behave as you suggested; I've also added a `K2DataLoader` that collates this dataset correctly.

pzelasko:
@danpovey will that work with you guys?

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,108 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
...
+    The PyTorch Dataset for the speech recognition task using K2 library.
+    Each item in this dataset is a dict of:
+
+    .. code-block::
+
+        {
+            'features': (T x F) tensor,
+            'supervisions': List[Dict] -> [
+                {
+                    'example_idx': int
```
danpovey, line 74: what does the example_idx represent?

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,108 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        if 'collate_fn' in kwargs:
+            raise ValueError('Cannot override collate_fn in K2DataLoader.')
+        super().__init__(*args, collate_fn=multi_supervision_collate_fn, **kwargs)
+
+
+def multi_supervision_collate_fn(batch: Sequence[Dict]) -> Dict:
+    """Custom collate_fn for K2SpeechRecognitionDataset. See its docs for details."""
```
danpovey, line 143: It's not clear to me, when reading "See its docs for details", where to go.  I wonder if you could point more specifically where to look, or maybe briefly summarize it here?

lhotse/dataset/speech_recognition.py
```diff
@@ -57,3 +58,108 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
...
+            d[key] = value
+        return d
+
+    supervisions = default_collate([
+        update(sup, example_idx=dataset_idx_to_batch_idx[sup['example_idx']])
+        for example in batch
+        for sup in example['supervisions']
+    ])
+    feats = default_collate([example['features'] for example in batch])
+    return {
```
danpovey, line 162: I wonder if you could explain what this function returns, in terms of tensors and their dims, or whatever?  It's not super easy to unpack from the code.

csukuangfj: The return value can be used in k2 to construct a dense fsa:

```python
def dense_fsa(log_probs: torch.Tensor,
              supervision_segments: torch.Tensor) -> _k2.DenseFsaVec:
    '''Construct a DenseFsaVec from neural net log-softmax outputs.
    Args:
      log_probs:
        A 3-D tensor of dtype ``torch.float32`` with shape ``(N, T, C)``, where
        ``N`` is the number of sequences, ``T`` the maximum input length,
        and ``C`` the number of output classes.
      supervision_segments:
        A 2-D CPU tensor of dtype ``torch.int32`` with 3 columns.
        Each row contains information for a supervision segment. Column 0 is the
        ``sequence_index`` indicating which sequence this segment comes from;
        column 1 specifies the ``start_frame`` of this segment within the
        sequence; column 2 contains the ``duration`` of this segment.
```

`feats` can be used as input to a neural network and its output is `log_probs`.

`supervisions` contains information for constructing `supervision_segments`:
 - column 0 of `supervision_segments` is from `supervisions.example_idx[i]`
 - column 1 is from `supervisions.start_frame[i]`
 - column 2 is from `supervisions.end_frame[i]`

`supervisions.text[i]` can be used to construct a linear FSA.

lhotse/dataset/speech_recognition.py
csukuangfj: The docs is at https://github.com/lhotse-speech/lhotse/blob/8ba0eb64e0110dbacf913ea227e00bcc0f2b11e6/lhotse/dataset/speech_recognition.py#L66-L80

The input argument `batch` is a list of the above dict.

lhotse/dataset/speech_recognition.py
danpovey: Thanks!
I wonder whether it would make sense to have the collate_fn directly create the supervision_segments tensor of shape (num_supervisions,3).  Not sure if that's considered py-torchy or not.

csukuangfj: I think it is possible. The returned `example_idx`, `start_frame` and `end_frame` are just 1-D torch.Tensors, we
can `stack` them and make a transpose to get a tensor of shape `(num_supervisions, 3)`.

danpovey:
OK, thanks.  We'll let Piotr decide whether it makes more sense to leave it
as-is or change it.

danpovey:
Fangjun, something  you might want to work on is figure out how to turn the
list of 'text' strings into a ragged list of integers and from there into
linear FSAs.

danpovey:
This is OK as far as I am concerned.  I'm inclined to have you merge it to make it easier for us to use this.

jimbozhang:
@pzelasko @danpovey Do you think it is a good idea to merge `K2SpeechRecognitionDataset` and `SpeechRecognitionDataset` into one? I can do it if needed.

lhotse/dataset/speech_recognition.py
pzelasko: Good point, I'll add the relevant dict structure description in that place. Writing clear docs is not easy, so I appreciate this kind of feedback :)

lhotse/dataset/speech_recognition.py
pzelasko: I think it's fine, I'll make these changes. I'll rename example_idx to sequence_idx for consistency with K2 (given what I see in `dense_fsa()` doc-string). Also, one mismatch I noted is that I'm returning `end_frame`, and the doc provided by Fangjun says it expects `duration` - should I change this?

pzelasko:
@jimbozhang I think we should actually remove `SpeechRecognitionDataset` after this is merged, as it is a toy-example Dataset class that likely won't be used with any ASR framework. It was very useful to extend into a K2 dataset, but if we want to add support for other toolkits, we can use the K2 dataset as the starting point going forward.

lhotse/dataset/speech_recognition.py
csukuangfj: > and the doc provided by Fangjun says it expects duration - should I change this

Yes, that would be nice.

PR #104, opened 2020-11-03
TED-LIUM v3 manifests preparation

pzelasko:


pzelasko:
I'll merge as it shouldn't break anything - we can make changes later as needed.

PR #109, opened 2020-11-03
Add the mobvoi wake word data example

freewym:
drafting a version without data augmentation first...

### DIFF
--- a/lhotse/recipes/mobvoihotwords.py
+++ b/lhotse/recipes/mobvoihotwords.py
@@ -1,0 +1,145 @@
+import json
+import logging
+import shutil
+import tarfile
+import urllib.request
+from collections import defaultdict
+from pathlib import Path
+from typing import Dict, NamedTuple, Optional, Union
+
+import torchaudio
+
+from lhotse.audio import AudioSource, Recording, RecordingSet
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike
+
+
+def download_and_untar(
+        target_dir: Pathlike = '.',
+        force_download: Optional[bool] = False,
+        base_url: Optional[str] = 'http://www.openslr.org/resources'
+) -> None:
+    """
+    Downdload and untar the dataset
+
+    :param target_dir: Pathlike, the path of the dir to storage the dataset.
+    :param force_download: Bool, if True, download the tars no matter if the tars exist.
+    :param base_url: str, the url of the OpenSLR resources.
+    """
+
+    url = f'{base_url}/87'
+    target_dir = Path(target_dir)
+    target_dir.mkdir(parents=True, exist_ok=True)
+    dataset_tar_name = 'mobvoi_hotword_dataset.tgz'
+    resources_tar_name = 'mobvoi_hotword_dataset_resources.tgz'
+    for tar_name in [dataset_tar_name, resources_tar_name]:
+        tar_path = target_dir / tar_name
+        if force_download or not tar_path.is_file():
+            urllib.request.urlretrieve(f'{url}/{tar_name}', filename=tar_path)
+        corpus_dir = target_dir / 'MobvoiHotwords'
+        extracted_dir = corpus_dir / tar_name[: -4]
+        completed_detector = extracted_dir / '.completed'
+        if not completed_detector.is_file():
+            shutil.rmtree(extracted_dir, ignore_errors=True)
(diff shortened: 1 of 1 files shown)


pzelasko:
Thanks! I'll review it later today or tomorrow. We'll be refining data augmentation in Lhotse soon to leverage recently added support for SoX transforms in torchaudio (I'm also thinking about simplifying the interface to use arbitrary python function for augmentation).

freewym:
@pzelasko @danpovey FYI, [here](https://github.com/freewym/espresso/compare/lhotse?expand=1) is an incomplete draft for the wake word detection task in Espresso (can be more generally for ASR), where I assume features have already been dumped into the disk as json.gz files. Need to figure out how to add data augmentations. k2 FST related things (supervision, loss, etc) haven't been added either.

pzelasko:
I took a quick peek, it looks good - I'll go through it in more detail tomorrow. We're currently using WavAugment which has to be installed separately ([there's a script for it here](https://github.com/lhotse-speech/lhotse/blob/master/tools/install_wavaugment.sh)) and then you can use `WavAugmenter` [from here](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/augmentation.py#L17) - a couple of lines below there are definitions of "standard" effect chain settings that I copied from the WavAugment repo examples.

freewym:
Ideally I also need speed perturbation and additive noises from another corpus (MUSAN). I will think about it.

pzelasko:
Mmm yeah, adding MUSAN to Lhotse is on our to-do list for some time now. If you could do it that would be great, otherwise I'll try to find the time this or the next week. 

As to speed perturbation you should be able to define it as part of the SoX effect chain in WavAugment - check out [their example notebook](https://github.com/facebookresearch/WavAugment/blob/master/examples/python/WavAugment_walkthrough.ipynb). EDIT: after I wrote this, I found out I actually already added it as "standard" effect chains [here](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/augmentation.py#L124) and [here](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/augmentation.py#L96))

Also, if you want to extract features and augment using the CLI, you can check out `lhotse feat extract --help` (its code is [here](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/bin/modes/features.py#L45)) - in that case if you want custom augmentation, you'd need to define a new augmentation effect chain and register it (it'll be simple, see e.g. [here](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/augmentation.py#L124) for an example)

lhotse/recipes/mobvoihotwords.py
```diff
@@ -0,0 +1,145 @@
+import json
+import logging
+import shutil
```
danpovey, line 3: I'm just wondering whether it would be obvious to the user how to use this.
I guess there is the same issue in the other recipes... dunno whether this could be addressed with a README, or a pointer at the top of each file, or something?  I'm guessing there must be a standard pattern.

danpovey:
Great!
When the espresso thing is ready we should probably have a link from the lhotse code/recipe to the espresso stuff, so it's clear how it's supposed to be used.

lhotse/recipes/mobvoihotwords.py
pzelasko: In the other files, we inserted a module-level docstring on top as a description. Yiming, could you add it? 

In the wider context of Lhotse, it's not ideal - there are at least two functions that should have the same description in their docstrings (`prepare_X` and the CLI `lhotse prepare X`). I'll add an issue, I think we can do it with decorators that attach docstrings dynamically (similarly like the `transformers` library does [here](https://github.com/huggingface/transformers/blob/master/src/transformers/modeling_longformer.py#L1071))

pzelasko:
This one looks good - it's just the description thing that's missing. I tried to leave some comments in the Espresso PR, but you haven't actually created it so it's not possible - could you do it and let me know then? @freewym

freewym:
@pzelasko One issue for this PR: I feel the corpus downloading takes long time and there are nothing being printed during the time. Maybe it would be better to add something like a progress bar. But I didn't figure out how to do it. Do you happen to know anything about it? It can probably be a universal feature for all recipes

pzelasko:
That's a great point. I will make an issue to do that.

pzelasko (approved):

PR #114, opened 2020-11-04
fix the name overloading (features->feats)

freewym:


pzelasko:
Thanks! I somehow overlooked it and I admit that overloading is very confusing, but it's funny that it seems to have worked fine, as the unit tests were passing 😁

pzelasko (approved):

Issue #113, opened 2020-11-04
Map-style vs Iterable-style Dataset


pzelasko:
I want to start a discussion to determine what's the right way to go forward with data loading efficiency and flexbility for Lhotse PyTorch datasets.

[PyTorch documentation for dataset API](https://pytorch.org/docs/stable/data.html) describes two types of datasets - map-style and iterable-style. Short summary:

### Map-style:

```python
for index in sampler:
    sample = dataset[index]
```

- loads a single item at a time
- `DataLoader` samples indices of the items to be loaded (using a `Sampler`) based on `len()` of the dataset
- `DataLoader` specifies a fixed batch size and takes care of collation (possibly with a custom `collate_fn`)

### Iterable-style:

```python
for batch in iter(dataset):  # batch/sample depending on the implementation
    pass
```

- may load multiple items at the same time
- has to care of collation, shuffling, batch size (can be dynamic), etc. itself
- does not have to specify `len()`

### Comments 

Our current examples are based on the map-style paradigm, which offloads more things to the DataLoader default settings. I'm wondering if we should explore the iterable-style dataset more going forward. It seems like it might make it easier to work with sharded data, or support non-random-access reading (e.g. block-random sampling for batches). I think it'd be good to start collecting some insights, experiences and requirements to make sure the design is solid.

Another question is should Lhotse concern itself with that, and if so, to what extent? I don't think we'll be able to provide an optimal solution for every use case, but I think it'd be good to provide an option that (perhaps with some tuning) is at least okay in most compute infrastructures. We'll also gain more insights regarding that from building and running K2 recipes.

qindazhu:
+1.

I believe iterable-style is very useful in scenario where the data is arriving as as part of a stream or the users can not map each data point into an index, e.g. they cannot load all the data into memory, or even cannot figure out the `len()`  without loading data as the data source is a stream or a very large file, they also have no meta-file like scp file in Kaldi. 
I'm also wondering if there's any scenario that the data may come from different sources and we need to combine them together on-the-fly? BTW, handling source on a remote server is much complex, we may also need to support, e.g. read throughput (instead of read a single file each time as that will be too slow), I'm not sure how much effort we should make about this? or just for 1-2 specific remote storage format?

BTW, do we have any plan to support prefetch data? i.e. prefetch data for the next few batches while training with the current batch.

danpovey:
But the way lhotse works is to pre-compute manifests of everything and only
the actual computation happens on the fly.
[...]

PR #116, opened 2020-11-05
Iterable Dataset for K2 ASR

### LINKED ISSUE #115: Padding in librispeech example
Piotr, in here
https://github.com/lhotse-speech/lhotse/blob/master/examples/librispeech/librispeech.ipynb
there is a call to .pad(), which I think pads all cuts to the longest utterance there- I assume that would be quite
inefficient when it comes time to train?
Is it possible to delay padding until after we form minibatches, and form minibatches in a more efficient way so that
things of a like size go together?

pzelasko:
This PR explores how to go forward with #113 and should resolve #115 

I managed to implement a first, working version of the iterable ASR dataset for K2. I also tested it with LibriSpeech train-clean-100 and discovered that I failed to guarantee that `num_frames` field in the manifests always returns correct values (this is mostly related to incosistent rounding behaviour for durations like 16.475). I put some workarounds in the code so that it can collate without issues (a dataloader on train-clean-100 works without issues). I will resolve them at some other time...

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -8,7 +8,6 @@ from typing import Any, Callable, Dict, FrozenSet, Iterable, List, Optional, Seq
 import numpy as np
 from cytoolz import sliding_window
 from cytoolz.itertoolz import groupby
-from tqdm import tqdm
 
 from lhotse import WavAugmenter
 from lhotse.audio import AudioMixer, Recording, RecordingSet
@@ -1288,7 +1287,7 @@ class CutSet(JsonMixin, YamlMixin, Sequence[AnyCut]):
                     mix_eagerly=mix_eagerly
                 )
             )
-        cut_set = CutSet.from_cuts(tqdm(f.result() for f in futures))
+        cut_set = CutSet.from_cuts(f.result() for f in futures)
         return cut_set
 
     def with_features_path_prefix(self, path: Pathlike) -> 'CutSet':
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -273,7 +273,7 @@ class Cut(CutUtilsMixin):
 
     @property
     def num_frames(self) -> Optional[int]:
-        return round(self.duration / self.frame_shift) if self.has_features else None
+        return self.features.num_frames if self.has_features else None
 
     @property
     def num_samples(self) -> Optional[int]:
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -259,7 +259,7 @@ class Cut(CutUtilsMixin):
 
     @property
     def end(self) -> Seconds:
-        return round(self.start + self.duration, ndigits=3)
+        return round(self.start + self.duration, ndigits=8)
 
     @property
     def has_features(self) -> bool:
@@ -375,9 +375,9 @@ class Cut(CutUtilsMixin):
         if duration_past_end > 0:
             # When the end of the Cut has been exceeded, trim the new duration to not exceed the old Cut's end.
             new_duration -= duration_past_end
-        # Round the duration to 1ms to avoid the possible loss of a single audio sample due to floating point
+        # Round the duration to avoid the possible loss of a single audio sample due to floating point
         # additions and subtractions.
-        new_duration = round(new_duration, ndigits=3)
+        new_duration = round(new_duration, ndigits=8)
         new_time_span = TimeSpan(start=0, end=new_duration)
         criterion = overlaps if keep_excessive_supervisions else overspans
         new_supervisions = (segment.with_offset(-offset) for segment in self.supervisions)
@@ -616,7 +616,7 @@ class MixedCut(CutUtilsMixin):
     @property
     def duration(self) -> Seconds:
         track_durations = (track.offset + track.cut.duration for track in self.tracks)
-        return round(max(track_durations), ndigits=3)
+        return round(max(track_durations), ndigits=8)
 
     @property
     def has_features(self) -> bool:
@@ -736,7 +736,7 @@ class MixedCut(CutUtilsMixin):
             return self
         total_num_frames = round(duration / self.frame_shift) if self.has_features else None
         total_num_samples = round(duration * self.sampling_rate) if self.has_recording else None
-        padding_duration = round(duration - self.duration, ndigits=3)
+        padding_duration = round(duration - self.duration, ndigits=8)
         return self.append(PaddingCut(
             id=str(uuid4()),
             duration=padding_duration,
@@ -1288,7 +1288,7 @@ class CutSet(JsonMixin, YamlMixin, Sequence[AnyCut]):
                     mix_eagerly=mix_eagerly
                 )
             )
-        cut_set = CutSet.from_cuts(f.result() for f in futures)
+        cut_set = CutSet.from_cuts(tqdm(f.result() for f in futures))
         return cut_set
 
     def with_features_path_prefix(self, path: Pathlike) -> 'CutSet':
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -405,8 +405,8 @@ class Cut(CutUtilsMixin):
         total_num_frames = compute_num_frames(duration=duration, frame_shift=self.frame_shift) \
             if self.has_features else None
         total_num_samples = round(duration * self.sampling_rate) if self.has_recording else None
-        padding_duration = round(duration - self.duration, ndigits=7)
-        return self.append(PaddingCut(
+        padding_duration = round(duration - self.duration, ndigits=8)
+        padded = self.append(PaddingCut(
             id=str(uuid4()),
             duration=padding_duration,
             num_features=self.num_features if self.features is not None else None,
@@ -415,6 +415,7 @@ class Cut(CutUtilsMixin):
             sampling_rate=self.features.sampling_rate if self.features is not None else self.recording.sampling_rate,
             use_log_energy=self.features.type in ('fbank', 'mfcc') if self.features is not None else False
         ))
+        return padded
 
     def map_supervisions(self, transform_fn: Callable[[SupervisionSegment], SupervisionSegment]) -> AnyCut:
         """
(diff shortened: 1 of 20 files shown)


pzelasko:
Note to self: for the purpose of checking if the total number of frames in batch is lesser than `max_frames`, I'm not taking padding into account atm - should probably change that.

lhotse/cut.py
```diff
@@ -37,6 +37,19 @@ class CutUtilsMixin:
...
+        """
+        Return the supervisions in this Cut that have modified time boundaries so as not to exceed
+        the Cut's start or end.
+
+        Note that when ``cut.supervisions`` is called, the supervisions may have negative ``start``
+        values that indicate the supervision actually begins before the cut, or ``end`` values
+        that exceed the Cut's duration (it means the supervision continued in the original recording
+        after the Cut's ending).
+        """
+        return [s.trim(self.duration) for s in self.supervisions]
```
danpovey, line 51: For ASR tasks, it would only be useful to train with trimmed cuts if we knew that they were trimmed (and on which ends) and could modify the supervision FSA accordingly.

lhotse/dataset/speech_recognition.py
```diff
@@ -60,6 +62,172 @@ def __len__(self) -> int:
         return len(self.cut_ids)
 
 
+class K2SpeechRecognitionIterableDataset(IterableDataset):
```
danpovey, line 65: I think i would be clearer to inherit from `torch.IterableDataset` explicitly, so readers will know it's not a k2 thing.

lhotse/dataset/speech_recognition.py
```diff
@@ -60,6 +62,172 @@ def __len__(self) -> int:
...
+    Dimension symbols legend:
+    * ``B`` - batch size (number of Cuts),
+    * ``S`` - number of supervision segments (greater or equal to B, as each Cut may have multiple supervisions),
+    * ``T`` - number of frames of the longest Cut
+    * ``F`` - number of features
+
+    The 'sequence_idx' field is the index of the Cut used to create the example in the Dataset.
+    """
+
+    def __init__(
```
danpovey, line 94: Would be nice to have some documentation for things like max_frames!

lhotse/cut.py
pzelasko: What do you think is the right way to return that information? It would be straightforward to return negative `start_frame` and `end_frame` exceeding the time dimension in the features, but I'm not sure if that is sufficient for K2 to adjust the supervision FSA (or how it would be adjusted). I'm curious what're your thoughts on this.

Alternatively, we can put an assertion in this dataset to ensure that there are no supervisions that need to be trimmed (at least for now).

danpovey: k2 would probably insert optional entry points and exit points.  pair of bool might be enough.
But I think it's better, for now, to just asser that this doesn't happen.  It's not something I've ever done in ASR training (cutting utterance without knowing the alignment).

pzelasko:
OK, I added documentation and implemented a first "smarter" batching approach. @danpovey please take a look if that works better.

The good news is that it seems to work OK when I iterate through `test-clean-100` and look at some batches (it also passes through the whole set without errors). 

The bad news is that I had to work around some off-by-one errors that are increasingly tough to address at the root of the problem. I think our assumption that `num_frames` and `duration` would have a simple relationship is not true; perhaps even not possible. It's quite difficult to account for float rounding (0.005) in mixed cuts, especially in cases where it consists of 3+ other cuts and there might be offsets that also have a 5 at the 3rd decimal place. I think I'm too tired ATM to come up with a systematic solution so I'm "fixing" it with tensor slicing during collation.

danpovey:
Isn't the num-frames also part of the metadata? ... perhaps that could be
used to decide everything?  I mean convert everything to frame counts at
the outset.

pzelasko:
It is, but at this moment there are some edge cases where that meta-data is incorrect. I resolved most of them, but mixed cuts that consist of multiple cuts seem to still suffer from that, and I need to have a fresh mind before I look at it again...

pzelasko:
Uh, it was simpler than I thought, just needed a fresh look on this. Anyway, I'm setting up a new issue to make randomized tests that check the invariants of the Cut classes - I want to be more confident that the code does what I think it does, even if it seems to work on LibriSpeech without issues now.

@danpovey please check if what the new dataset class is doing makes sense to you.

pzelasko:
FYI I ran it on the CLSP grid, loading `train-clean-100` data for 10 epochs with randomization and "packing" cuts together for more efficient batching, this is the benchmark of how long it took to read + collate data (each row represents an epoch, number of iterations == number of batches). The features were stored on NFS, in separate files per array with lilcom compression on. I used 8 DataLoader parallel workers.

```
1435it [01:01, 23.33it/s]
1434it [00:42, 33.66it/s]
1435it [00:37, 38.09it/s]
1434it [00:37, 38.03it/s]
1436it [00:35, 40.07it/s]
1434it [00:33, 42.45it/s]
1434it [00:34, 41.96it/s]
1433it [00:33, 42.89it/s]
1435it [00:32, 44.39it/s]
1435it [00:32, 43.57it/s]
```

The mean time spent waiting to get a batch was 26ms, and the median was 1ms. I expect it to be even less when the consumer thread will actually do some work in between of reading batches.

I also found one more issue with `MixedCut` num_frames that I patched. After that, the whole 10 data loading epochs passed without errors.

It's ready to merge from my side.

qindazhu:
+2, it would be great if you can update pip package after merging this PR,  as I want to use `K2SpeechRecognitionDataset` to build recipes (seems it was not exported in current pip package)

lhotse/cut.py
```diff
@@ -244,7 +258,7 @@ def recording_id(self) -> str:
 
     @property
     def end(self) -> Seconds:
-        return round(self.start + self.duration, ndigits=3)
+        return round(self.start + self.duration, ndigits=8)
```
danpovey, line 261: just wondering why we need to do rounding here... is there a rationale?

lhotse/cut.py
```diff
@@ -767,12 +801,19 @@ def load_features(self, mixed: bool = True) -> Optional[np.ndarray]:
...
+            #  relationship between num_frames and duration we strived for is not true;
+            #  i.e. the duration is 10.125 + 5.715 = 15.84, but the number of frames is
+            #  1013 + 572 = 1585. If the frame_shift is 0.01, we have gained an extra 0.01s...
+            if feats.shape[0] - self.num_frames == 1:
+                feats = feats[:self.num_frames, :]
+            # TODO(pzelasko): This can sometimes happen in a MixedCut with >= 5 different Cuts,
+            #   with a regular Cut at the end, when the mix offsets are floats with a lot of decimals.
+            #   For now we're duplicating the last frame to match the declared "num_frames" of this cut.
+            if feats.shape[0] - self.num_frames == -1:
+                feats = np.concatenate((feats, feats[-1:, :]), axis=0)
```
danpovey, line 816: Perhaps assert, after all of this, that feats.shape[0] == self.num_frames, so examples that differ by more than one crash early.

pzelasko:
Yeah, limited float arithmetic precision. I originally put 3 decimals there
without too much thought, but it was too coarse. With 8 I think it's enough
not to mess up the duration and num_frames/num_samples relationship. If I
don't round it, it will result in num_frames off by one errors in cuts with
durations like 10.444999999998...

Maybe it'd be cleaner to use the Decimal type instead? I think we'll run
into these problems again with high resolution recordings (44.1kHz and
higher fs).

pon., 9 lis 2020, 22:39 użytkownik Daniel Povey <notifications@github.com>
napisał:

danpovey:
Up to you; if it works, it's OK with me.

qindazhu:
Got below error when I was trying to use `K2SpeechRecognitionIterableDataset` with `Mfcc` feature,
```
>>> dataset = K2SpeechRecognitionIterableDataset(cuts_test, shuffle=False)  #cuts_test using Mfcc featrue extractor.
>>> dloader = DataLoader(dataset, batch_size=None, num_workers=num_workers)
>>> batch = next(iter(dloader))
  File "<stdin>", line 1, in <module>
  File "/home/linuxbrew/.linuxbrew/Cellar/python@3.8/3.8.6_1/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 363, in __next__
    data = self._next_data()
  File "/home/linuxbrew/.linuxbrew/Cellar/python@3.8/3.8.6_1/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 989, in _next_data
    return self._process_data(data)
  File "/home/linuxbrew/.linuxbrew/Cellar/python@3.8/3.8.6_1/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1014, in _process_data
    data.reraise()
  File "/home/linuxbrew/.linuxbrew/Cellar/python@3.8/3.8.6_1/lib/python3.8/site-packages/torch/_utils.py", line 395, in reraise
    raise self.exc_type(msg)
ValueError: Caught ValueError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/home/linuxbrew/.linuxbrew/Cellar/python@3.8/3.8.6_1/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 185, in _worker_loop
    data = fetcher.fetch(index)
  File "/home/linuxbrew/.linuxbrew/Cellar/python@3.8/3.8.6_1/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 34, in fetch
    data = next(self.dataset_iter)
  File "/ceph-hw/lhotse/lhotse/dataset/speech_recognition.py", line 206, in __next__
    features = _collate_features(cuts)
  File "/ceph-hw/lhotse/lhotse/dataset/speech_recognition.py", line 280, in _collate_features
    features[idx] = torch.from_numpy(cut.load_features())
  File "/ceph-hw/lhotse/lhotse/cut.py", line 791, in load_features
    mixer = FeatureMixer(
  File "/ceph-hw/lhotse/lhotse/features/mixer.py", line 48, in __init__
    self.reference_energy = feature_extractor.compute_energy(base_feats)
  File "/ceph-hw/lhotse/lhotse/features/base.py", line 100, in compute_energy
    raise ValueError('The feature extractor\'s "compute_energy" is undefined.')
ValueError: The feature extractor's "compute_energy" is undefined.
```
It's OK when using `Fbank()` together with `K2SpeechRecognitionIterableDataset`,Maybe it's better to add some documentation about this to say we can't use Mfcc for `K2SpeechRecognitionIterableDataset` (or there is already one but I don't know)?

danpovey:
It might be that you were missing some args or options.   Maybe search in
lhotse code for examples, e.g. git grep Mfcc ?

danpovey:
Oh I see, it's missing code, not options.  It should be possible to compute
the energy, by reversing the transform.
However there's probably no harm in using the fbank features.  We can
always decorrelate them by applying the
cosine transform and then batchnorm.

pzelasko:
@qindazhu You'll getting this error because we haven't implemented feature-domain mixing for MFCC features. I think I attempted it, but was getting too big a difference between the features extracted from mixed audio vs. mixing the features extracted from separate audios, and figured out we can work on it later if needed. If you need it, can you open an issue?

pzelasko:
Anyway, the least I can do for now is to improve the error message before merging.

qindazhu:
Not urgent, we can use `Fbank` for now. Thanks for the reply, I'm just learning to use lhotse.

pzelasko:
Re: the num_frames and rounding thing, I am not 100% happy with the current solution but it seems to work for now so I'll address the review remarks and then merge. I'm thinking going forward maybe we should count the offsets and durations in samples rather than seconds, thus avoiding the float precision problems. If you know the sampling rate and frame shift, you can always derive duration and num frames from that info precisely. That'd be a fairly big change so I'd like to know your opinions guys before attempting it.

danpovey:
For things like MFCCs that don't have samples (assuming you've already discarded the wave data) I'm not sure if it makes sense to count samples?   Presumably if you appended those at the feature level it would be based on the frame count.  Maybe the problem there, if you want to be super exact about the length, is that an object like that really has 2 durations, one if you consider it a sequence of frames and another if you consider it a sequence of samples.  And of course if you were to resample, the length would change slightly too.

pzelasko:
Maybe you're right. I'll let it be for some time. Then I'll add the randomized tests and we'll see what kinds of issues come up.

PR #118, opened 2020-11-08
Fix for libsox build for WavAugment installation

pzelasko:
@freewym check this out and let me know if you're still having issues on CLSP grid - it works well for me.


freewym:
Yes, it works!

Issue #121, opened 2020-11-09
Feature extraction got stuck when both augmenter and executor is not None


freewym:
When `augmenter` is not None, the feature extraction seems to get stuck (meaning that it never enters into `_extract_and_store_features_helper_fn`) at the line below

https://github.com/lhotse-speech/lhotse/blob/master/lhotse/cut.py#L1263

There is no problems if `augmenter` is None, or `executor ` is None

pzelasko:
Thanks I'll check it out tomorrow. Might have to wait a bit as our grid
will be under maintenance till 3pm.

pon., 9 lis 2020, 18:55 użytkownik Yiming Wang <notifications@github.com>
napisał:

PR #122, opened 2020-11-10
Test for wav augment + parallel executor

pzelasko:


pzelasko:
@freewym can you check if this test runs alright for you? I tested both on my laptop and the CLSP grid, and there seems to be no issue.

freewym:
It's alright for me

pzelasko:
Ok. How can I reproduce issue #121 ?

freewym:
https://github.com/freewym/espresso/pull/45/files

checkout this PR; install  Espresso as in README;  `make lhotse`;  run `run.sh`

freewym:
The issue is gone if lambda expressions are replaced by functions (e.g, replace 
 lambda at this line https://github.com/freewym/espresso/pull/45/files#diff-3724bd15a663fd58b5fc92daa096bc9eb0c0a96a4c070a1a9e999be0682c985aR319 with `_get_factor()`, and `_get_factor()' should be defined outside of `speed()`)

I don't know why this matters...

danpovey:
How did you figure that out?

freewym:
Well, just gradually change from a working example to the problematic one...

pzelasko:
AFAIK It's related to how Python does multi processing. It forks the main
process and sends all data via pickles. Lambdas with captures are not
pickle-able, but functions are (don't ask me why...) although I would have
expected it to crash rather than hang.

pon., 9 lis 2020, 23:59 użytkownik Yiming Wang <notifications@github.com>
napisał:

pzelasko:
Oh damn now this test is hanging in the CI. When I have some more time again I will just add support for torchaudio, I'm tired of working around WavAugment quirks...

mthrok:
Let me know if you need help for `torchaudio`.

Note that `torchaudio` also has a similar edge case where transform implemented in C++ hangs. It only happens on macOS, Python 3.6/3.7 and when the function is called in subprocess. It's being tracked here.
https://github.com/pytorch/pytorch/issues/46409

pzelasko:
ooh la la... thanks for letting us know. That libsox turns out to be quite naughty...

PR #124, opened 2020-11-11
Augmentation refactoring and torchaudio SoX effects support

pzelasko:
TL;DR 

- Changing the data augmentation APIs in Lhotse to accept a callable with signature like: `def augment_fn(audio: Union[torch.tensor, np.ndarray], sampling_rate: int) -> np.ndarray`
- mirroring `WavAugment` capabilities with `torchaudio.sox_effects`

### DIFF
--- a/lhotse/augmentation/__init__.py
+++ b/lhotse/augmentation/__init__.py
@@ -1,3 +1,3 @@
 from .common import AugmentFn
 from .torchaudio import *
-from .wavaugment import *
+from .wavaugment import WavAugmenter, is_wav_augment_available
--- a/lhotse/augmentation/__init__.py
+++ b/lhotse/augmentation/__init__.py
@@ -3,5 +3,5 @@ import torchaudio as _torchaudio
 from .common import AugmentFn
 from .wavaugment import *
 
-if _torchaudio.__version__ >= '0.7.0':
+if str(_torchaudio.__version__) >= '0.7.0':
     from .torchaudio import *
--- a/lhotse/augmentation/__init__.py
+++ b/lhotse/augmentation/__init__.py
@@ -1,13 +1,3 @@
-import torchaudio as _torchaudio
-from packaging.version import parse as _parse
-
 from .common import AugmentFn
+from .torchaudio import *
 from .wavaugment import *
-
-# Note: we cannot directly compare the Version objects return from _parse because
-# Anaconda torchaudio has a version string '0.7.0a0+ac17b64' that is interpreted as
-# lesser than 0.7.0.
-_ta_version = _parse(_torchaudio.__version__)
-_req_version = _parse('0.7')
-if _ta_version.major >= _req_version.major and _ta_version.minor >= _req_version.minor:
-    from .torchaudio import *
--- a/lhotse/augmentation/__init__.py
+++ b/lhotse/augmentation/__init__.py
@@ -1,7 +1,13 @@
 import torchaudio as _torchaudio
+from packaging.version import parse as _parse
 
 from .common import AugmentFn
 from .wavaugment import *
 
-if str(_torchaudio.__version__) >= '0.7.0':
+# Note: we cannot directly compare the Version objects return from _parse because
+# Anaconda torchaudio has a version string '0.7.0a0+ac17b64' that is interpreted as
+# lesser than 0.7.0.
+_ta_version = _parse(_torchaudio.__version__)
+_req_version = _parse('0.7')
+if _ta_version.major >= _req_version.major and _ta_version.minor >= _req_version.minor:
     from .torchaudio import *
--- a/lhotse/augmentation/__init__.py
+++ b/lhotse/augmentation/__init__.py
@@ -1,0 +1,7 @@
+import torchaudio as _torchaudio
+
+from .common import AugmentFn
+from .wavaugment import *
+
+if _torchaudio.__version__ >= '0.7.0':
+    from .torchaudio import *
--- a/lhotse/augmentation/torchaudio.py
+++ b/lhotse/augmentation/torchaudio.py
@@ -1,4 +1,3 @@
-import random
 import warnings
 from dataclasses import dataclass
 from typing import List, Union
@@ -21,7 +20,7 @@ class RandomValue:
     end: Union[int, float]
 
     def sample(self):
-        return random.uniform(self.start, self.end)
+        return np.random.uniform(self.start, self.end)
 
 
 # Input to the SoxEffectTransform class - the values are either effect names,
--- a/lhotse/augmentation/torchaudio.py
+++ b/lhotse/augmentation/torchaudio.py
@@ -52,10 +52,22 @@ class SoxEffectTransform:
             tensor = torch.from_numpy(tensor)
         effects = self.sample_effects()
         augmented, new_sampling_rate = torchaudio.sox_effects.apply_effects_tensor(tensor, sampling_rate, effects)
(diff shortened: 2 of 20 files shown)


pzelasko:
It's good to review.

pzelasko:
@freewym please check this out - it will be small but breaking for the recipe you're creating, but hopefully makes the whole setup way easier (no need to compile libsox, wavaugment, etc. just install the latest pytorch + torchaudio with anaconda) and gets us rid of various quirks with multiprocessing.

mthrok (commented):
Looks good regarding the usage of torchaudio's Sox Effects.

vincentqb:
(glad to see the migration here! cc facebookresearch/WavAugment#16)

freewym:
> @freewym please check this out - it will be small but breaking for the recipe you're creating, but hopefully makes the whole setup way easier (no need to compile libsox, wavaugment, etc. just install the latest pytorch + torchaudio with anaconda) and gets us rid of various quirks with multiprocessing.

Cool. I think I am supposed to install PyTorch 1.7 in order to test it. I will do later today.

lhotse/augmentation/__init__.py
```diff
@@ -0,0 +1,7 @@
+import torchaudio as _torchaudio
+
+from .common import AugmentFn
+from .wavaugment import *
+
+if str(_torchaudio.__version__) >= '0.7.0':
```
mthrok, line 6: FYI: If `torchaudio` hits `0.10.0` release in the future (we do not know if we will move to the major 1.0 release or not), this could produce a wrong result.

```
$ python
>>> '0.7.0' > '0.10.0'
True
```

A future-proof way would be to split the version string and compare major version and minor version as number.

lhotse/features/base.py
```diff
@@ -10,7 +10,7 @@
 import torch
 
 from lhotse.audio import Recording
-from lhotse.augmentation import WavAugmenter
+from lhotse.augmentation import AugmentFn, WavAugmenter
```
freewym, line 13: WavAugmenter is no longer useful (?) If it is the case, all appearances of WavAugmenter should be removed within this file

pzelasko: Good point. I want to keep it for now so that people with PyTorch older than 1.7 can also use it, but I will probably add a deprecation warning...

lhotse/augmentation/__init__.py
pzelasko: Well spotted!

lhotse/augmentation/torchaudio.py
```diff
@@ -0,0 +1,111 @@
...
+
+def speed(sampling_rate: int) -> List[List[str]]:
+    return [
+        # Random speed perturbation factor between 0.9x and 1.1x the original speed
+        ['speed', RandomValue(0.9, 1.1)],
+        ['rate', sampling_rate],  # Resample back to the original sampling rate (speed changes it)
+    ]
+
+
+def reverb(sampling_rate: int) -> List[List[str]]:
```
freewym, line 99: Is it possible to make such functions more general, in that they can accept more arguments, e.g., the lower/up bound of room sizes can been passed into this function?

lhotse/augmentation/torchaudio.py
```diff
@@ -0,0 +1,111 @@
...
+@dataclass
+class RandomValue:
+    """
+    Represents a uniform distribution in the range [start, end].
+    """
+    start: Union[int, float]
+    end: Union[int, float]
+
+    def sample(self):
+        return random.uniform(self.start, self.end)
```
freewym, line 24: can we use numpy random functions? It may be easier for me to seed it from outside

lhotse/augmentation/torchaudio.py
pzelasko: Sure, I'll change that

lhotse/augmentation/torchaudio.py
pzelasko: Would using this function help (note that it also makes the random cut ID creation deterministic)? 

https://github.com/lhotse-speech/lhotse/blob/master/lhotse/utils.py#L33

freewym: OK, never mind. I will seed it in my own code.

lhotse/augmentation/torchaudio.py
```diff
@@ -0,0 +1,111 @@
...
+            ]
+            for effect in self.effects
+        ]
+
+
+def speed(sampling_rate: int) -> List[List[str]]:
+    return [
+        # Random speed perturbation factor between 0.9x and 1.1x the original speed
+        ['speed', RandomValue(0.9, 1.1)],
+        ['rate', sampling_rate],  # Resample back to the original sampling rate (speed changes it)
```
freewym, line 95: Looks like this line makes the running hang. It works without this line.

lhotse/augmentation/__init__.py
csukuangfj: `packaging.version` may be helpful in this case.
See https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

lhotse/augmentation/torchaudio.py
pzelasko: I'll change it to numpy anyway, I guess more people are used to seeding `numpy` than `random` in training loops

lhotse/augmentation/torchaudio.py
pzelasko: 😫

I'll have a look

lhotse/augmentation/torchaudio.py
pzelasko: Actually I'd rather make a follow-up PR later on, as I'm not sure which parameters it makes sense to tweak and how general they should be. If we want to tweak everything it's simpler to just write your own chain...

lhotse/augmentation/torchaudio.py
pzelasko: I was able to replicate the issue and add a unit test that causes it. I submitted the issue to torchaudio here: https://github.com/pytorch/audio/issues/1021

freewym: BTW, if replacing RandomValue() above with a function  `_get_value(factor)` where `factor` is simply returned, the running hangs as well.

pzelasko: Is the function defined as a closure (i.e. within another function) and captures some variable outside of its scope? That could explain it... Otherwise, I don't know.

freewym: No, it's a independent function outside of any other functions.

https://github.com/freewym/espresso/pull/45/commits/e995f0d61ea9777a366a5a2d3b0bbe9eef900169#diff-3724bd15a663fd58b5fc92daa096bc9eb0c0a96a4c070a1a9e999be0682c985aR298

pzelasko: @freewym some good news, if you create executor like: `ProcessPoolExecutor(..., mp_context=multiprocessing.get_context("spawn"))` it solves the segfault/hanging problem. Could you try? If it works I'll go on and merge this

pzelasko: Credits to @mthrok for suggesting this

pzelasko: (to make it clear: it works for me on the grid, on my mac, and in GitHub CI, so it should be okay)

freewym: Yeah, it works!

pzelasko: 🎉

Issue #129, opened 2020-11-16
Progress indicator


danpovey:
@pzelasko how hard would it be to implement a progress indicator?
IDK how that would interact with the multiprocessing?
Its a bit disconcerting to start running something when you have no idea how long it will take.

Issue #111, opened 2020-11-04
Progress bars for time-taking operations


pzelasko:
Lhotse's UX could be improved by leveraging some progress bar library (e.g. [tqdm](https://pypi.org/project/tqdm/)) in corpus downloading. Maybe it would also work for feature extraction. I'm not sure yet if any other operations would take that much time.

PR #131, opened 2020-11-16
Progress bars for feature extraction and downloads

pzelasko:
Ask, and ye shall receive.

Looks like this for downloads:
image

And like this for feature extraction:
image

In the terminal, it will automatically display a text progress bar instead.


pzelasko:
Resolves #129 
Resolves #111

pzelasko:
I'll just merge, if there are any comments we can do a follow-up PR

Issue #132, opened 2020-11-16
Read previously prepared manifests when available in prepare_X functions


pzelasko:
Per: https://github.com/k2-fsa/snowfall/issues/16

PR #133, opened 2020-11-17
Skip LibriSpeech manifest preparation if they can be read instead

pzelasko:
It addresses #132, but only for LibriSpeech, as I'm not sure ATM if this type of "caching" can be implemented in a recipe-independent way. It would probably be easier if every recipe explicitly declared the list of files that the `prepare_X` function creates. Maybe that's a good direction to move forward.


danpovey (commented):
LGTM, except I think the caching behavior should be documented in prepare_librispeech().

PR #134, opened 2020-11-17
Prevent too many files in a feature directory by using sub-directories

### LINKED ISSUE #130: Directory structure of features
I notice that by default lhotse seems to put all the .llc files in one directory.
On many (maybe even most) file systems this scales very poorly because you spend too much time manipulating the meta-info of the directory- particularly when writing.
I think a better default would be to use the first few characters of the hash as a sub-directory name.
Was thinking of making this dependent on how much data there is to process, but I'm concerned people will write their own scripts that assume a certain directory structure, and then get surprised when things are not as they expect.

pzelasko:
Resolves #130 

@danpovey check out if this is what you had in mind (note that no changes in the reading code are needed).

### DIFF
--- a/lhotse/features/io.py
+++ b/lhotse/features/io.py
@@ -203,11 +203,17 @@ class LilcomFilesWriter(FeaturesWriter):
         return self.storage_path_
 
     def write(self, key: str, value: np.ndarray) -> str:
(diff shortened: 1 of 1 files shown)


danpovey (commented):
LGTM!!

Issue #144, opened 2020-11-20
Feature request: Train test split of CutSet based on supervision ids instead of recordings


janvainer:
I have a list of utterance ids that should be in test set or train set. Unfortunately, the train test split may use utterances that share the same recording. To create test set cuts, I use the following:
```
cuts = CutSet.from_yaml("feats/features.yaml")
for cut in cuts:
    cut.supervisions = [supervision for supervision in cut.supervisions if supervision.id in test_ids]
```

Would it make sense to you to provide CutSet api for splitting based on supervision ids? For example:
```
set1, set2, ..., set3 = CutSet.split_on_supervisions(supervision_ids: Iterable[Iterable[str]])
```
This would be used like:
```
train, dev, test = cuts.split_on_supervisions([train_ids, dev_ids, test_ids])
```

pzelasko:
That's a good addition. I'd rather have it be more general by accepting a flat iterable, and then what you need could be achieved by:
```python
train = cuts.subset(supervision_ids=train_ids)
dev = cuts.subset(supervision_ids=dev_ids)
test = cuts.subset(supervision_ids=test_ids)
```

I was thinking we could eventually "mimic" Kaldi `subset_data_dir.sh` a bit, by naming this method as `subset()` and giving it a number of optional args to subset based on different criteria. We could start with what you are suggesting:

```python
def subset(supervision_ids: Optional[Iterable[str]] = None) -> CutSet:
```

Would you like to contribute that?

janvainer:
Hi, your version sounds better. I will look into it over the weekend and make a PR.

janvainer:
Was resolved by #145

PR #145, opened 2020-11-21
Add cutset subset

janvainer:
Preliminary PR for `CutSet.subset`. Solves #144 

1. fastcopy did not create correct copies of CutSet (test was failing). I used deepcopy instead. Other suggestions welcome.
2. Cuts without supervisions are removed

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -3,7 +3,6 @@ import warnings
 from concurrent.futures import ProcessPoolExecutor
 from dataclasses import dataclass, field
 from functools import reduce
-from copy import deepcopy
 from math import ceil, floor
 from typing import Any, Callable, Dict, FrozenSet, Iterable, List, Optional, Sequence, Union
 
@@ -449,6 +448,16 @@ class Cut(CutUtilsMixin):
         new_cut = fastcopy(self, supervisions=[s.map(transform_fn) for s in self.supervisions])
         return new_cut
 
+    def filter_supervisions(self, supervision_ids: Iterable[str]) -> AnyCut:
+        """
+        Modify cut to store only supervisions contained in `supervision_ids`
+
+        :param supervision_ids: an iterable of supervision ids to keep
+        :return: a modified Cut
+        """
+        new_cut = fastcopy(self, supervisions=[s for s in self.supervisions if s.id in set(supervision_ids)])
+        return new_cut
+
     @staticmethod
     def from_dict(data: dict) -> 'Cut':
         features = Features.from_dict(data.pop('features')) if 'features' in data else None
@@ -581,6 +590,15 @@ class PaddingCut(CutUtilsMixin):
         )
 
     def map_supervisions(self, transform_fn: Callable[[Any], Any]) -> AnyCut:
+        """
+        Just for consistency with `Cut` and `MixedCut`.
+
+        :param transform_fn: a dummy function that would be never called actually.
+        :return: the PaddingCut itself.
+        """
+        return self
+
+    def filter_supervisions(self, supervision_ids: Iterable[str]) -> AnyCut:
         """
         Just for consistency with `Cut` and `MixedCut`.
 
@@ -972,6 +990,18 @@ class MixedCut(CutUtilsMixin):
             track.cut.supervisions = [segment.map(transform_fn) for segment in track.cut.supervisions]
         return new_mixed_cut
 
+    def filter_supervisions(self, supervision_ids: Iterable[str]) -> AnyCut:
+        """
+        Modify cut to store only supervisions contained in `supervision_ids`
+
+        :param supervision_ids: an iterable of supervision ids to keep
+        :return: a modified Cut
+        """
+        new_mixed_cut = fastcopy(self)
+        for track in new_mixed_cut.tracks:
+            track.cut.supervisions = [s for s in track.cut.supervisions if s.id in set(supervision_ids)]
+        return new_mixed_cut
+
     @staticmethod
     def from_dict(data: dict) -> 'MixedCut':
         return MixedCut(id=data['id'], tracks=[MixTrack.from_dict(track) for track in data['tracks']])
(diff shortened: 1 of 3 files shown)


lhotse/cut.py
```diff
@@ -1123,6 +1124,41 @@ def split(self, num_splits: int, randomize: bool = False) -> List['CutSet']:
...
+        """
+
+        assert supervision_ids is not None, \
+            "supervision ids can not be None."
+
+        # use set for faster search
+        supervision_ids = set(supervision_ids)
+
+        # fails test if fastcopy is used
+        filtered_cutset = deepcopy(self)
```
pzelasko, line 1147: I think this can get pretty slow for large `CutSet`s; I'd suggest the following approach instead:

```
filtered_cutset = CutSet.from_cuts(
    cut for cut in self if any(s.id in supervision_ids for s in cut.supervisions)
)
```

BTW do you think that when a cut contains more than one supervision, and only one is in the supplied list, we should drop the supervisions outside of it? That would be more complex to handle as `MixedCut.supervisions` cannot be set directly (as in line 1150 below).

pzelasko: (I think to handle that properly we could add `.filter_supervisions` methods to each of the three `Cut` classes that returns a copy of the cut - maybe using `fastcopy` - with filtered supervisions; that'd be consistent with the `map_supervisions` method that I think is already there)

pzelasko (commented):
Thanks! I left some comments.

lhotse/cut.py
janvainer: Thanks, ok I will replace deepcopy by comprehension. :) regarding dropping supervisions not mentioned in supervision ids, I think dropping them entirely makes sense. filter_supervisions seems like a good solution for this given that Mixed cuts supervisions cannot be reassigned easily. I will look into it

janvainer: @pzelasko FYI I added `filter_supervisions` to all Cut variants and adjusted the subset method according to your comment.

lhotse/cut.py
```diff
@@ -448,6 +448,16 @@ def map_supervisions(self, transform_fn: Callable[[SupervisionSegment], Supervis
         new_cut = fastcopy(self, supervisions=[s.map(transform_fn) for s in self.supervisions])
         return new_cut
 
+    def filter_supervisions(self, supervision_ids: Iterable[str]) -> AnyCut:
```
pzelasko, line 451: I was thinking of a signature that more closely corresponds to the one of Python's `filter` function, i.e. `filter_supervisions(self, predicate: Callable[[SupervisionSegment], bool] -> Cut`

It's what we did with `map_supervisions` too. It allows to use an arbitrary filtering criterion. Examples of use:
```python
cut = cut.filter_supervisions(lambda s: s.id in supervision_ids)
cut = cut.filter_supervisions(lambda s: s.duration < 5.0)
cut = cut.filter_supervisions(lambda s: s.text is not None)
```

lhotse/cut.py
```diff
@@ -971,6 +990,18 @@ def map_supervisions(self, transform_fn: Callable[[SupervisionSegment], Supervis
...
+        """
+        Modify cut to store only supervisions contained in `supervision_ids`
+
+        :param supervision_ids: an iterable of supervision ids to keep
+        :return: a modified Cut
+        """
+        new_mixed_cut = fastcopy(self)
+        for track in new_mixed_cut.tracks:
+            track.cut.supervisions = [s for s in track.cut.supervisions if s.id in set(supervision_ids)]
```
pzelasko, line 1002:
```suggestion
            track.cut = track.cut.filter_supervisions(predicate)
```

pzelasko (commented):
Thanks! Left some more comments.

lhotse/cut.py
janvainer: Good idea, lets also make it for the `CutSet.filter_supervisions(predicate)` instead of `CutSet.subset(supervision_ids)`. 
Getting a train subset based on ids: `CutSet.filter_supervisions(lambda s: s.id in trainset_ids)`. WYT?

pzelasko: I am not sure TBH. I think we could add it, but the semantics are a bit different than those of `subset` - `filter_supervisions` only removes the supervisions, while `subset` would also remove the cuts. So we could have both actually. But I don't think that `subset` should accept `predicate` because we might want to add options like `n_shortest`, `n_longest`, etc. that are not easily expressed with a predicate.

janvainer: Um why would we want to keep cuts without supervisions if we remove all their supervisions by `.filter_supervisions`? It seems a bit cleaner if only one filtering option for supervisions exists. Having options such as `n_longest` would be useful though.  I pushed some updates - there is only `CutSet.filter_supervisions` right now, but we can easily revert back once the interface is clear. 

Is there some guideline on how to write tests for Lhotse? I adjusted the test for `CutSet.filter_supervisions` and added test for `Cut.filter_supervisions` Shoud they be adjusted somehow or moved somewhere else?

lhotse/cut.py
janvainer: changed it

lhotse/cut.py
pzelasko: In general a Cut without a supervision is a valid object in Lhotse, and it could be used for some kind of unsupervised learning, or as "in-domain" noise to be added to other cuts, etc. We might want to filter out some supervisions (e.g. too short, not containing text but e.g. indicating music/noise, etc.) and split the resulting CutSet into a supervised and unsupervised part, etc. 

I'm thinking of `filter_supervisions` as a "low/mid level" API (or a building block) that would likely be followed by other operations to make the CutSet actually useful, and of `subset` as a more "high level" API.

pzelasko: I haven't written down a guideline for tests yet but it seems like a good idea, I'll add it to my todo-list. I will review the PR later/tomorrow and let you know.

janvainer: Ok to sum up:

```
CutSet.subset(subset_ids: Iterable[str] = None, ...space for future args...)  # removes cuts with empty supervisions
CutSet.filter_supervisions(predicate)   # only removes supervisions, can contain unsupervised cuts
Cut.filter_supervisions(predicate)  # removes supervisions that do not satisfy predicate, may return Cut without supervision
``` 
:)

pzelasko:
Looks great! Sorry that I couldn't get back to you earlier - I see you figured out the tests well though :)

LGTM

pzelasko (approved):

PR #149, opened 2020-11-23
National Speech Corpus data prep, optimizations in Cut, limited export to Kaldi data dir

pzelasko:
I'm adding a data prep script for National Speech Corpus (thousands of hours of Singaporean English), I actually want to build a Kaldi system for it so I am adding a utility to export RecordingSet + SupervisionSet into a Kaldi data dir (it is limited but maybe a good starting point to extend for anybody who needs more). Since the recordings in this corpus have 2h, there is a lot of SupervisionSegments and it revealed a quadratic complexity in some CutSet operations - I am optimizing them by building an IntervalTree of supervisions that can be used for efficient queries about overlaps and over-spanning segments.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -104,11 +104,16 @@ class Recording:
     duration: Seconds
 
     @staticmethod
-    def from_sphere(sph_path: Pathlike, relative_path_depth: Optional[int] = None) -> 'Recording':
+    def from_sphere(
+            sph_path: Pathlike,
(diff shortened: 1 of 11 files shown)


pzelasko:
I think this is okay to merge; I might eventually extend the NSC data prep to other parts of the corpus.

janvainer:
Nice, export to kaldi datadir will be very handy! Thanks :)

PR #154, opened 2020-11-30
AMI recipe extension to all microphone settings and official scenarios

desh2608:
This PR contains AMI data preparation recipe. We provide options to prepare data according to any of the [3 official partitions](http://groups.inf.ed.ac.uk/ami/corpus/datasets.shtml) (`scenario-only`, `full-corpus`, and `full-corpus-asr`) and for any of the microphone settings (`ihm`, `ihm-mix`, `sdm`, and `mdm`). 

### Example usage

```python
from pathlib import Path
from lhotse.recipes.ami import download_ami, prepare_ami

root_dir = Path('data')
output_dir = root_dir / 'ami_nb'

download_ami(root_dir, mic='sdm')
ami_manifests = prepare_ami(root_dir, output_dir, mic='sdm', partition='full-corpus-asr')
```

### Options for `download_ami()`

  * `target_dir`: Pathlike, the path to store the data.
  * `force_download`: bool (default = False), if True, download even if file is present.
  * `url`: str (default = 'http://groups.inf.ed.ac.uk/ami'), AMI download URL.
  * `mic`: str {`ihm`,`ihm-mix`,`sdm`,`mdm`}, type of mic setting.

### Options for `prepare_ami()`

  * `data_dir`: Pathlike, the path of the data dir.
  * `output_dir`: Pathlike, the path where to write the manifests.
  * `mic`: str {`ihm`,`ihm-mix`,`sdm`,`mdm`}, type of mic to use.
  * `partition`: str {`full-corpus`,`full-corpus-asr`,`scenario-only`}, AMI official data split
  * `max_pause`: float (default = 0.0), max pause allowed between word segments to combine segments

_Returns_: a Dict whose key is ('train', 'dev', 'eval'), and the values are dicts of manifests under keys 
        'recordings' and 'supervisions'.

The `partition` and `max_pause` must be chosen depending on the task. For example:

  * Speaker diarization: set `partition="full-corpus"` and `max_pause=0`
  * ASR: set `partition="full-corpus-asr"` and `max_pause=0.3` (or some value in the range 0.2-0.5)


lhotse/cut.py
```diff
@@ -1132,6 +1132,26 @@ def describe(self) -> None:
         with pd.option_context('precision', 1):
             print(durations.describe().drop('count'))
 
+    def add_supervision(
```
pzelasko, line 1135: I'm trying to figure out what is the motivation for this method. Am I correct in guessing that it allows you to replace the original supervisions with different ones from another source (e.g. ground truth -> forced alignment or predictions) and get a new CutSet? I'm not sure if this is the right way to do this, since these supervisions could actually exceed the bounds of the cut. I think I'd rather create a new CutSet using the updated SupervisionSet and the original RecordingSet. Does that make sense for your use-case?

lhotse/recipes/ami.py
```diff
@@ -13,31 +13,31 @@
...
+import soundfile as sf
 
 from lhotse.audio import AudioSource, Recording, RecordingSet
 from lhotse.supervision import SupervisionSegment, SupervisionSet
 from lhotse.utils import Pathlike, Seconds
 
-# TODO: support the "sdm" and "mdm" microphone settings
-mic = 'ihm'
+# TODO: support the "mdm" microphone setting. For "sdm" we currently use Array1-01.
+mics = ['ihm','ihm-mix','sdm']
```
pzelasko, line 40: Could you show an example of 1-2 recordings from the `RecordingSet` manifest for the new setups? I remember that some time ago I changed the recipe to have all channels from a single session grouped as `AudioSource`s in a single `Recording` object, and I'm curious if the new setups also have that property. 

It might actually be nice to add such an example in the docstring of `prepare_ami` for illustration.

pzelasko: (you can call `recording_set.to_yaml(path)` and paste one Recording - YAML is much easier to read than JSON)

pzelasko (commented):
It looks pretty nice! 

I just noticed that we forgot to add a CLI mode for downloading/preparing AMI - if that'd would be useful to you, could you add it? It's probably not a lot of work (see `lhotse/bin/modes` directory for examples).

lhotse/cut.py
desh2608: Okay yeah that makes sense. I was using `trim_to_supervisions()` after adding the new supervisions which probably circumvented the bounds problem. But it's better to create a new CutSet altogether.

sw005320:
Why don't you include `mdm`?

lhotse/recipes/ami.py
desh2608: > Why don't you include `mdm`?

I'll add that soon.

desh2608:
> Why don't you include `mdm`?

@sw005320 I'll add that soon.

It appears that in the past, different groups have been using different reference RTTMs and train-test splits for AMI diarization evaluation. Paola and I are meeting with folks from BUT tomorrow to decide upon a standardized reference and data split. I can then make those a part of the Kaldi recipe and also the Lhotse recipe, and BUT will do the same in their code repository. This should make future evaluations easier.

desh2608:
> > Why don't you include `mdm`?
> 
> @sw005320 I'll add that soon.
> 
> It appears that in the past, different groups have been using different reference RTTMs and train-test splits for AMI diarization evaluation. Paola and I are meeting with folks from BUT tomorrow to decide upon a standardized reference and data split. I can then make those a part of the Kaldi recipe and also the Lhotse recipe, and BUT will do the same in their code repository. This should make future evaluations easier.

@pzelasko This would mean that the data preparation would be different for ASR and diarization (different splits, references). How should this be handled?

lhotse/cut.py
pzelasko: I meant cases where a different SupervisionSet implies re-segmentation, e.g. segment 5s - 10s changed into a segment of 5.2s to 10.5s - then it's mismatched with the cut (which would have matched the "old" segments boundaries when created with `trim_to_supervisions`.

pzelasko:
If you are able to agree on a standard split for diarization, why couldn't it become a standard split for ASR going forward? Especially if you re-train e.g. Kaldi SOTA and show the results for the same model on the new split. You could also add a note in Kaldi and here about this so that the interested parties can become aware.

In any case, we could also support an option to create different kinds of splits (with proper documentation).

desh2608:
@pzelasko I have updated the AMI setup. There are now separate "ASR" and "diarization" recipes which can be used as:

```
from lhotse.recipes.ami.asr import download, prepare_ami
```

Could you please take a look when you have some time?

pzelasko:
Thanks!

Do we need to support both types of splits? If you guys figured out that the new partitioning is better to evaluate diarization, can't we use it for ASR as well? Will also make it easier to evaluate the joint system performance. We don't have to support both old and new scheme: 1) it seems quite cumbersome; 2) Lhotse/K2 are forward-looking projects, it's ok to drop backward compatibility sometimes.

But if you think it's important to support both anyway, I'd rather have a single method for preparation with an extra argument that controls the behavior rather than two modules. I think it's easier to discover from a user's perspective when there's consistently a single entry point to these things across various corpora (see e.g. [LibriSpeech](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/librispeech.py#L33) for an example of choosing between mini and full variant). The asr/diarization naming for them is also a bit misleading in context of Lhotse's goal of decoupling datasets from specific tasks; it's only at the PyTorch dataset stage when it becomes task-specific. Maybe v1/v2 could be better?

lhotse/recipes/ami/diarization.py
```diff
@@ -0,0 +1,191 @@
...
+from typing import Dict, List, NamedTuple, Optional, Union
+
+from lhotse.qa import validate_recordings_and_supervisions
+from lhotse.audio import AudioSource, Recording, RecordingSet
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike, Seconds
+
+from lhotse.recipes.ami.common import download_audio, prepare_audio_other, remove_supervision_exceeding_audio_duration
+
+# TODO: support the "ihm" and "mdm" microphone settings
```
pzelasko, line 29: Do you intend to add these too? (it's ok to say no)

desh2608:
> Thanks!
> 
> Do we need to support both types of splits? If you guys figured out that the new partitioning is better to evaluate diarization, can't we use it for ASR as well? Will also make it easier to evaluate the joint system performance. We don't have to support both old and new scheme: 1) it seems quite cumbersome; 2) Lhotse/K2 are forward-looking projects, it's ok to drop backward compatibility sometimes.
> 
> But if you think it's important to support both anyway, I'd rather have a single method for preparation with an extra argument that controls the behavior rather than two modules. I think it's easier to discover from a user's perspective when there's consistently a single entry point to these things across various corpora (see e.g. [LibriSpeech](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/librispeech.py#L33) for an example of choosing between mini and full variant). The asr/diarization naming for them is also a bit misleading in context of Lhotse's goal of decoupling datasets from specific tasks; it's only at the PyTorch dataset stage when it becomes task-specific. Maybe v1/v2 could be better?

I think there is a middle ground here which we can aim for. We cannot use the same data partition for all tasks, since it is recommended to use different splits in the [official docs](http://groups.inf.ed.ac.uk/ami/corpus/datasets.shtml). However, I can have an argument to choose between `scenario-only`, `full-corpus`, and `full-corpus-asr` partitions. Question: Should I just declare these partitions as global constants?

I think it should be okay to use just one reference. Earlier I had 2 to be compatible with Kaldi, but the Kaldi recipe uses older annotations which are known to be erroneous, so it would actually be better to drop those.

pzelasko:
That sounds good. Global constants are perfectly fine.

lhotse/recipes/ami/diarization.py
desh2608: Yes I'm adding them now in the next commit.

lhotse/recipes/ami/diarization.py
```diff
@@ -0,0 +1,191 @@
...
+    if force_download or not rttm_path.is_dir():
+        rttm_path.mkdir(parents=True, exist_ok=True)
+        for split in dataset_parts:
+            for session in dataset_parts[split]:
+                cur_rttm_path = rttm_path / f'{session}.rttm'
+                rttm_url = base_url.format(split, session)
+                urllib.request.urlretrieve(rttm_url, filename=cur_rttm_path)
+
+
+class AmiSegmentAnnotation(NamedTuple):
```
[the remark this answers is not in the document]
desh2608: IIRC `SupervisionSegment` requires an id and a recording_id associated with the segment. I use this class to parse annotations into segments, and later create the supervision (depending on the mic setting).

lhotse/recipes/ami.py
```diff
@@ -84,15 +177,59 @@ def download(
...
+                for channel in MDM_CHANNELS:
+                    wav_name = f'{item}.{array}-{channel}.wav'
+                    wav_url = f'{url}/AMICorpusMirror/amicorpus/{item}/audio/{wav_name}'
+                    wav_dir = target_dir / 'wav_db' / item / 'audio'
+                    wav_dir.mkdir(parents=True, exist_ok=True)
+                    wav_path = wav_dir / wav_name
+                    if force_download or not wav_path.is_file():
+                        urlretrieve_progress(wav_url, filename=wav_path, desc=f'Downloading {wav_name}')
 
+def download(
```
pzelasko, line 208: Can we rename to `download_ami`? I'll be updating other recipes in a similar way.

lhotse/recipes/ami.py
```diff
@@ -101,179 +238,277 @@ class AmiSegmentAnnotation(NamedTuple):
...
+        max_pause: Optional[float] = 0.0
 ) -> Dict[str, Dict[str, Union[RecordingSet, SupervisionSet]]]:
     """
     Returns the manifests which consist of the Recordings and Supervisions
-
     :param data_dir: Pathlike, the path of the data dir.
     :param output_dir: Pathlike, the path where to write the manifests.
-    :return: a Dict whose key is ('train', 'dev', 'eval'), and the value is Dicts with keys 'audio' and 'supervisions'.
+    :param mic: str {'ihm','ihm-mix','sdm','mdm'}, type of mic to use.
+    :param partition: str {'full-corpus','full-corpus-asr','scenario-only'}, AMI official data split
```
pzelasko, line 458: Can you add the same comment you made in the PR about which partition to select for ASR and which for diarization?

pzelasko (commented):
Amazing work! It really looks like the one data prep recipe to rule them all 🔥 I left two minor comments, otherwise LGTM!

PR #158, opened 2020-12-05
Python 3.6 support

pzelasko:


lhotse/utils.py
```diff
@@ -316,3 +316,26 @@ def urlretrieve_progress(url, filename=None, data=None, desc=None):
     with tqdm(unit='B', unit_scale=True, unit_divisor=1024, miniters=1, desc=desc) as t:
         reporthook = tqdm_urlretrieve_hook(t)
         return urlretrieve(url=url, filename=filename, reporthook=reporthook, data=data)
+
+
+class nullcontext(AbstractContextManager):
```
csukuangfj, line 321: Should the name of a class be `CamelCase`?

pzelasko: It's a std library class introduced only in Python 3.7; I am simply after the compatibility when using 3.6 :)

Issue #151, opened 2020-11-24
Feature: CMVN


desh2608:
It would be useful to have CMVN for feature transformation, including offline and online CMVN (per speaker or per utterance) as in Kaldi, and a simple sliding window CMVN (which can be just a wrapper around https://pytorch.org/audio/stable/transforms.html#slidingwindowcmn).

pzelasko:
@desh2608 can you take a look at PR #159 ?

desh2608:
@pzelasko LGTM.

PR #159, opened 2020-12-05
Some feature normalization support and docs

pzelasko:
Addresses #151

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -1,3 +1,4 @@
+import logging
 import random
 import warnings
 from concurrent.futures import ProcessPoolExecutor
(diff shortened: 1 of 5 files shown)


pzelasko:
@danpovey could you review? I was also considering adding an argument(s) like “normalize” to load_features() that could allow performing global/instance normalization, but I’m not sure about that.

lhotse/cut.py
```diff
@@ -1478,6 +1480,30 @@ def compute_and_store_features(
         ))
         return cut_set
 
+    def compute_global_feature_stats(self, storage_path: Optional[Pathlike] = None) -> Dict[str, np.ndarray]:
+        """
+        Compute the global means and standard deviations for each feature bin in the manifest.
```
danpovey, line 1485: It might make sense to add a max_cuts argument, e.g. max_cuts = 5000, and have it randomly sample that many cuts if there are more than that available.  Otherwise could be wastefully slow.

PR #167, opened 2020-12-18
Lazy time-domain data augmentation on Recording / Cut

### LINKED ISSUE #166: Speed perturbation (sp < 1) cuts out speech
Since we're using "sox speed" and then "sox resample" with end up with a signal of a different duration than the original one. So far we've adopted WavAugment's default behavior of padding it with zeros (sp > 1) or truncating (sp < 1), but the latter can hurt ASR training. Example:

image

I think the right way to go forward could be to make "perturb_speed" a method on `Recording`, and possibly `Cut` and `CutSet`. It will have to adjust the duration/num_samples attributes, and when applied on `Cut` it should also adjust the supervisions' duration.  The transform should be applied lazily during audio loading to avoid excessive storage. We'd probably have to store the transform information in the Recording object itself. If we choose to do it, this opens the door to eventually performing all time-domain data augmentation in the same manner (e.g. reverb can also make the signal longer).

We should also validate that the Cut does not contain any pre-computed features (or warn that we are "detaching" them from Cut). I guess it's okay to apply this transform to each track independently in MixedCuts. If a user calls this method twice, it would probably make sense to apply it twice (or raise an exception that don't support it which is easier at this point). 

@danpovey WDYT?

pzelasko:
Resolves #166

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1,15 +1,16 @@
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 
 import numpy as np
 import warnings
 from io import BytesIO
-from math import sqrt
+from math import floor, sqrt
 from pathlib import Path
 from subprocess import PIPE, run
 from typing import Callable, Dict, Iterable, List, Optional, Sequence, Tuple, Union
 
 from lhotse.utils import (Decibels, JsonMixin, Pathlike, Seconds, SetContainingAnything, YamlMixin, fastcopy,
                           split_sequence)
+from lhotse.augmentation import AudioTransform, Speed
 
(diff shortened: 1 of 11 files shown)


danpovey (commented):
LGTM!

pzelasko:
I still have to add it to `Cut` to update the supervisions; will try to add it today.

pzelasko:
Closer to ready; need to test it a bit more and add to `CutSet` and other cut classes. There's quite a few edge cases in estimating the number of samples and start/duration correctly without actually running the speed perturbation (high potential for off-by-one errors).

pzelasko:
It seems to work okay - I'll merge and we can address any issues later if they arise.

PR #172, opened 2020-12-30
WIP: Proposal for "universal" PyTorch speech dataset

pzelasko:
# Request for comments

I've been thinking how to leverage the batching efficiency improvements we made in K2 IterableDataset in other dataset classes, and if it's possible to avoid the major code overlap between the most Dataset classes in Lhotse. I came up with some ideas how to create a highly-customizable dataset class that you only need to implement the relevant bits for to adapt it to a new task.

The core idea is that there's one class that does the auto-batching stuff (`SpeechDataset`), and we give a list of "fields" that we want to see in the batches. There's "signal fields" (e.g. audio/features, single/multi channel) and "supervision fields" (e.g. text/chars/phones/fsas/supervision start frames and number of frames/speaker id/etc.). See the implementations of `SpeechDataset` in `lhotse/dataset/core.py` and the fields in `lhotse/dataset/fields.py`. To illustrate the benefits, I changed the K2, VAD and TTS dataset implementations to use this class.

As a teaser this is how a TTS dataset is implemented:

```python
class SpeechSynthesisDataset(SpeechDataset):
    def __init__(self, cuts: CutSet, *args, **kwargs):
        super().__init__(
            cuts,
            *args,
            signal_fields=[fields.Audio(), fields.Feats()],
            supervision_fields=[fields.CharacterIds(cuts)],
            **kwargs
        )
```

Note that this becomes a very declarative way to define a task (minus some args/kwargs noise; I'll try to figure out if that can be improved upon).

Things that still need to be worked out in this approach:
- handling tasks when we might need more than a single cut set in the input (e.g. for speech enhancement, speech separation, speech translation, voice conversion, and some implementations of self-supervised learning)
- more flexible auto-batching criteria (currently we support max_frames and max_cuts, but we might want max_samples or something we will come up with later, etc.)

As a proof of concept I'm going to modify the K2 dataset to output FsaVec alongside the text (so that the graph construction is performed in dataloader processes rather than the main training processes; BTW the graphs can still be cached like we do it in snowfall).

I think this might be the right way to go but I'm not 100% sure. One alternative is just writing each Dataset separately, like we did so far. Some very specific tasks might actually still require that, but I think we can be flexible enough to cover a lot of use cases with this proposal. 

@danpovey WDYT

### DIFF
--- a/lhotse/dataset/fields.py
+++ b/lhotse/dataset/fields.py
@@ -7,7 +7,7 @@ from typing_extensions import Protocol, runtime_checkable
 
 from lhotse import CutSet, SupervisionSegment
 from lhotse.cut import AnyCut, MixedCut
-from lhotse.utils import Seconds, compute_num_frames, compute_num_samples
+from lhotse.utils import compute_num_frames, compute_num_samples
 
 
 class SignalField(Protocol):
@@ -184,22 +184,3 @@ def _collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
     for idx, cut in enumerate(cuts):
         audio[idx] = torch.from_numpy(cut.load_audio())
     return audio
-
-
-def _asserted_num_frames(total_num_frames: int, start: Seconds, duration: Seconds, frame_shift: Seconds) -> int:
-    """
-    This closure with compute the num_frames, correct off-by-one errors in edge cases,
-    and assert that the supervision does not exceed the feature matrix temporal dimension.
-    """
-    offset = compute_num_frames(start, frame_shift=frame_shift)
-    num_frames = compute_num_frames(duration, frame_shift=frame_shift)
-    diff = total_num_frames - (offset + num_frames)
-    # Note: we tolerate off-by-ones because some mixed cuts could have one frame more
-    # than their duration suggests (we will try to change this eventually).
-    if diff == -1:
-        num_frames -= 1
-    assert offset + num_frames <= total_num_frames, \
-        f"Unexpected num_frames ({offset + num_frames}) exceeding features time dimension for a supervision " \
-        f"({total_num_frames}) when constructing a batch; please report this in Lhotse's GitHub issues, " \
-        "ideally providing the Cut data that triggered this."
-    return num_frames
--- a/lhotse/dataset/fields.py
+++ b/lhotse/dataset/fields.py
@@ -1,4 +1,11 @@
-from decimal import ROUND_FLOOR, ROUND_HALF_DOWN
+"""
+The *fields* defined in this file conform to "Protocols" defined at the top.
+"Protocols" are like concepts in C++20, and you can think of them as an attempt
+to specify an expected interface without forcing the user to inherit from a common base class.
+It should make creating custom fields simpler for the user
+(they can in fact be implemented even as a simple method).
+"""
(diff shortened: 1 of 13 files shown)


test/dataset/test_speech_dataset_randomized.py
```diff
@@ -0,0 +1,72 @@
...
+            return_cuts=True,
+            # Enabling cut concatenation (efficient batching mechanism) increases the "risk" of
+            # off-by-one errors as num_frames may be wrongly computed with randomized offsets.
+            concat_cuts=True,
+            concat_cuts_duration_factor=3.0
+        )
+        ### End of test data preparation ###
+        # Test the invariants
+        for batch in dataset:
+            spvns = batch['supervisions']
```
danpovey, line 65: maybe `sups` would be easier to parse than `supervisions`

lhotse/dataset/fields.py
```diff
@@ -0,0 +1,338 @@
...
+    Output:
+
+    .. code-block:
+
+        output = {'char_ids': List[int], 'char_ids_len': int}
+    """
+
+    def __init__(self, cuts):
+        self.pad = '<PAD>'
+        self.id2sym = sorted({char for cut in cuts for s in cut.supervisions for char in s.text}) + [self.pad]
```
danpovey, line 234: Mm, it seems to me a little problematic to set the ids2sym and sym2id in this way (this is essentially a symbol table, I think; and in k2 we'd probably be wanting to load it from disk...)
This makes it a function of the dataset, but sometimes there are rare characters that might not appear in a given collection of text, or characters that should be considered as OOV.

lhotse/dataset/fields.py
```diff
@@ -0,0 +1,338 @@
...
+            output['audio_is_voiced'] = cut.supervisions_audio_mask()
+        if self.use_features:
+            output['frame_is_voiced'] = cut.supervisions_feature_mask()
+        return output
+
+
+def _collate_features(cuts: CutSet) -> torch.Tensor:
+    assert all(cut.has_features for cut in cuts)
+    first_cut = next(iter(cuts))
+    features = torch.empty(len(cuts), first_cut.num_frames, first_cut.num_features)
```
danpovey, line 303: Perhaps a comment could clarify this?  Are we assuming the cuts all have the same length, or that the 1st one is longest [if the latter, the use of `torch.empty()` is problematic]

danpovey (commented):
Mm.  Ultimately it's up to you, but I am a little concerned- and this is a very general concern, not specifically about this PR- that this code may be difficult to maintain for anyone but you.  You are using extremely advanced Python, and generally speaking any given invocation will require a lot of other code.  My instinct would be to try to design it so that as far as possible objects that hold data are quite simple, and most of the logic goes into functions that have clear (and fairly simple) inputs and outputs.

Of course there are other design goals too.  The way you've done it may actually be the best way, I'm just not sure.

pzelasko:
Alright, I hear you. Let me give it some more thought then.

pzelasko:
I moved some fix/test changes to a separate PR (#173) and rebased this one on it.

pzelasko:
Closing, as we're not gonna go this way; #194 performs a refactoring that separates many of these things into clearly defined modules, and those can be used to put together probably any dataset one can imagine with not too much code.

Issue #63, opened 2020-08-02
Add MUSAN data preparation


pzelasko:
We'll need to:

1. Download MUSAN
2. Create separate recording manifests for the subsets: `noise`, `music`, `babble`. I don't think we need a supervision manifest.

It should be done similarly to how `mini_librispeech` and `librimix` are being prepared.

PR #174, opened 2021-01-03
MUSAN data preparation + some fixes

pzelasko:
Some fixes:
- LibriSpeech CLI fix/update
- Store Recordings without null fields


pzelasko:
Resolves #63

PR #175, opened 2021-01-03
Validation methods for various Lhotse manifests

### LINKED ISSUE #13: Validate manifest

pzelasko:
It's conceptually similar to Kaldi's `validate_data_dir.sh`, but probably performs different checks.

I haven't tested it yet, just pushing the code written so far.

Resolves #13

### DIFF
--- a/lhotse/__init__.py
+++ b/lhotse/__init__.py
@@ -5,5 +5,6 @@ from .features import *
 from .kaldi import load_kaldi_data_dir
 from .manipulation import combine, load_manifest, to_manifest
 from .supervision import SupervisionSegment, SupervisionSet
+from .qa import validate
 
 from . import recipes
--- a/lhotse/__init__.py
+++ b/lhotse/__init__.py
@@ -5,6 +5,6 @@ from .features import *
 from .kaldi import load_kaldi_data_dir
 from .manipulation import combine, load_manifest, to_manifest
 from .supervision import SupervisionSegment, SupervisionSet
-from .qa import validate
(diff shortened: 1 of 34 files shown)


danpovey:
Cool..
How can we ensure that this actually gets used?
In Kaldi you follow the recipes and they call the validation, not sure if there is an exact analogue here unless we call it whenever we write something to disk, or read from disk (by default), or something like that.

pzelasko:
We can add them to all prepare_corpus functions, and use them in snowfall recipes too, which is probably the closest analogue to Kaldi. Do you think that’s enough?

danpovey:
sure, that sounds OK

pzelasko:
I extended the validation to check recording <-> supervision consistency, and added that to each corpus prep method. 

I suppose it could break some of the corpus data prep scripts if they did not have rigorous checks before. I'll check some of them (so far LibriSpeech is OK) and then merge. If anybody in the community wants to check a corpus they are using with these changes, let me know (I'll probably merge this in 1-2 days).

pzelasko:
BTW after merging this PR, likely all CutSet's will have to be re-created (I found and fixed a bug in supervision <-> cut channel consistency and supervision <-> recording ID consistency after speed perturbation).

pzelasko:
I'm merging, if there are any issues with the recipes we can resolve them later.

PR #176, opened 2021-01-04
Parallel feature extraction with HDF5 storage

pzelasko:
The `CutSet.compute_and_store_features` method was not able to store features in HDF5 files when using an executor because of parallel writes into a single file. I've made that method private (prefixed it with a dash) and added two new methods: `CutSet.compute_and_store_features_in_files` and `CutSet.compute_and_store_features_in_hdf5`. 

The former method does exactly the same thing as e.g. in the snowfall recipe. The latter will split the CutSet into multiple parts and process each part in a separate job, writing it to a separate HDF5 file. 

This introduces a breaking API change but I think it's useful. Anyway, `cut.py` is definitely becoming too cluttered, and I am planning a refactoring of it sometime in the near-ish future.


danpovey:
There isn't a way to somehow have a more uniform interface, possibly at the expense of some nonoptimality?  
I feel like we are making the user see more complexity than he/she wants.

pzelasko:
Thanks, that's a great point! Here's an alternative API - it retains the old `compute_and_store_features` method, just changes the args a bit. Proposed signature + docs:

```python
def compute_and_store_features(
        self,
        extractor: FeatureExtractor,
        output_dir: Pathlike,
        num_jobs: Optional[int] = None,
        augment_fn: Optional[AugmentFn] = None,
        storage_type: Type[FW] = LilcomFilesWriter,
        executor: Optional[Executor] = None,
        mix_eagerly: bool = True
) -> 'CutSet':
  """
  Extract features for all cuts, possibly in parallel,
  and store them using the specified storage object.
  
  Examples:
  
      Extract fbank features on one machine using 8 processes,
      store each array in a separate file with lilcom compression:
  
      >>> cuts = CutSet(...)
      ... cuts.compute_and_store_features(
      ...     extractor=Fbank(),
      ...     output_dir='feats',
      ...     num_jobs=8
      ... )
  
      Extract fbank features on one machine using 8 processes,
      store arrays partitioned in 8 HDF5 files with lilcom compression:
  
      >>> cuts = CutSet(...)
      ... cuts.compute_and_store_features(
      ...     extractor=Fbank(),
      ...     output_dir='feats',
      ...     num_jobs=8,
      ...     storage=LilcomHdf5Writer
      ... )
  
      Extract fbank features on multiple machines using a Dask cluster
      with 80 jobs,
      store arrays partitioned in 80 HDF5 files with lilcom compression:
  
      >>> from distributed import Client
      ... cuts = CutSet(...)
      ... with Client(...) as ex:
      ...     cuts.compute_and_store_features(
      ...         extractor=Fbank(),
      ...         output_dir='feats',
      ...         num_jobs=80,
      ...         storage=LilcomHdf5Writer,
      ...         executor=ex
      ...     )
  
  :param extractor: A ``FeatureExtractor`` instance (either Lhotse's built-in or a custom implementation).
  :param output_dir: The path to directory where we will store the features.
      The exact type and layout of stored files will be dictated by the ``storage_type`` argument.
  :param num_jobs: The number of parallel processes used to extract the features.
      We will internally split the CutSet into this many chunks and process each chunk in parallel.
  :param augment_fn: an optional callable used for audio augmentation.
      Be careful with the types of augmentations used: if they modify the start/end/duration
      times of the cut and its supervisions, you will end up with incorrect supervision
      information when using this API.
      E.g. for speed perturbation, use ``CutSet.perturb_speed()`` instead.
  :param storage_type: a ``FeaturesWriter`` subclass type.
      It determines how the featurs are stored to disk,
      e.g. separate file per array, HDF5 files with multiple arrays, etc.
  :param executor: when provided, will be used to parallelize the feature extraction process.
      By default, we will instantiate a ProcessPoolExecutor.
      Learn more about the ``Executor`` API at https://lhotse.readthedocs.io/en/latest/parallelism.html
  :param mix_eagerly: Related to how the features are extracted for ``MixedCut`` instances, if any are present.
      When False, extract and store the features for each track separately,
      and mix them dynamically when loading the features.
      When True, mix the audio first and store the mixed features, returning a new ``Cut`` instance
      with the same ID. The returned ``Cut`` will not have a ``Recording`` attached.
  :return: Returns a new ``CutSet`` with ``Features`` manifests attached to the cuts.
  """
```

danpovey:
Cool!

danpovey (commented):
Looks better than the previous approach...

pzelasko:
We can merge once the snowfall changes are reviewed and good to go.

@freewym you'll need to update your recipe in Espresso (they're small changes)

freewym:
I see. Thanks!

danpovey (approved):
LGTM!

PR #180, opened 2021-01-12
Musan on-the-fly data augmentation for ASR

pzelasko:
I'm making a corresponding PR in Snowfall - the training is running, I'll write an update about %WER results tomorrow.

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -1601,7 +1601,9 @@ class CutSet(JsonMixin, YamlMixin, Sequence[AnyCut]):
                     to_mix = cuts.sample()
                     # Keep the SNR constant for each cut from "self".
                     mixed = mixed.mix(other=to_mix, snr=snr, offset_other_by=mixed_in_duration)
-                    mixed_in_duration += to_mix.duration
+                    # Since we're adding floats, we can be off by an epsilon and trigger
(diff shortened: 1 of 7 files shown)


lhotse/cut.py
```diff
@@ -1531,9 +1535,84 @@ def cut_into_windows(self, duration: Seconds, keep_excessive_supervisions: bool
...
+                # Keep sampling until we mixed in a "duration" amount of noise.
+                while mixed.duration < duration:
+                    to_mix = cuts.sample()
+                    # Keep the SNR constant for each cut from "self".
+                    mixed = mixed.mix(other=to_mix, snr=snr, offset_other_by=mixed_in_duration)
+                    # Since we're adding floats, we can be off by an epsilon and trigger
+                    # some assertions for exceeding duration; do precautionary rounding here.
+                    mixed_in_duration = round(mixed_in_duration + to_mix.duration, ndigits=8)
+            # We truncate the mixed to either the original duration or the requested duration.
+            mixed = mixed.truncate(duration=cut.duration if duration is not None else duration)
```
freewym, line 1612: should be 
```python
if duration is None
```

pzelasko: Thanks, I'll add tests before merging ;)

PR #190, opened 2021-01-30
Frame shift as a field in Features + solved issues for more combinations of sampling_rate and frame_shift

pzelasko:
We originally assumed that `frame_shift` can be inferred from `duration/num_frames`, but it's not true for "exotic" frame shifts, e.g. when using a sampling rate of 22050, with frame_length of 1024 samples (librosa's "window_len") and frame_shift of 256 samples (librosa's "window_hop"), the frame_shift in seconds is a whopping `0.011609977324263039`s... we'll never get it right this way because we'll either run into float precision problems that we can't fix (e.g. frame_shift of 0.0099999999998) or we'll round too much and get a wrong estimate of the number of frames (e.g. 0.0116099773... -> 0.012).

I feel like constantly working around the fact that the time is measured in seconds (floats). I am really considering making a big change in Lhotse to measure the time in samples instead, and compute durations from that + sampling rate... There is still going to be an inconsistency between samples and frames (when a frame discarded leftover samples or used padding), but maybe that can be amended by updating the number of samples+duration and discarding/padding when calling `load_audio()`. I think I'd rather manipulate a few samples than a feature frame to achieve consistency between the manifests and the data. @danpovey does that make sense to you?

For now, I'm adding `frame_shift` as its own data field.


danpovey:
Will storing things in samples still make sense if we do things like extracting features and then, maybe, combining features from streams that had different sampling rates?  I'm concerned that after we have features, it may not make sense to view durations in terms of num_samples.

pzelasko:
I think the only other alternative that avoids consistency issues is to have the core manipulation methods accept three different time representations, e.g. `pad(num_samples=None, duration=None, num_frames=None)` (you specify only one and it tries to do the right thing with the others). It feels to me like maybe a too brute-forcey solution though?

pzelasko:
I'll try to summarize all the issues that I've encountered with this and maybe then a clear solution will emerge.

danpovey:
mm OK.

pzelasko:
I think that these issues boil down to two points described below. I think I resolved the second point with my recent commits, but I still need to fix some tests after that.

**Estimating number of frames for mixed cuts**

When we append two cuts that have duration like 10.005, then each of them has an extra frame at the end. The summed duration is 20.01, but the number of frames is 2002 instead of 2001 (assuming frame_shift 0.01). In some rare cases the reverse seems to also happen (I debugged it but I don't remember what was the exact reason anymore). We have a hacky workaround (removing/duplicating last frame) that seems to work for now, but it would be good to deal with that in a more systematic way.

**Frame shifts rounding**

Practically all feature extractors compute the number of frames using numbers samples - for the input and window shift at the least. Our "compute_num_features" worked so far because we only tested it for the special case of frame_shift 0.01 and sampling_rate 8k or 16k, but it does not always work with other settings (e.g. sampling_rate 22.05k, frame_shift 0.012).

Torchaudio "Kaldi" extractors seem to compute `num_frames` as `(input_num_samples + frame_shift_samples // 2) // frame_shift_samples`, so I updated our method to do the same. However, it seems that defining frame_shift in milliseconds can be a trap; e.g. with sampling rate 22050, frame_shift of 0.01s is physically impossible as it results in a fractional shift in terms of samples (220.5).

danpovey:
Mm.  I'm concerned that if in future we end up dealing with features obtained with different sampling rates, e.g. audio and video, doing everything in terms of samples might itself be a trap.  At least seconds is a shared/universal concept...
In HTK they have the concept of a 'tick', I forget how they define it.  Using some kind of common multiple of the sampling rates might be an option.
Just concerned that making everything multiples of samples might be a different kind of trap.

pzelasko:
I like the video example, this is definitely something we need to be able to support.

I was thinking about introducing a class (e.g. called `Duration`) that would "encapsulate" all this stuff, it would conceptually look sth. like:

```python
@dataclass
class Duration:
    sampling_rate: int
    num_samples: Optional[SampleCount] = None
    num_frames: Optional[FrameCount] = None
    frame_shift: Optional[SampleCount] = None

    @property
    def seconds(self) -> float: ...
```

You'd either initialize it with `num_samples` or `num_frames` and `frame_shift` (and always pass the sampling_rate). If you use samples, it will use that and the sampling rate to compute the seconds. If you use frames, it will use them instead both for computing the seconds and the number of samples. It could have operators like `<`, `==`, and others, that would use the "seconds" representation (it is up to the user to determine whether it makes sense to compare `Duration` with different sampling_rates etc.). Then, when we extract the features for a cut, we would express its duration in seconds and num_samples as a function of num_frames and frame_shift.

Since there would be a lot of these objects around, we could try to optimize the storage to encode these quantities in a single integer with bit shifting (only if necessary). Otherwise, I think it could work well (although it's hard to foresee all the possible problems with any of these approaches...).

danpovey:
There are situations where we'd know the duration in seconds but not in samples: for example if it comes from some kind of supervision information.  So I think it would make sense to handle that explicitly rather than trying to hack it somehow.

I think if we use tools that do some kind of rounding for the frame-shift, it would be best if we invoke them in small enough chunks that the rounding doesn't matter, and shift the chunks such that on longer time-scales it works out exactly proportional.

In a sense I think the duration itself is just a number (e.g. a rational number or a float), and the frame-shift and num-samples are somewhat separate issues.  Which is not to say they couldn't somehow be encapsulated in the same class.  But there might be objects that have a duration but not a frame-shift or num-samples.

pzelasko:
> There are situations where we'd know the duration in seconds but not in samples: for example if it comes from some kind of supervision information. So I think it would make sense to handle that explicitly rather than trying to hack it somehow.

Good point.

> I think if we use tools that do some kind of rounding for the frame-shift, it would be best if we invoke them in small enough chunks that the rounding doesn't matter, and shift the chunks such that on longer time-scales it works out exactly proportional.

I didn't fully get that, could you elaborate?

I think all the issues are "contained" for now, I'll not merge yet in case you'd like to review.

danpovey:
There are situations where we'd know the duration in seconds but not in
> samples: for example if it comes from some kind of supervision information.
> So I think it would make sense to handle that explicitly rather than trying
> to hack it somehow.
>
> Good point.
>
> I think if we use tools that do some kind of rounding for the frame-shift,
> it would be best if we invoke them in small enough chunks that the rounding
> doesn't matter, and shift the chunks such that on longer time-scales it
> works out exactly proportional.
>
> I didn't fully get that, could you elaborate?
>

I mean, if the feature extraction code is causing the frame-shift to differ
from the nominal frame-shift, I suggest to feed in chunks of, say, 100
frames to the feature extraction, with the chunks shifted by the nominal
frame-shift, so that things don't get out of sync.  Having to deal with a
situation where the nominal and real frame-shift differ would be a terrible
headache.

Fine to merge this.. was talking about the broader/future issue.

pzelasko:
I think I understand your point now. I believe I "disabled" the ability to use a frame_shift that would fall out of sync with the nominal frame_shift in this PR; e.g.: if you specify a frame_shift of 10ms with sampling_rate of 22050 (window_hop of 220.5 samples), the feature extraction code will raise an exception that the frame_shift is "physically impossible".

PR #194, opened 2021-02-06
Refactoring - modularizing K2 ASR dataset

pzelasko:
This is mostly a refactoring PR, that splits the K2SpeechRecognitionIterableDataset into multiple modules with separation of concerns. It's an inverse approach to the "universal dataset" in #172 - instead of one big blob with a lot of hooks, I'm just going to provide multiple components for building "the perfect dataset".

The dataset itself now only takes care of converting a batch CutSet into an actual batch. There is a `CutSampler` abstraction that takes care of multi/distributed processing. Other classes can inherit to specify the filtering behaviour (e.g. what we had so far is now called `SingleCutSampler`, as an example that I'm using in another project I introduced `CutPairSampler` as well). Maybe there is an even cleaner way (no inheritance) to separate the "partitioning" from the "filtering" but I don't have an idea atm.

Finally, the cut concatenation and noise mixing are separate transforms performed on batch CutSets, optionally provided to the dataset's constructor (nobody else is going to be surprised that batch_size of features and supervisions is not the same thing...).

I yet have to test whether the distributed behaviour is correct (needed for multi-GPU training), and create some examples (maybe a notebook) that illustrate this stuff. Preliminary examples of usage can be seen in the tests.

### DIFF
--- a/lhotse/dataset/sampling.py
+++ b/lhotse/dataset/sampling.py
@@ -3,7 +3,7 @@ import math
 import random
 import warnings
 from abc import ABC, abstractmethod
-from typing import Iterable, List, Optional, Tuple
+from typing import Any, Iterable, List, Optional, Tuple
 
 import torch
 
@@ -65,7 +65,8 @@ class CutSampler(ABC):
         return self
 
     @abstractmethod
-    def __next__(self): ...
+    def __next__(self) -> Any:
+        pass
 
 
 class SingleCutSampler(CutSampler):
@@ -147,6 +148,7 @@ class SingleCutSampler(CutSampler):
 class CutPairsSampler(CutSampler):
     """
     Samples pairs of cuts from a "source" and "target" CutSet.
+    It expects that both CutSet's strictly consist of Cuts with corresponding IDs.
     It will try to satisfy the criteria of max_source_frames, max_target_frames, and max_cuts.
     Use as an iterable that yields pairs of ``CutSet`` objects (individual batches).
     """
@@ -163,7 +165,8 @@ class CutPairsSampler(CutSampler):
         """
         CutPairsSampler's constructor.
 
-        :param cuts: the ``CutSet`` to sample data from.
+        :param source_cuts: the first ``CutSet`` to sample data from.
+        :param target_cuts: the second ``CutSet`` to sample data from.
         :param max_frames: The maximum number of feature frames that we're going to put in a single batch.
             The padding frames do not contribute to that limit, since we pack the batch by default to minimze
             the amount of padding.
--- a/lhotse/dataset/sampling.py
+++ b/lhotse/dataset/sampling.py
@@ -1,20 +1,20 @@
 import logging
-import math
 import random
 import warnings
-from abc import ABC, abstractmethod
-from typing import Any, Iterable, List, Optional, Tuple
+from math import ceil
+from typing import Iterable, List, Optional
 
-import torch
+import torch.distributed as dist
+from torch.utils.data import Sampler
 
 from lhotse import CutSet
 
 
-class CutSampler(ABC):
+class CutSampler(Sampler[List[str]]):
     """
     CutSampler is responsible for collecting batches of cuts, given specified criteria.
-    It implements correct handling of multiprocessing in DataLoader,
-    as well as distributed training, so that the cuts are not duplicated across workers.
+    It implements correct handling of distributed sampling in DataLoader,
+    so that the cuts are not duplicated across workers.
 
     Sampling in a CutSampler is intended to be very quick - it only uses the metadata in
     ``CutSet`` manifest to select the cuts, and is not intended to perform any I/O.
@@ -25,13 +25,11 @@ class CutSampler(ABC):
 
     Example usage::
 
-        >>> dataset = K2SpeechRecognitionIterableDataset(
-        ...     cuts,
-        ...     sampler=SingleCutSampler(cuts, world_size=4, local_rank=0)
-        ... )
-        >>> loader = DataLoader(dataset, batch_size=None)
+        >>> dataset = K2SpeechRecognitionDataset(cuts)
+        >>> sampler = SingleCutSampler(cuts, shuffle=True)
+        >>> loader = DataLoader(dataset, sampler=sampler, batch_size=None)
         >>> for epoch in range(start_epoch, n_epochs):
-        ...     dataset.sampler.set_epoch(epoch)
+        ...     sampler.set_epoch(epoch)
         ...     train(loader)
 
     .. note::
@@ -39,7 +37,7 @@ class CutSampler(ABC):
         For implementers of new samplers:
         Subclasses of CutSampler are expected to implement ``__next__()`` to introduce specific
         sampling logic (e.g. based on filters such as max number of frames/tokens/etc.).
-        CutSampler defines ``__iter__()`` which optionally shuffles the cut IDs, and resets
+        CutSampler defines ``__iter__()``, which optionally shuffles the cut IDs, and resets
         ``self.current_idx`` to zero (to be used and incremented inside of ``__next__()``.
     """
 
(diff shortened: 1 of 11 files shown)


lhotse/dataset/sampling.py
```diff
@@ -0,0 +1,288 @@
...
+            target_cuts: CutSet,
+            max_source_frames: int = 26000,
+            max_target_frames: int = 26000,
+            max_cuts: Optional[int] = None,
+            **kwargs
+    ):
+        """
+        CutPairsSampler's constructor.
+
+        :param cuts: the ``CutSet`` to sample data from.
```
danpovey, line 166: this seems wrong

lhotse/dataset/transforms/cat.py
```diff
@@ -0,0 +1,79 @@
+from typing import Optional, Sequence
+
```
danpovey, line 2: I think calling this concatenate rather than cat would be clearer (filename and class name), but up to you.

lhotse/dataset/transforms/cat.py
```diff
@@ -0,0 +1,79 @@
...
+from lhotse import CutSet
+from lhotse.cut import AnyCut
+from lhotse.utils import Seconds
+
+
+class CutCat:
+    """
+    A transform on batch of cuts (``CutSet``) that concatenates the cuts to minimize the total amount of padding;
+    e.g. instead of creating a batch with 40 examples, we will merge some of the examples together
+    adding some silence between them to avoid a large number of padding frames that waste the computation.
```
danpovey, line 12: Bear in mind that this could be a little dangerous to use especially if there is batchnorm, as it messes with statistical independence (especially if the data comes from multiple sources with different length distributions).
You might want to warn about that.

lhotse/dataset/transforms/mix.py
```diff
@@ -0,0 +1,40 @@
+from typing import Optional, Tuple, Union
+
+from lhotse import CutSet
+from lhotse.utils import Decibels
+
+
+class CutMix:
+    """
+    A transform for batches of cuts (CutSet's) that stochastically performs n
+    oise augmentation with a constant or varying SNR.
```
danpovey, line 10: line break in middle of word

lhotse/dataset/transforms/mix.py
```diff
@@ -0,0 +1,40 @@
...
+    ) -> None:
+        """
+        CutMix's constructor.
+
+        :param cuts: a ``CutSet`` containing augmentation data, e.g. noise, music, babble.
+        :param snr: either a float, a pair (range) of floats, or ``None``.
+            It determines the SNR of the speech signal vs the noise signal that's mixed into it.
+            When a range is specified, we will uniformly sample SNR in that range.
+            When it's ``None``, the noise will be mixed as-is -- i.e. without any level adjustment.
+            Note that it's different from ``aug_snr=0``, which will adjust the noise level so that the SNR is 0.
```
danpovey, line 27: I think you mean ``snr=0``.

danpovey (commented):
Looks good to me!!
A few minor comments.

lhotse/dataset/transforms/cat.py
pzelasko: Good point, we might want to try removing that from snowfall expts. I am planning to implement a "BucketingCutSampler" soon that will group cuts based on histograms of durations and draw from a single bucket for any individual batch, to avoid wasting frames on padding as an alternative to this approach (should work great with dynamic batch size). We can check in snowfall then.

lhotse/dataset/transforms/cat.py
pzelasko: Does "pack" sound better maybe? Trying to avoid long names (also found cutcat funny somehow but clear would be indeed better)

lhotse/dataset/sampling.py
```diff
@@ -0,0 +1,288 @@
...
+        # We can only retrieve the cut ids once iter is called, not before!
+        # Iter is called inside a dataloader process, whereas init is called when
+        # instantiating this class (typically before dataloader).
+        self.cut_ids = partition_cut_ids(self.all_cut_ids, world_size=self.world_size, local_rank=self.local_rank)
+        if self.shuffle:
+            random.shuffle(self.cut_ids)
+        self.current_idx = 0
+        return self
+
+    @abstractmethod
```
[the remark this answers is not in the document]
pzelasko: This is a pythonic way of saying `virtual CutSet next() = 0` in C++; it will raise an exception saying that the method is abstract when you try to run it in the base class. That's why this class also inherits from `ABC`, you need both that (or setting the metaclass to `ABCMeta`) and the `abstractmethod` decorator to get this behavior.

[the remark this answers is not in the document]
pzelasko: Oh, and the `...` in the place of the function's body is equivalent to writing `pass`; in some codebases, I noticed that stubs like this are customarily ended with an ellipsis instead...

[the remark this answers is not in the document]
danpovey: I think pass is better because everyone knows what it is.

lhotse/dataset/transforms/cat.py
danpovey: I think concatenate is clearer than pack.

lhotse/dataset/transforms/cat.py
danpovey: I assume there is an option to have truly random sampling?  Would be nice if someone could try this in snowfall, it might slightly impove the results due to interaction with batchnorm (and also optimization issues if the batches' order is no randomized).

csukuangfj: I will try to test it in snowfall.

lhotse/dataset/sampling.py
```diff
@@ -0,0 +1,322 @@
...
+
+    Example usage::
+
+        >>> dataset = K2SpeechRecognitionIterableDataset(
+        ...     cuts,
+        ...     sampler=SingleCutSampler(cuts, world_size=4, local_rank=0)
+        ... )
+        >>> loader = DataLoader(dataset, batch_size=None)
+        >>> for epoch in range(start_epoch, n_epochs):
+        ...     dataset.sampler.set_epoch(epoch)
```
danpovey, line 34: this set_epoch stuff, I suppose we'd need to change the snowfall scripts..
although I believe I've seen a paper saying that using a deterministic order can work better.

danpovey (commented):
Generally looks OK; I don't know how much changes it would require to snowfall?
Would be nice if you could make a PR for that if it requires changes.

danpovey (approved):

lhotse/dataset/sampling.py
pzelasko: Interesting, could you share the reference?

pzelasko:
It will require some changes but not too drastic. I will keep snowfall in sync before I merge. But first, I'll check if I can use a map-style dataset like I mentioned in https://github.com/k2-fsa/snowfall/pull/71

lhotse/dataset/sampling.py
danpovey: Sorry couldn't find it.

pzelasko:
Map-style dataset works well. It's better because the `sampler` can go to `dataloader` now instead of the `dataset`; the separation of concerns is even better, and the user code looks more like typical PyTorch data code. 

This one is good to go. We should merge at the same time as https://github.com/k2-fsa/snowfall/pull/95

lhotse/dataset/transforms/cat.py
pzelasko: > I assume there is an option to have truly random sampling?

As a result of this PR, the shuffling is "truly random" now (before it was only "truly random inside partitions")

lhotse/dataset/sampling.py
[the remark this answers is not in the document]
csukuangfj: I find `...` is also used in PyTorch, see
https://github.com/pytorch/pytorch/blob/2f895f790a84bffd2aedb1dc184da9837ceb814a/torch/nn/modules/module.py#L704
```python
    @overload
    def to(self: T, device: Optional[Union[int, device]] = ..., dtype: Optional[Union[dtype, str]] = ...,
           non_blocking: bool = ...) -> T:
        ...

    @overload
    def to(self: T, dtype: Union[dtype, str], non_blocking: bool = ...) -> T:
        ...

    @overload
    def to(self: T, tensor: Tensor, non_blocking: bool = ...) -> T:
        ...
```

and
https://github.com/pytorch/pytorch/blob/2f895f790a84bffd2aedb1dc184da9837ceb814a/torch/nn/modules/module.py#L268
```python
    forward: Callable[..., Any] = _forward_unimplemented
```

---

It turns out `...` is an object in Python, similar to `None`. For example,

```python
>>> a = ...
>>> print(a)
Ellipsis
>>> type(a)
<class 'ellipsis'>
>>> b = ...
>>> id(a) == id(b)
True
>>> c = None
>>> print(c)
None
>>> type(None)
<class 'NoneType'>
```

So all of the following definitions have similar effects:

```python
def f(): pass
def f(): ...
def f(): None
def f(): 1
```

though `def f(): pass` is more common.

PR #197, opened 2021-02-10
BucketingSampler for batches with variable size and similar cut duration 

pzelasko:
I need to add the unit tests, but I tested it manually with mini librispeech and verified it works correctly. I gave some examples in the docs, hopefully it's clear enough to use. `BucketingSampler` uses `SingleCutSampler`/`CutPairsSampler` internally for each bucket, depending on users' choice. Thanks to that it doesn't have to know the filtering criteria, just passes them down to the specific sampler.

Batch durations from mini librispeech with max_frames=10000 look like the following:
```
[15.165, 15.165, 15.12, 15.04, 15.03, 14.99]
[14.19, 14.145, 14.05, 14.03, 13.985, 13.985, 13.92]
[14.685, 14.66, 14.575, 14.555, 14.555, 14.49]
[12.67, 12.655, 12.59, 12.585, 12.57, 12.4, 12.2699375]
[13.585, 13.435, 13.315, 13.315, 13.275, 13.275, 13.245]
[14.205, 14.135, 14.12, 14.115, 14.085, 14.065, 13.95]
[12.655, 12.59, 12.585, 12.585, 12.47, 12.465, 12.345, 12.265]
[14.155, 14.14, 14.1, 14.085, 14.06, 13.975, 13.925]
[15.12, 15.105, 15.075, 15.035, 14.985, 14.96]
[4.03, 3.985, 3.865, 3.8, 3.775, 3.63, 3.59, 3.475, 3.435, 3.355, 3.35, 3.31, 3.24, 3.225, 3.175, 2.985, 2.955, 2.95, 2.945, 2.925, 2.745, 2.67, 2.59, 2.575, 2.515, 2.485, 2.475, 2.44, 2.315, 2.21, 2.11, 1.99, 1.87]
[15.875, 15.815, 15.8, 15.8, 15.8, 15.785]
[16.025, 15.97, 15.84, 15.81, 15.81, 15.775]
[8.525, 8.47, 8.015, 7.915, 7.7, 7.4, 7.275, 7.19, 6.965, 6.95, 6.84, 6.5699375, 6.295]
[13.59, 13.515, 13.38, 13.37, 13.315, 13.315, 13.215]
[13.61, 13.485, 13.475, 13.465, 13.445, 13.395, 13.26]
[14.695, 14.695, 14.65, 14.59, 14.59, 14.55]
```

Will merge this after #194, and see how it affects snowfall training (probably after I finish the multi GPU stuff, but CLSP grid was down today so couldn't work on that).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,4 +69,4 @@ autodoc_default_options = {
     'exclude-members': '__weakref__'
 }
 
-autodoc_mock_imports = ["torchaudio"]
+autodoc_mock_imports = ["torchaudio", "SoundFile"]
--- a/docs/datasets.rst
+++ b/docs/datasets.rst
@@ -6,7 +6,6 @@ PyTorch Datasets
 
 We supply subclasses of the :class:`torch.data.Dataset` for various audio/speech tasks.
 These datasets are created from :class:`CutSet` objects and load the features from disk into memory on-the-fly.
-Each dataset accepts an optional ``root_dir`` argument which is used as a prefix for the paths to features and audio.
 
 Currently, we provide the following:
 
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -1338,7 +1338,9 @@ class CutSet(JsonMixin, YamlMixin, Sequence[AnyCut]):
 
     def subset(
             self,
+            *,  # only keyword arguments allowed
             supervision_ids: Optional[Iterable[str]] = None,
+            cut_ids: Optional[Iterable[str]] = None,
             first: Optional[int] = None,
             last: Optional[int] = None
     ) -> 'CutSet':
@@ -1351,12 +1353,13 @@ class CutSet(JsonMixin, YamlMixin, Sequence[AnyCut]):
             >>> train_set = cuts.subset(supervision_ids=train_ids)
             >>> test_set = cuts.subset(supervision_ids=test_ids)
 
-        :param supervision_ids: List of ``supervision_ids`` to keep.
+        :param supervision_ids: List of supervision IDs to keep.
+        :param cut_ids: List of cut IDs to keep.
         :param first: int, the number of first cuts to keep.
         :param last: int, the number of last cuts to keep.
         :return: a new ``CutSet`` with the subset results.
         """
-        assert exactly_one_not_null(supervision_ids, first, last), "subset() can handle only one non-None arg."
+        assert exactly_one_not_null(supervision_ids, cut_ids, first, last), "subset() can handle only one non-None arg."
 
         if first is not None:
             assert first > 0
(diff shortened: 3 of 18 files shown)


danpovey:
Piotr, I had an idea for dealing with context.  The idea is to avoid ever relying on the parts of the network output near the edges that would be affected by the padding.  (It won't be necessary to ensure that the amount of context we provide is actually >= the model context).  So we'd try to ensure that within a specified number of frames close to the edge, e.g. 10 or 20 frames, we never have a supervised segment (for ASR tasks, at least).  We would make the context be silence from the original where possible; or otherwise we could pad with silence or noise.

pzelasko:
Great, I was thinking along the same lines. I intend to finish the sampler and multi GPU stuff as this helps me with another project too, then add the hooks for graph construction (I think it should be simple/quick), and then get back to the context thing.

pzelasko:
I'm training a snowfall LFMMI model to see whether the WER/training time is affected. The batch size is now much more variable.

image

pzelasko:
Snowfall's LF-MMI result was the following:
```
2021-02-11 09:34:01,471 INFO [mmi_bigram_decode.py:291] %WER 10.50% [5523 / 52576, 775 ins, 487 del, 4261 sub ]
```

The training time for a single epoch seems to have went down from 49 minutes to 40 minutes - HOWEVER that may be just CLSP grid flakiness! Would be nice if somebody could compare `SingleCutSampler` vs `BucketingSampler` and let us know...

I'm merging.

danpovey:
OK, we need to start recording the results in the RESULTS.md every time we run this, along with a note of who ran it or at least context like a link to the PR.
Everyone please take note!!  I'm not even sure what the result was supposed to be (which is a problem...) I am concerned that this may have degraded WER results due to interaction with batchnorm.

danpovey:
mm the best result I see in my directory is 10.57%,that seems to be at epoch 7:
```
grep -w WER exp-lstm-adam-mmi-bigram-musan/log/log-decode-2021-0*
exp-lstm-adam-mmi-bigram-musan/log/log-decode-2021-02-04-18-40-55:2021-02-04 18:59:56,516 INFO [mmi_bigram_decode.py:295] %WER 10.57% [5555 / 52576, 671 ins, 596 del, 4288 sub ]
```

pzelasko:
OK, I'll make a PR to snowfall with an option to select either sampler, and update the results file.

pzelasko:
BTW why would this interact badly with batchnorm? In this sampling strategy, every cut is a separate sample in a batch (there is no concatenation), so they are independent samples (unlike when we concatenate them).

danpovey:
The concern is that with such varying lengths between minibatches, we
wouldn't expect the activations to have identical means and variances.

PR #202, opened 2021-02-22
Update most dataset classes and add docs for new the dataset concepts

pzelasko:


pzelasko:
In short, this PR brings most existing dataset classes up to speed with what `SpeechRecognitionDataset` does, and to use cut samplers. I described the motivation and general concepts in the docs. 

The only dataset type I did not update is speech separation; these are the "oldest" datasets in Lhotse (together with librimix data prep script) and I think they should be updated together (they can probably be simplified a lot now). Not sure if anybody's interested In them though so I don't want to spend the time on it for now.

danpovey (commented):
Like it!

PR #203, opened 2021-02-22
Proof of concept for creating K2 FSAs inside the DataLoader processes

pzelasko:
I'll test it in snowfall when I have more time (if it's needed sooner feel free to take it, just LMK). I think we should disable caching when doing it this way to avoid problems with increasing memory, as each dataloader process will keep its own cache (and I think by default pytorch re-starts the dataloader processes in each epoch, making caching super wasteful in such case).

@danpovey @csukuangfj

### DIFF
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -1,4 +1,4 @@
-from typing import Callable, Dict, List, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
 from torch.utils.data.dataloader import DataLoader, default_collate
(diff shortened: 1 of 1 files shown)


danpovey:
Great!
BTW at some point I believe we found that it was much faster to do the graph generation in parallel on GPU.  I'm pretty sure that must be supported now, possibly the default.

csukuangfj:
The removal of the cache was done in https://github.com/k2-fsa/snowfall/pull/84#issuecomment-771659598

Please see https://github.com/k2-fsa/snowfall/blob/master/snowfall/training/mmi_mbr_graph.py#L163

However, it is available only in `mmi_mbr_graph.py`.

`ctc_graph.py` and `mmi_graph.py` are not changed and are still using the cache.

pzelasko:
Good to know, thanks. In that case, do you think that it still makes sense to pursue this? I don’t think that the dataloader thread should be using the GPU, as it would have been in conflict with the main training thread.

danpovey:
I think it should be OK in principle to use the GPU even in another thread,
but either way this is still interesting.

pzelasko:
@danpovey so this essentially works on the Lhotse side but is not compatible with MMI training on the Snowfall side. Does it make sense to merge this, or should I close?

pzelasko:
If it's not helpful I'm not going to merge it for now. We can revisit this in the future

PR #204, opened 2021-02-24
Fix ASR dataset supervision validation

pzelasko:


pzelasko:
Resolves https://github.com/k2-fsa/snowfall/issues/107#issuecomment-785099472

Issue #205, opened 2021-02-26
LJspeech recipe error: exceeded the bounds of its corresponding recording 


janvainer:
The `prepare_ljspeech` recipe on version `lhotse==0.4.0` fails with the following error:
```
Traceback (most recent call last):
  File "examples/prepare_ljspeech.py", line 9, in <module>
    ljspeech = prepare_ljspeech("LJSpeech-1.1", "manifests")
  File "/home/jan/aligntts/venv/lib/python3.8/site-packages/lhotse/recipes/ljspeech.py", line 96, in prepare_ljspeech
    validate_recordings_and_supervisions(recording_set, supervision_set)
  File "/home/jan/aligntts/venv/lib/python3.8/site-packages/lhotse/qa.py", line 52, in validate_recordings_and_supervisions
    assert 0 <= s.start <= s.end <= r.duration, \
AssertionError: Supervision LJ001-0001: exceeded the bounds of its corresponding recording (supervision spans [0.0, 9.65501134]; recording spans [0, 9.65501133786848])
```
It runs fine on `lhotse==0.3.0`. There seems to be related a issue #201 - there is also **a rounding problem**.
I am looking into it now, but not sure where is the problem.

Edit:
 It seems like the duration of a [Recording](https://github.com/lhotse-speech/lhotse/blob/cd8ed4d2057f8d90da0362dd79331b6777af0086/lhotse/audio.py#L132) does not do the same rounding that [happens in Supervision](https://github.com/lhotse-speech/lhotse/blob/cd8ed4d2057f8d90da0362dd79331b6777af0086/lhotse/supervision.py#L25).

Supervision:
 ```
    @property
    def end(self) -> Seconds:
        return round(self.start + self.duration, ndigits=8)
```
Recording:
```
Recording(
            id=recording_id if recording_id is not None else sph_path.stem,
            sampling_rate=sphf.format['sample_rate'],
            num_samples=sphf.format['sample_count'],
            duration=sphf.format['sample_count'] / sphf.format['sample_rate'],
            sources=[
                AudioSource(
                    type='file',
                    channels=list(range(sphf.format['channel_count'])),
                    source=(
                        '/'.join(sph_path.parts[-relative_path_depth:])
                        if relative_path_depth is not None and relative_path_depth > 0
                        else str(sph_path)
                    )
                )
            ]
        )
```
Is it possible that there should also be rounding in the Recording? I cam make a PR for this :)

pzelasko:
[...]

PR #208, opened 2021-02-26
Add tolerance to validate_recordings_and_supervisions (#205)

janvainer:
Discussion is here: #205

### DIFF
--- a/lhotse/qa.py
+++ b/lhotse/qa.py
@@ -52,7 +52,7 @@ def validate_recordings_and_supervisions(
     # Errors
     for s in supervisions:
         r = recordings[s.recording_id]
-        assert 0 <= s.start <= s.end <= r.duration, \
+        assert -1e-3 <= s.start <= s.end <= r.duration + 1e-3, \
             f'Supervision {s.id}: exceeded the bounds of its corresponding recording ' \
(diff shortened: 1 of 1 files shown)


danpovey:
Guys, 
my wife Nadira was getting a problem with

ValueError: Cannot load features for recording None starting from 1003.52s. The available range is (1003.52001953125, 1021.28001953125) seconds.

presumably from this line:  https://github.com/lhotse-speech/lhotse/blob/c1d443288373ceaf50aee665a9bc9831a82356a4/lhotse/features/base.py#L475

... I'm not 100% sure of the workflow she was doing & where she got this manifest, but do you have any idea what is the recommended fix at this point?

Screen Shot 2024-06-02 at 11 06 12 PM

pzelasko:
That's weird, the diff is less than the duration of a single sample for 16kHz. But anyway that tolerance seems too tight for features, we should probably change it to sth like 1e-3 to avoid these issues.

pzelasko:
Try with this PR https://github.com/lhotse-speech/lhotse/pull/1349

Another possible workaround is to manually fix the manifest updating feature.start = cut.start

npovey:
In the example above I was just running libriheavy script on "medium" subset.  

PR #1349 fixed my problem. 
thanks

PR #211, opened 2021-02-27
Return lengths from feature and token collations

janvainer:
Summary:

* Return feature lengths from collate_features
* Add TokenCollater and corresponding test
* Add Standardize feature transform
* SpeechSynthesisDataset returns lengths for tokens and features
* SpeechSynthesisDataset applies feature transforms after collation
----

Problem: Where to pad features, normalize features, tokenize etc?

Proposal: Use stateful Collate functions that can be passed by user, but are pre-set with reasonable defaults.

Example:
```
collate_features = CollateFeatures(
    pad_mode="reflect",
    normalize_features=StandardizeFeatures(**cuts.compute_global_feature_stats())
)
dataset = SpeechSynthesisDataset(cuts, collate_features=collate_features)
```
Benefits:

1. Stateful collations allow user to pre-configure eg. normalization with pre-computed mean and std without using `partial`
2. More control over padding schemes, tokenization, normalization etc. 
3. Does not bloat already heavy CutSet with additional functionality
4. Allows user to create custom collations
5. Simple enough - can work with simple collation function without any state, can be completely be avoided by user if she is satisfied with the defaults.

More Ideas/Questions:
1. Should collation and online feature augmentation be treated on the same level? (eg they would be passed in a list of callables)I think maybe not, because collation returns also feature lengths
2. Should the normalization be considered online augmentation instead of part of Collation? (probably yes)

More generic suggestion:
```
collate_features = CollateFeatures(
    pad_mode="reflect",
)
transform_features = Transforms(
    StandardizeFeatures(**cuts.compute_global_feature_stats()),
    SpecAugment(some_other_args),
)

dataset = SpeechSynthesisDataset(cuts, collate_features=collate_features, transform_features=transform_features)
```

The tests may fail now, but I will fix that based on reactions to the proposal :)

### DIFF
--- a/lhotse/dataset/collation.py
+++ b/lhotse/dataset/collation.py
@@ -1,4 +1,4 @@
-from typing import Iterable, Union
+from typing import Iterable, Union, Tuple, List
 
 import numpy as np
 import torch
@@ -8,34 +8,127 @@ from lhotse import CutSet
 from lhotse.cut import MixedCut
 
 
-def collate_features(cuts: CutSet) -> torch.Tensor:
+class TokenCollater:
+    """Collate list of tokens
+
+    Map sentences to integers. Sentences are padded to equal length.
+    Beginning and end-of-sequence symbols can be added.
+    Call .inverse(tokens_batch, tokens_lens) to reconstruct batch as string sentences.
+
+    Example:
+        >>> token_collater = TokenCollater(cuts)
+        >>> tokens_batch, tokens_lens = token_collater(cuts.subset(first=32))
+        >>> original_sentences = token_collater.inverse(tokens_batch, tokens_lens)
+
+    Returns:
+        tokens_batch: IntTensor of shape (B, L)
+            B: batch dimensoion, number of input sentences
+            L: length of the longest sentence
+        tokens_lens: IntTensor of shape (B,)
+            Length of each sentence after adding <eos> and <bos>
+            but before padding.
+    """
+    def __init__(
+        self,
+        cuts: CutSet,
+        add_eos: bool = True,
+        add_bos: bool = True,
+        pad_symbol: str = "<pad>",
+        bos_symbol: str = "<bos>",
+        eos_symbol: str = "<eos>",
+        unk_symbol: str = "<unk>",
+    ):
+        self.pad_symbol = pad_symbol
+        self.bos_symbol = bos_symbol
+        self.eos_symbol = eos_symbol
+        self.unk_symbol = unk_symbol
+
+        self.add_eos = add_eos
+        self.add_bos = add_bos
+
+        tokens = {
+            char for cut in cuts
+            for char in cut.supervisions[0].text
+        }
+        tokens_unique = (
+            [pad_symbol, unk_symbol]
+            + ([bos_symbol] if add_bos else [])
+            + ([eos_symbol] if add_eos else [])
+            + sorted(tokens)
+        )
+
+        self.token2idx = {token: idx for idx, token in enumerate(tokens_unique)}
+        self.idx2token = [token for token in tokens_unique]
+
+    def __call__(self, cuts: CutSet) -> Tuple[torch.Tensor, torch.Tensor]:
+        token_sequences = [
+            " ".join(supervision.text for supervision in cut.supervisions)
+            for cut in cuts
+        ]
+        max_len = len(max(token_sequences, key=len))
+
+        seqs = [
+            ([self.bos_symbol] if self.add_bos else [])
+            + list(seq)
+            + ([self.eos_symbol] if self.add_eos else [])
+            + [self.pad_symbol] * (max_len - len(seq))
+            for seq in token_sequences
+        ]
+
+        tokens_batch = torch.from_numpy(np.array([
+            [self.token2idx[token] for token in seq]
+            for seq in seqs
+        ], dtype=np.int32))
+
+        tokens_lens = torch.IntTensor([
+            len(seq) + int(self.add_eos) + int(self.add_bos)
+            for seq in token_sequences
+        ])
+
+        return tokens_batch, tokens_lens
+
+    def inverse(self, tokens_batch: torch.IntTensor, tokens_lens: torch.IntTensor) -> List[str]:
+        start = 1 if self.add_bos else 0
+        sentences = [
+            "".join([
+                self.idx2token[idx]
+                for idx in tokens_list[start:end - int(self.add_eos)]
+            ])
+            for tokens_list, end in zip(tokens_batch, tokens_lens)
+        ]
+        return sentences
+
+
+def collate_features(cuts: CutSet) -> Tuple[torch.Tensor, torch.IntTensor]:
     """
     Load features for all the cuts and return them as a batch in a torch tensor.
     The output shape is ``(batch, time, features)``.
     The cuts will be padded with silence if necessary.
     """
     assert all(cut.has_features for cut in cuts)
+    features_lens = torch.tensor([cut.num_frames for cut in cuts], dtype=torch.int)
     cuts = maybe_pad(cuts)
     first_cut = next(iter(cuts))
     features = torch.empty(len(cuts), first_cut.num_frames, first_cut.num_features)
     for idx, cut in enumerate(cuts):
         features[idx] = torch.from_numpy(cut.load_features())
-    return features
+    return features, features_lens
 
 
-def collate_audio(cuts: CutSet) -> torch.Tensor:
+def collate_audio(cuts: CutSet) -> Tuple[torch.Tensor, torch.IntTensor]:
     """
     Load audio samples for all the cuts and return them as a batch in a torch tensor.
     The output shape is ``(batch, time)``.
     The cuts will be padded with silence if necessary.
     """
     assert all(cut.has_recording for cut in cuts)
+    audio_lens = torch.tensor([cut.num_samples for cut in cuts], dtype=torch.int)
     cuts = maybe_pad(cuts)
     first_cut = next(iter(cuts))
     audio = torch.empty(len(cuts), first_cut.num_samples)
     for idx, cut in enumerate(cuts):
         audio[idx] = torch.from_numpy(cut.load_audio()[0])
-    return audio
+    return audio, audio_lens
 
 
 def collate_multi_channel_features(cuts: CutSet) -> torch.Tensor:
--- a/lhotse/dataset/diarization.py
+++ b/lhotse/dataset/diarization.py
@@ -28,6 +28,7 @@ class DiarizationDataset(Dataset):
 
         {
             'features': (B x T x F) tensor
+            'features_lens': (B, ) tensor
             'speaker_activity': (B x num_speaker x T) tensor
         }
 
@@ -60,8 +61,10 @@ class DiarizationDataset(Dataset):
 
     def __getitem__(self, cut_ids: Iterable[str]) -> Dict[str, torch.Tensor]:
         cuts = self.cuts.subset(cut_ids=cut_ids)
+        features, features_lens = collate_features(cuts)
         return {
-            'features': collate_features(cuts),
+            'features': features,
+            'features_lens': features_lens,
             'speaker_activity': collate_matrices(
                 (cut.speakers_feature_mask(
                     min_speaker_dim=self.min_speaker_dim,
--- a/lhotse/dataset/speech_synthesis.py
+++ b/lhotse/dataset/speech_synthesis.py
@@ -1,10 +1,10 @@
-from typing import Dict, Iterable
+from typing import Dict, Iterable, Callable, Union, Sequence
 
 import torch
 
 from lhotse import validate
 from lhotse.cut import CutSet
-from lhotse.dataset.collation import collate_audio, collate_features
+from lhotse.dataset.collation import TokenCollater, collate_features, collate_audio
 
 
 class SpeechSynthesisDataset(torch.utils.data.Dataset):
@@ -15,38 +15,61 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
     .. code-block::
 
         {
-            'audio': (B x NumSamples) tensor
-            'features': (B x NumFrames x NumFeatures) tensor
-            'tokens': list of lists of characters (str)
+            'audio': (B x NumSamples) float tensor
+            'features': (B x NumFrames x NumFeatures) float tensor
+            'tokens': (B x NumTokens) int tensor
+            'audio_lens': (B, ) int tensor
+            'features_lens': (B, ) int tensor
+            'tokens_lens': (B, ) int tensor
         }
     """
 
-    def __init__(self, cuts: CutSet) -> None:
+    def __init__(
+        self,
+        cuts: CutSet,
+        feature_transforms: Union[Sequence[Callable], Callable] = None,
+        add_eos: bool = True,
+        add_bos: bool = True,
+    ) -> None:
         super().__init__()
-        validate(cuts)
-        self.cuts = cuts
 
-        # generate tokens from text
-        self.cut_id_to_token = {}
-        self.tokens = set()
+        validate(cuts)
         for cut in cuts:
-            assert len(cut.supervisions) == 1, 'Only the Cuts with single supervision are supported.'
-            characters = list(cut.supervisions[0].text)
-            self.tokens.update(set(characters))
-            self.cut_id_to_token[cut.id] = characters
-        self.tokens = sorted(list(self.tokens))
+            assert (
+                len(cut.supervisions) == 1
+            ), "Only the Cuts with single supervision are supported."
+
+        self.cuts = cuts
+        self.token_collater = TokenCollater(cuts, add_eos=add_eos, add_bos=add_bos)
+
+        if feature_transforms is None:
+            feature_transforms = []
+        elif not isinstance(feature_transforms, Sequence):
+            feature_transforms = [feature_transforms]
+
+        assert all(isinstance(transform, Callable) for transform in feature_transforms), \
+            "Feature transforms must be Callable"
+        self.feature_transforms = feature_transforms
 
     def __getitem__(self, cut_ids: Iterable[str]) -> Dict[str, torch.Tensor]:
         cuts = self.cuts.subset(cut_ids=cut_ids)
 
-        features = collate_features(cuts)
-        audio = collate_audio(cuts)
+        audio, audio_lens = collate_audio(cuts)
+        features, features_lens = collate_features(cuts)
+
+        for transform in self.feature_transforms:
+            features = transform(features)
+
+        tokens, tokens_lens = self.token_collater(cuts)
+
         return {
-            'audio': audio,
-            'features': features,
-            # TODO: consider extending the dataset to create a collated tensor of integer token IDs
-            'tokens': [self.cut_id_to_token[cut.id] for cut in cuts]
+            "audio": audio,
+            "features": features,
+            "tokens": tokens,
+            "audio_lens": audio_lens,
+            "features_lens": features_lens,
+            "tokens_lens": tokens_lens,
         }
 
     def __len__(self) -> int:
-        return len(self.cut_ids)
+        return len(self.cuts)
--- a/lhotse/dataset/unsupervised.py
+++ b/lhotse/dataset/unsupervised.py
@@ -12,8 +12,13 @@ from lhotse.features import FeatureExtractor
 class UnsupervisedDataset(torch.utils.data.Dataset):
     """
     Dataset that contains no supervision - it only provides the features extracted from recordings.
-    The returned features are a :class:`torch.Tensor` of shape ``(T x F)``, where T is the number of frames,
-    and F is the feature dimension.
+
+    .. code-block::
+
+        {
+            'features': (B x T x F) tensor
+            'features_lens': (B, ) tensor
+        }
     """
 
     def __init__(self, cuts: CutSet) -> None:
@@ -23,7 +28,11 @@ class UnsupervisedDataset(torch.utils.data.Dataset):
 
     def __getitem__(self, cut_ids: Iterable[str]) -> torch.Tensor:
         cuts = self.cuts.subset(cut_ids=cut_ids)
-        return collate_features(cuts)
+        features, features_lens = collate_features(cuts)
+        return {
+            'features': features,
+            'features_lens': features_lens,
+        }
 
     def __len__(self):
         return len(self.cuts)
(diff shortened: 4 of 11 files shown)


lhotse/dataset/collation.py
```diff
@@ -1,11 +1,109 @@
...
+        max_len = len(max(token_sequences, key=len))
+        seqs = [
+            ([self.bos_symbol] if self.add_bos else [])
+            + seq
+            + ([self.eos_symbol] if self.add_eos else [])
+            + [self.pad_symbol] * (max_len - len(seq))
+            for seq in token_sequences
+        ]
+
+        tokens_batch = torch.Tensor([[self.token2int(token) for token in seq] for seq in seqs], dtype=int)
```
danpovey, line 44: this doesn't work for me in torch 1.6.0 (this method of creating a tensor as torch.Tensor([list], dtype=int).
also int appears to map to torch.int64, better to use torch dtype.

lhotse/dataset/speech_synthesis.py
```diff
@@ -15,38 +16,66 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
...
-        self.cut_id_to_token = {}
-        self.tokens = set()
-        for cut in cuts:
-            assert len(cut.supervisions) == 1, 'Only the Cuts with single supervision are supported.'
-            characters = list(cut.supervisions[0].text)
-            self.tokens.update(set(characters))
-            self.cut_id_to_token[cut.id] = characters
-        self.tokens = sorted(list(self.tokens))
+        self.collate_tokens = (
+            CollateTokens(unique_tokens(self.cuts))
```
danpovey, line 45: I don't understand how it manages to preserve the mapping between token-sequences and feature-sequences, with this unique_tokens() call.

janvainer: The initialization creates a mapping from tokens to integers based on all used tokens in the dataset. This mapping is then used to map tokens loaded in the dataset to integers through the CollateTokens.__call__ method. Or maybe I do not understand the question correctly. Could you try to re-iterate?

lhotse/dataset/collation.py
janvainer: agreed. Btw I haven't tested all code aspects. This is more like a design idea - a draft PR.

janvainer: fixed

pzelasko:
Thanks! I just skimmed through it - I will need to read more in-depth to develop a more formed opinion. I think this PR is helpful in converging on the design of these things.

I’m not sure if we really need to customize the collation fn - I think that any given task-specific dataset has a single „correct” method of collating the data? Then it makes sense to have it coded directly in dataset definition. OTOH one argument against that approach is that we might have single and multi channel versions of some tasks, and we will have to decide where to have separate dataset classes for that or a single one with input customization. I am leaning slightly towards the former — I had an attempt at a „generic” dataset class that is very customizable and I think it was confusing for others.

Last comment - another factor is also „on-the-fly” feature extraction, should there be a separate dataset class for that or another customization point for the existing classes. Again, I am slightly leaning towards having a lot of modularized functionality, so that writing each dataset classes takes 20~30 LOC — rather having 10+ arguments that customize a single class. I think it’s just harder for the end-users to get a grasp on that.

janvainer:
_...any given task-specific dataset has a single „correct” method of collating the data_

I agree with this. But even if there is a single correct way how to do collations, there are small tweaks that need configuring. For example, the user may need to specify if the `<bos>` or the `<eos>` token should be used or what padding scheme should be used. I think the collation params could be passed directly to the dataset, and the augmentation features could be passed as a list of transforms. Continuing with speech synthesis example..

The args would be:
1. collations - only pass basic args such as `add_bos`, initialize inside dataset
2. online transforms - pass list of transforms objects

```
dataset =  SpeechSynthesisDataset(
    cuts,
    feature_transforms=Transforms(  # follows audiomentations-like api
        Rescale(0, 1),
    ),
    feature_pad_mode=reflect,
    use_bos=True,
    use_eos=True,
)

...
# the collation functions are configured internally inside the dataset without direct user control
feature_collate = CollateFeatures(pad=reflect, dtype=torch.float32), 
token_collate = CollateTokens(unique_tokens, use_bos=True, use_eos=False) 
```

If the above params could not be at least slightly customized, there would literaly be a separate dataset for every speech synthesis model (and they would be almost the same except for some small changes), which is not great either. How about a mix:

1. Have one very configurable SpeechSynthesisDataset like the original one that I posted with composable abstractions
2. Have wrapper datasets for specific speech synthesis tasks that would pre-configure the SpeechSynthesisDataset for the user with settings from the original papers.
    - FastSpeechDataset
    - SpeedySpeech dataset
    - Tacotron 2 dataset
    - ParallelWaveGan dataset etc.
It would work just like the dataset recipes that lhotse already provides - eg prepare_ljspeech. The recipes take specific dataset and return common representation - CutSet. The wrapper datasets would take specific speech synthesis subtask and return a preconfigutred common representation - the SpeechSynthesisDataset.

This would break eg in the case of FastSpeech or SpeedySpeech that also need **durations** of phonemes/chars for training which are not returned right now and also in multi-speaker case, where some **speaker embeddings** have to be used. But this could be solved with inheritance - FastSpeechDataset would call SpeechSynthesisDataset.__get__item__() and simply add its own additional keys to the batch dict.

lhotse/dataset/collation.py
```diff
@@ -1,11 +1,109 @@
...
 import torch
 from torch.nn import CrossEntropyLoss
 
 from lhotse import CutSet
 from lhotse.cut import MixedCut
+from lhotse.utils import LOG_EPSILON
+
+
+@dataclass
+class CollateTokens:
```
pzelasko, line 14: style - either it's mimicking a function and then we should name it `collate_tokens`, or it's an object and then we should name it `TokenCollater` - I am leaning towards the latter

same for other classes of this type

lhotse/dataset/collation.py
```diff
@@ -1,11 +1,109 @@
...
+
+    tokens: Iterable[str]
```
pzelasko, line 16: I think this + post init is a hack in this case, because we don't need the original iterable after initialization. I think this should be a regular class with init that either accepts an iterable or a previously prepared mapping (or one of both)

lhotse/dataset/collation.py
```diff
@@ -1,11 +1,109 @@
...
+    def __call__(self, features):
+        return self.new_min + (features - self.feature_min) * self.new_range / self.old_range
+
+    def inverse(self, features):
+        return self.feature_min + (features - self.new_min) * self.old_range / self.new_range
+
+
+@dataclass
+class CollateFeatures:
+    pad_mode: Literal["constant", "reflect"] = "constant"
```
pzelasko, line 82: Why do we need "reflect"?

lhotse/dataset/collation.py
```diff
@@ -1,11 +1,109 @@
...
+    pad_value: float = LOG_EPSILON
+    normalize_features: Callable = None
```
pzelasko, line 84: as discussed this should be separated from this class

lhotse/dataset/collation.py
```diff
@@ -94,18 +192,19 @@ def collate_vectors(
...
     return result
 
 
 def collate_matrices(
-        tensors: Iterable[Union[torch.Tensor, np.ndarray]],
-        padding_value: Union[int, float] = 0,
-        matching_shapes: bool = False
+    tensors: Iterable[Union[torch.Tensor, np.ndarray]],
+    padding_value: Union[int, float] = 0,
+    matching_shapes: bool = False,
```
pzelasko, line 207: which style checker are you using?

pzelasko (commented):
OK, I like where this is going. Some comments:

Re `SpeechSynthesisDataset` - I think this looks good, let's do this.

**Re phone durations:** - for now, Lhotse does not have an explicit representation for alignment, so getting this might be a bit difficult right now. We'll need to think how to represent it consistently with K2/Snowfall at some point (it will probably be an FSA, so it might be a bit tricky not to introduce a K2 dependency in Lhotse - @danpovey)

**Re speaker embeddings** - we can make it a convention that when a specific field is present in the SupervisionSegment object, we use it to read the speaker embeddings and automatically add them to batches. Similarly for other types of inputs and supervision, which might help us reduce the number of inherited dataset classes in total.


> But this could be solved with inheritance - FastSpeechDataset would call SpeechSynthesisDataset.__getitem__() and simply add its own additional keys to the returned batch dict.

I am not sure I like this - I think it makes the extension quite complicated since it's done after collation. But it's either some form of inheritance, or hooks with sth like "field" objects ([considered previously here](https://github.com/lhotse-speech/lhotse/pull/172/files#diff-6256dd88bea7f5803dffb53146cc885012a973dfa4d7b2b4824025bcae89a98bR1)), or "one-dataset-rules-them-all", or manual re-writing. TBH I think I still need to collect more data/use-cases to make a final decision on which way to go. So we can implement whichever of those but as far as I'm concerned it's still subject to change in the future.

lhotse/dataset/collation.py
janvainer: Fair point, I was a bit lazy with making regular init. Will change it to regular classes :)

lhotse/dataset/collation.py
janvainer: Agree, lets use `TokenCollater`

lhotse/dataset/collation.py
janvainer: Reflect can be used for batch norm to have unbiased estimate for batch variance (which usually happens if zero padding or other constant is used.) I put this here to show what configurations could be used and left the implementation for this blank. We can completely purge it for now.

lhotse/dataset/collation.py
janvainer: I used black with default settings.

pzelasko: I guess I'll need to make it the chosen autoformatter for Lhotse eventually, as the number of contributors grows. I wonder if it's possible to prevent it from changing unmodified code for now, as that could mess up git blame and makes PRs less readable in terms of actual changes.

janvainer: I will re-write the files anyway, so I ll keep the formatting as it was for this PR in the next push (probably tomorrow).

janvainer:
Thanks for the review! 

**RE SpeechSynthesisDataset**: Great, I will polish the PR and resolve your style comments and implement the agreed API.
**RE Phone Durations**: I was thinking monotonic alignment (durations) could be stored as an array of integers and be loaded just like the features are loaded now, but not sure how this would interact with k2.
**RE Speaker embeddings**: I like the idea of putting them as fields in the SupervisionSegments. Autobatching of such fields sounds doable.
**RE Inheritance**: Good point with the after-collation. On the other hand, the collation happens separately for each field in the returned dict so inheritance would just chain one more collation on one more field or so. Could you let me know once the decision is made please? I would definitely be interested in pros and cons of the design in this case as its quite unclear which way would be the least painful.

pzelasko:
Cool. Your have a good point about the collation too - I'm OK with inheritance for now if needed (maybe we can delay the need for that using custom fields in `SupervisionSegment` for now). You can store phone durations array in `SupervisionSegment.custom` field BTW, and make the dataset code look for it there.

janvainer:
Ok it should be ready now. I decided not to overcomplicate things for now and left some of the things we mentioned for later.

lhotse/dataset/collation.py
```diff
@@ -8,19 +8,115 @@
...
     """
     assert all(cut.has_features for cut in cuts)
+    features_lens = torch.tensor([cut.num_frames for cut in cuts], dtype=torch.int)
     cuts = maybe_pad(cuts)
     first_cut = next(iter(cuts))
     features = torch.empty(len(cuts), first_cut.num_frames, first_cut.num_features)
     for idx, cut in enumerate(cuts):
         features[idx] = torch.from_numpy(cut.load_features())
-    return features
+    return features, features_lens
```
pzelasko, line 119: Do you mind updating the rest of `collate_X` functions? There is an API mismatch at the moment, and they would have also benefitted from your changes.

lhotse/dataset/collation.py
```diff
@@ -8,19 +8,115 @@
...
+            + sorted(tokens)
+        )
+
+        self.token2idx = {token: idx for idx, token in enumerate(tokens_unique)}
+        self.idx2token = [token for token in tokens_unique]
+
+    def __call__(self, cuts: CutSet) -> Tuple[torch.Tensor, torch.Tensor]:
+
+        assert all(len(cut.supervisions) == 1 for cut in cuts), \
+            "Expects only Cuts with a single supervision"
```
pzelasko, line 69: Is that a fundamental limitation? I think we could just concatenate two utterances' texts.

lhotse/dataset/collation.py
```diff
@@ -8,19 +8,115 @@
...
+        }
+        tokens_unique = (
+            [pad_symbol, unk_symbol]
+            + ([bos_symbol] if add_bos else [])
+            + ([eos_symbol] if add_eos else [])
+            + sorted(tokens)
+        )
+
+        self.token2idx = {token: idx for idx, token in enumerate(tokens_unique)}
+        self.idx2token = [token for token in tokens_unique]
```
pzelasko, line 64: it is a cosmetic change but consider sorting idx2token and then creating token2idx from that; it might be a bit faster to debug/lookup stuff

lhotse/dataset/diarization.py
```diff
@@ -61,7 +61,7 @@ def __init__(
     def __getitem__(self, cut_ids: Iterable[str]) -> Dict[str, torch.Tensor]:
         cuts = self.cuts.subset(cut_ids=cut_ids)
         return {
-            'features': collate_features(cuts),
+            'features': collate_features(cuts)[0],
```
pzelasko, line 64: I think this dataset could possibly benefit from adding `features_lens` as a field in batch

lhotse/dataset/speech_synthesis.py
```diff
@@ -15,38 +15,54 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
...
-        # generate tokens from text
-        self.cut_id_to_token = {}
-        self.tokens = set()
-        for cut in cuts:
-            assert len(cut.supervisions) == 1, 'Only the Cuts with single supervision are supported.'
-            characters = list(cut.supervisions[0].text)
-            self.tokens.update(set(characters))
-            self.cut_id_to_token[cut.id] = characters
-        self.tokens = sorted(list(self.tokens))
+        if not isinstance(feature_transforms, Iterable):
```
pzelasko, line 44: I think it has to be at least `Sequence`, `Iterable` can be a one-pass iterator

lhotse/dataset/speech_synthesis.py
```diff
@@ -15,38 +15,54 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
...
+            'features_lens': (B x NumFrames) int tensor
+            'tokens_lens': (B x NumTokens) int tensor
         }
     """
 
-    def __init__(self, cuts: CutSet) -> None:
+    def __init__(
+        self,
+        cuts: CutSet,
+        feature_transforms: Union[Iterable[Callable], Callable] = [],
```
pzelasko, line 29: mutable default arguments are dangerous; it is better to write `= None` and handle the case with `if x is None: self.x = []`

lhotse/dataset/collation.py
janvainer: Good point, I am off the computer already, so I will push an update tomorrow :)

lhotse/dataset/vad.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, cuts: CutSet) -> None:
     def __getitem__(self, cut_ids: Iterable[str]) -> Dict[str, torch.Tensor]:
         cuts = self.cuts.subset(cut_ids=cut_ids).sort_by_duration()
         return {
-            'features': collate_features(cuts),
+            'features': collate_features(cuts)[0],
```
pzelasko, line 32: can we add `features_lens` here?

lhotse/dataset/unsupervised.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, cuts: CutSet) -> None:
 
     def __getitem__(self, cut_ids: Iterable[str]) -> torch.Tensor:
         cuts = self.cuts.subset(cut_ids=cut_ids)
-        return collate_features(cuts)
+        return collate_features(cuts)[0]
```
pzelasko, line 26: can we add `features_lens` here?

test/dataset/test_collation.py
```diff
@@ -0,0 +1,27 @@
...
+            idx,
+            idx,
+            supervisions=[dummy_supervision(
+                idx, idx, text=sentence
+            )]
+        ) for idx, sentence in enumerate(test_sentences)
+    )
+
+    token_collater = TokenCollater(cuts)
+    tokens_batch, tokens_lens = token_collater(cuts)
```
pzelasko, line 24: would be good to assert some properties of tokens_batch and tokens_lens too (that they are int tensors, have the right shapes, and maybe specific values for lens)

test/dataset/test_speech_synthesis_dataset.py
```diff
@@ -11,7 +12,7 @@ def cut_set():
 
 def test_speech_synthesis_dataset(cut_set):
     ids = cut_set.ids
-    dataset = SpeechSynthesisDataset(cut_set)
+    dataset = SpeechSynthesisDataset(cut_set, feature_transforms=Standardize(cut_set))
```
pzelasko, line 15: Can we make it a parametrized test?

```python
@pytest.mark.parametrize('transform', [None, Standardize])
def test_speech_synthesis_dataset(cut_set, transform):
    transform = transform(cut_set) if transform is not None else None
    dataset = SpeechSynthesisDataset(cut_set, feature_transforms=transform)
```

pzelasko (commented):
Looks good! I left some comments.

pzelasko:
> Ok it should be ready now. I decided not to overcomplicate things for now and left some of the things we mentioned for later.

Sounds good to me - let's take it one step at a time :)

lhotse/dataset/collation.py
janvainer: Hm I guess we can. I can not think of a scenario when this would be a problem tbh.

lhotse/dataset/collation.py
janvainer: Thanks, good suggestion. Will adjust.

lhotse/dataset/diarization.py
janvainer: I am thinking of adding features lens to most of the datasets, maybe except the K2 dataset. WYT?

lhotse/dataset/speech_synthesis.py
janvainer: good catch, fixing

lhotse/dataset/speech_synthesis.py
janvainer: Should it then be typed with Optional? `Optional[Sequence[Callable]] = None`?

lhotse/dataset/vad.py
janvainer: yes :)

lhotse/dataset/unsupervised.py
janvainer: yes :D

test/dataset/test_collation.py
janvainer: adding

test/dataset/test_speech_synthesis_dataset.py
janvainer: Good idea. Thanks!

lhotse/dataset/collation.py
janvainer: Looking more into it, sorting the idx2token would mean that the pad symbol would not have the index zero, which is default padding index eg in pytorch Embedding layer. Or am I missing something?

lhotse/dataset/speech_synthesis.py
pzelasko: I think it's going to become difficult to read. I'm OK to omit Optional in such cases.

lhotse/dataset/collation.py
pzelasko: Oh sorry I didn't notice it is already sorted a few lines before. My bad!

janvainer:
I pushed some changes.

- parameterized tests and more checks
- length is returned from more datasets

lhotse/dataset/speech_synthesis.py
```diff
@@ -15,38 +15,60 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
...
-            characters = list(cut.supervisions[0].text)
-            self.tokens.update(set(characters))
-            self.cut_id_to_token[cut.id] = characters
-        self.tokens = sorted(list(self.tokens))
+        if feature_transforms is None:
+            feature_transforms = []
+        elif not isinstance(feature_transforms, Sequence):
+            feature_transforms = [feature_transforms]
+
+        print(feature_transforms)
```
pzelasko, line 49: left-over print

test/dataset/test_speech_synthesis_dataset.py
```diff
@@ -9,9 +10,12 @@ def cut_set():
     return CutSet.from_json('test/fixtures/ljspeech/cuts.json')
 
 
-def test_speech_synthesis_dataset(cut_set):
+@pytest.mark.parametrize('transform', [None, Standardize])
```
pzelasko, line 13: actually we should add a third argument `[Standardize]` in there to ensure that it also works with sequence arguments

lhotse/dataset/speech_synthesis.py
janvainer: thanks, fixing

test/dataset/test_speech_synthesis_dataset.py
janvainer: true, adding

janvainer:
It seems like the tests in the CI sometimes fail even though I always test everything locally before pushing. Here are the cases I recall something weird happened:

1. All python versions except python3.6 passed. The 3.6 version passed due to some timeout error
2. Some randomized test failed, specifiaclly the `test/test_resample_randomized.py`

I am not able to reproduce any of these locally, but they happened in CI for some reason.

pzelasko:
Sorry about that. The timeout error should already be fixed in latest master. The resample error is on my radar, I'll get to it soon. A fun fact is that the randomized tests failures can be replicated - the test output has a line like `use @reproduce_failure(...) decorator to reproduce`, you have to add that decorator on top of the function and you can easily replicate and debug then (but don't - I will take care of this particular bug).

janvainer:
Cool, good to hear. I was not sure if this was being handled. :)

lhotse/dataset/collation.py
pzelasko: I think at the very least add lengths to collate audio below, and update TTS dataset to have `audio_lens` too. Multi-channel collations are ok the be left behind for now. 

I'll merge after that. Thanks!

janvainer: I added the audio_lens, adjusted test for the unsupervised audio dataset and added audio_lens to the speech synthesis dataset 👍🏻

pzelasko (approved):
Looks good to me, merging!

Issue #212, opened 2021-02-28
Bucket sampler __len__ gives inconsistent result w.r.t the number of actually sampled batches


janvainer:
There seems to be an inconsitency in how length of the lhotse samplers is calculated. Since the dataset returns full batches with a single call to `__getitem__`, the length of the dataset should correspond to the total number of batches that can be sampled without repetition. The dataset length in the DataLoader is accessed through the Sampler, but the length of lhote samplers equals the number of provided cuts, which does not equal the number of batches that will be sampled in one epoch.

Example:
```
batch_size = 16
# sample buckets of size 16
train_sampler = BucketingSampler(cuts, max_frames=float("inf"), max_cuts=batch_size)
train_loader = DataLoader(
       dataset, # created from cuts
       sampler=train_sampler,
       batch_size=None,
)

i = 0
for b in train_loader:
    i += 1

print(i)  # 730
print(len(train_sampler))  # 11611 <- this should probably equal 730
print(len(cuts))  # 11611
```

This is problematic for example when you want to train your model with pytorch lightning. They save checkpoints after several epochs, but with SimpleCutSampler no epoch is ever completed and the checkpoints do not get saved to disk. it is also a bit confusing from the usage standpoint. 
I guess it would be a bit difficult to know ahead of time, how many batches there will be if the SingleCutSampler is used (except for the example shown above). Maybe we could pre-determine (force) how many batches will be sampled with the SImpleCutSampler ahead of time given that we may skip some datapoints in the given epoch. Or we could simulate the sampling process with fixed random seed before each epoch (should be ok for medium sized datasets) and know the number of batches exactly. WYT?

pzelasko:
Thanks for bringing this up! I agree that we should respect these conventions. Your last suggestion seems like a natural extension to the current samplers. I think we can do this by separating the sampling logic from batch iteration, so that we’re able to „simulate” what batches are returned in a full epoch and return an accurate number of batches. The random seed mechanism is actually implemented there already.

pzelasko:
Forgot to ask - would you like to take this one?

janvainer:
Sure, will look into it :)

janvainer:
Added PR with a fix #213

pzelasko:
Solved by #213

PR #213, opened 2021-03-01
Make sampler length equal to the number of returned batches (#212)

janvainer:
Solves issue #212 

* Add _len attribute and _set_sampler_len method
* Add tests that check len(sampler) == len([batch for batch in sampler])

### DIFF
--- a/lhotse/dataset/sampling.py
+++ b/lhotse/dataset/sampling.py
@@ -93,30 +93,16 @@ class CutSampler(Sampler[List[str]]):
         self.num_batches = None
 
     def _maybe_init_distributed(self, world_size: Optional[int], rank: Optional[int]):
-        try:
-            # We will ask PyTorch about distributed training metadata,
-            # and it might blow in our faces.
-            if world_size is None:
-                if not dist.is_available():
-                    raise RuntimeError("Requires distributed package to be available")
-                self.world_size = dist.get_world_size()
-            else:
-                self.world_size = world_size
-            if rank is None:
-                if not dist.is_available():
-                    raise RuntimeError("Requires distributed package to be available")
-                self.rank = dist.get_rank()
-            else:
-                self.rank = rank
-        except AssertionError as e:
-            if 'process group is not initialized' in str(e):
-                # Distributed training not active OR somebody forgot to initialize the process group
-                # (which will come up anyway at some point, so we can ignore it here).
-                self.world_size = 1
-                self.rank = 0
-            else:
-                # A different error - pass it on.
-                raise
+        if not dist.is_available() or not dist.is_initialized():
+            self.world_size = 1
+            self.rank = 0
+            return
+
+        if world_size is None:
+            self.world_size = dist.get_world_size()
+
+        if rank is None:
+            self.rank = dist.get_rank()
 
     def set_epoch(self, epoch: int) -> None:
         r"""
--- a/lhotse/dataset/sampling.py
+++ b/lhotse/dataset/sampling.py
@@ -10,6 +10,23 @@ from torch.utils.data import Sampler
 from lhotse import CutSet
 
 
+class DataSource:
+    def __init__(self, items):
+        self._items = items
+        self._permutation = list(range(len(self._items)))
+
+    def shuffle(self, seed):
+        r = random.Random(seed)
+        self._permutation = list(range(len(self._items)))
+        r.shuffle(self._permutation)
+
+    def __getitem__(self, idx):
+        return self._items[self._permutation[idx]]
+
+    def __len__(self):
+        return len(self._items)
+
+
 class CutSampler(Sampler[List[str]]):
     """
     CutSampler is responsible for collecting batches of cuts, given specified criteria.
@@ -67,8 +84,13 @@ class CutSampler(Sampler[List[str]]):
         self.shuffle = shuffle
         self.seed = seed
         self.epoch = 0
+
         self._maybe_init_distributed(world_size=world_size, rank=rank)
-        self.data_source = partition_cut_ids(self.full_data_source, world_size=self.world_size, rank=self.rank)
+        self.data_source = DataSource(
+            partition_cut_ids(self.full_data_source, world_size=self.world_size, rank=self.rank)
+        )
+
+        self.num_batches = None
 
     def _maybe_init_distributed(self, world_size: Optional[int], rank: Optional[int]):
         try:
@@ -105,19 +127,21 @@ class CutSampler(Sampler[List[str]]):
         :param epoch: Epoch number.
         """
         self.epoch = epoch
-
-    def __len__(self) -> int:
-        return len(self.data_source)
+        self.num_batches = None
 
     def __iter__(self) -> 'CutSampler':
         """
         Prepare the dataset for iterating over a new epoch. Will shuffle the data if requested.
         """
         if self.shuffle:
-            r = random.Random(self.seed + self.epoch)
-            r.shuffle(self.data_source)
+            self.data_source.shuffle(self.seed + self.epoch)
         self.current_idx = 0
         return self
+
+    def __len__(self) -> int:
+        if self.num_batches is None:
+            self.num_batches = sum(1 for item in self)
+        return self.num_batches
 
     def __next__(self) -> List[str]:
         raise NotImplemented
@@ -325,6 +349,7 @@ class BucketingSampler(CutSampler):
             *cuts: CutSet,
             sampler_type: Type = SingleCutSampler,
             num_buckets: int = 10,
+            seed: int = 0,
             **kwargs
     ):
         """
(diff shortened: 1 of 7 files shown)


lhotse/dataset/sampling.py
```diff
@@ -96,6 +98,9 @@ def _maybe_init_distributed(self, world_size: Optional[int], rank: Optional[int]
                 # A different error - pass it on.
                 raise
 
+    def _set_sampler_len(self):
+        self._len = sum(1 for item in self)
```
pzelasko, line 102: Unfortunately, I don't think this works as expected. If you look at `__iter__`, it is going to shuffle the cuts every time it's called, so after this loop, the order of the cuts will be different.

Also this is not unset anywhere, so it is going to be incorrect in subsequent epochs.

test/dataset/test_sampling.py
```diff
@@ -40,6 +40,11 @@ def test_single_cut_sampler_shuffling():
     # Invariant 3: the items are shuffled, i.e. the order is different than that in the CutSet
     assert sampler_cut_ids != [c.id for c in cut_set]
 
+    # check that len(sampler) gives consistent result
```
pzelasko, line 43: I think this test's inputs are insufficient to test this, as they all have equal lengths

janvainer: But that should be the point no? The lengths should equal

lhotse/dataset/sampling.py
janvainer: Hm so the shuffling should not happen in-place I guess?

test/dataset/test_sampling.py
pzelasko: sorry I didn't express myself well... I meant that the cut inputs have equal lengths, so whatever is their order inside the sampler, you will always get the same len(sampler) with your changes, regardless of the epoch. If we create cut inputs with varying duration, then you can actually test whether len(sampler) is consistent in the same epoch and different in different epochs

lhotse/dataset/sampling.py
janvainer: @pzelasko is it necessary that the expression `list(sampler)` gives correct result in the case of samplers? The `list()` constructor behaves different from eg `[item for item in sampler]`. The former seems to call `sampler.__len__` asynchronously many times while the latter simply calls `__iter__` and `__next__` in sequential order as is expected. 
Under the current state of code, I am getting `list(sampler) == []` as can be seen from the test. Any idea how this could be surpassed?

pzelasko: one option is to shuffle in a deterministic way, so that the same seed + epoch will always result in the same shuffling order (this actually makes sense and makes it possible to re-start training from a checkpoint mid-epoch in an exact way, as we can know how many batches to skip and from which batch to resume)

test/dataset/test_sampling.py
janvainer: Yup I figured in the end :) The updated version uses varying lengths.

lhotse/dataset/sampling.py
pzelasko: Oh, I did not know that `list(x)` behaves this way. Anyway with fully deterministic shuffling that shouldn't be an issue anymore I think.

janvainer: The shuffling you describe is basically how things work now I think. The iter method uses the self.seed + epoch random seed for shuffling so the process should be deterministic. I tried to determine the length already in init, but still getting this weird behavior.

pzelasko: I don't think that's true, because we re-shuffle the list multiple times, so the elements are permuted in the same manner (because of the seed + epoch), but their order does change (just deterministically).

janvainer: Sorry I dont follow. Maybe it i would understand if i see some code. How the current `__iter__` should change? Right now it looks like this
```
    def __iter__(self) -> 'CutSampler':
        """
        Prepare the dataset for iterating over a new epoch. Will shuffle the data if requested.
        """
        if self.shuffle:
            r = random.Random(self.seed + self.epoch)
            r.shuffle(self.data_source)
        self.current_idx = 0
        self._len = None
        return self

```

pzelasko: OK take a look at the following example:
```python
>>> x = list(range(10))
>>> x
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> r = Random(42)
>>> r.shuffle(x)
>>> x
[7, 3, 2, 8, 5, 6, 9, 4, 0, 1]
>>> r = Random(42)
>>> r.shuffle(x)
>>> x
[4, 8, 2, 0, 6, 9, 1, 5, 7, 3]
```

If you call iter multiple time, the same happens with cut ids (even though the seed + epoch are constant). We might need to hold the copy of the original cut ID ordering and re-compute the shuffling each time iter is called using the original order.

janvainer: Ok I updated the code to remember the original ordering. A list of indices is remembered and shuffled instead of the cut_ids. The random permutation does not change unless the user specifies a different seed. Also the length gets calculated correctly (number of sampled batches == len(sampler)) and the `list()` constructor works too. Is this what you had on mind?

pzelasko: Great! Yes that sounds right. I will review later or tomorrow. Thanks!

lhotse/dataset/sampling.py
```diff
@@ -67,8 +84,13 @@ def __init__(
...
         self.seed = seed
         self.epoch = 0
+
         self._maybe_init_distributed(world_size=world_size, rank=rank)
-        self.data_source = partition_cut_ids(self.full_data_source, world_size=self.world_size, rank=self.rank)
+        self.data_source = DataSource(
+            partition_cut_ids(self.full_data_source, world_size=self.world_size, rank=self.rank)
+        )
+
+        self._len = None
```
pzelasko, line 93: I suggest renaming this to `num_batches` everywhere, it seems more intuitive to set in the sub-classes (like `BucketingSampler`)

test/dataset/test_sampling.py
```diff
@@ -136,6 +175,26 @@ def test_bucketing_sampler_cut_pairs():
     assert set(cut_set2.ids) == set(cut_ids)
 
 
+def test_bucketing_sampler_len():
+    # total duration is 55 seconds
```
pzelasko, line 179: not true for this test

test/dataset/test_sampling.py
```diff
@@ -41,6 +41,25 @@ def test_single_cut_sampler_shuffling():
     assert sampler_cut_ids != [c.id for c in cut_set]
 
 
+def test_single_cut_sampler_len():
```
pzelasko, line 44: can these tests check that the batch ordering is different for every epoch too? you can just store the "batches" as lists and assert that they are not equal (should work since these are just collections of string ids, not dicts with actual data)

pzelasko: (this is mostly to avoid regressions in the future, I can see that the current code should work as expected)

pzelasko: I think it would be good to also have a test that shows that the order is deterministic given the same epoch number. This is not an obvious property in general and I wouldn't be surprised if we accidentally break it trying to update this code in the future.

pzelasko (commented):
Thanks, it's great! Just one cosmetic comment and some notes about testing.

test/dataset/test_sampling.py
janvainer: good catch, forgot to adjust

lhotse/dataset/sampling.py
janvainer: renaming

test/dataset/test_sampling.py
janvainer: I added several tests for ordering and determinism. The sampling of buckets in bucket sampler is now also deterministic and dependent on seed and epoch. I also added seed parameter for the bucket sampler to be able to control the bucket sampling itself.

One more idea: We could add a parameter that would automatically increase epoch after iteration so that user does not have to manually call `set_epoch`. WYT?

janvainer: Hm and one more problem - the `list(sampler)` works fine in `python3.8` but not in `python3.7` for some reason. I don't understand what exactly is wrong.

pzelasko: > One more idea: We could add a parameter that would automatically increase epoch after iteration so that user does not have to manually call `set_epoch`. WYT?

I'd rather not - that would be inconsistent with what PyTorch's DistributedSampler does.

> Hm and one more problem - the list(sampler) works fine in python3.8 but not in python3.7 for some reason. I don't understand what exactly is wrong.

I wouldn't worry about that - if this is problematic later we can look into it.

pzelasko:
Be careful with force-pushing - I merged master to check if it's merge-able

pzelasko (approved):

pzelasko:
Thank you, great work!

lhotse/dataset/sampling.py
```diff
@@ -367,6 +380,7 @@ def __iter__(self) -> 'BucketingSampler':
         return self
 
     def __next__(self) -> List[str]:
+        random.seed(self.seed + self.epoch)
```
pzelasko, line 383: I missed this during the review, but it silently breaks the bucketing sampler - it will always choose the same bucket until depletion

Issue #189, opened 2021-01-30
Padding with acoustic context


danpovey:
Piotr, this is an issue for both Lhotse and Snowfall...
Firstly, I don't think zero-padding at all the intermediate layers of a TDNN is a good idea; it forces
the network to deal with data that's nothing like the normal data, which would probably waste a lot of
capacity on edge cases.

What would be most ideal, IMO, is to have the network reduce the context (i.e. to have left- and right-context, like
networks do in Kaldi) and to give it the actual acoustic context from the original recordings, padded somehow
in case we were at the edge of the original recordings.  Obviously this is going to get a little complicated to implement, but I think it's the right thing to do, and having the ability to deal with context is likely going to be useful in other scenarios as well, e.g. if we do signal processing.

danpovey:
... I assume the way this would interact with feature extraction is, we'd specify the maximum left and right context when the features are dumped, and it would be the user's responsibility to not exceed that when creating the model.
I'm not saying that we shouldn't use padding at all... e.g. we could use no padding in the lower layers, but if there are layers near the top of the network with large stride, that would increase the context too much, maybe use padding there.

pzelasko:
I was also contemplating that and it makes sense to me. For corpora that have the "real" context available (like SWBD or AMI) we can build the cuts to use that; for others that do not have it (like LibriSpeech) we can add a bit of padding (whether noise or "zero" energy) at the beginning and ending of the cut, just enough to avoid padding in all layers of the network (except the input). I might not do it right away (there's a lot of stuff on my plate recently) but as soon as I find a bit more time, I will adjust the Snowfall libri recipe like this, if that course of action sounds OK to you.

danpovey:
Sounds good!

PR #216, opened 2021-03-02
More padding configurations - left/right, duration/num_frames/num_samples

pzelasko:
@danpovey this is going to enable adding a specific number of padding frames in left and right context for ASR training (#189).

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -2157,14 +2157,21 @@ def pad(
     if num_frames is not None:
         assert cut.has_features, 'Cannot pad a cut using num_frames when it is missing pre-computed features ' \
                                  '(did you run cut.compute_and_store_features(...)?).'
-        if num_frames <= cut.num_frames:
-            return cut
         total_num_frames = num_frames
         duration = total_num_frames * cut.frame_shift
         total_num_samples = compute_num_samples(
             duration=duration,
             sampling_rate=cut.sampling_rate
         ) if cut.has_recording else None
+        # It is possible that two cuts have the same number of frames,
+        # but they differ in the number of samples.
+        # In that case, we need to pad them anyway so that they have truly equal durations.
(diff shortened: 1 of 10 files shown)


lhotse/cut.py
```diff
@@ -422,34 +423,38 @@ def truncate(
...
+        or a specific number of samples `num_samples`. The three arguments are mutually exclusive.
 
+        :param cut: Cut to be padded.
         :param duration: The cut's minimal duration after padding.
+        :param num_frames: The cut's total number of frames after padding.
+        :param num_samples: The cut's total number of samples after padding.
         :param pad_feat_value: A float value that's used for padding the features.
             By default we assume a log-energy floor of approx. -23 (1e-10 after exp).
-        :return: a padded MixedCut if ``duration`` is greater than this cut's duration, otherwise ``self``.
+        :param direction: string, 'left' or 'right'. Determines whether the padding is added before or after
```
danpovey, line 446: Perhaps it would be useful to support 'both', or is there a reason why this isn't a good idea?

danpovey: .. 'both' could just recurse to the 'left' and 'right' forms if convenient.

pzelasko: makes sense, will add

lhotse/dataset/transforms/extra_padding.py
```diff
@@ -0,0 +1,40 @@
+from lhotse import CutSet
+from lhotse.utils import LOG_EPSILON
+
+
+class ExtraPadding:
```
pzelasko, line 5: @danpovey we can use this one in snowfall (and when we move on to long-recording corpora we can extend it to find and use the real acoustic context)

pzelasko:
I merged since the tests are OK, I can address any comments in a separate PR.

PR #222, opened 2021-03-08
Fix setting world size and rank in distributed samplers

pzelasko:
Fixes issue introduced in #213

### DIFF
--- a/lhotse/dataset/sampling.py
+++ b/lhotse/dataset/sampling.py
@@ -97,12 +97,8 @@ class CutSampler(Sampler[List[str]]):
             self.world_size = 1
             self.rank = 0
             return
-
-        if world_size is None:
-            self.world_size = dist.get_world_size()
-
-        if rank is None:
-            self.rank = dist.get_rank()
+        self.world_size = dist.get_world_size() if world_size is None else world_size
+        self.rank = dist.get_rank() if rank is None else rank
 
     def set_epoch(self, epoch: int) -> None:
         r"""


danpovey (commented):
LGTM!

danpovey (approved):
forgot to approve

lhotse/dataset/sampling.py
```diff
@@ -97,12 +97,8 @@ def _maybe_init_distributed(self, world_size: Optional[int], rank: Optional[int]
             self.world_size = 1
             self.rank = 0
             return
-
-        if world_size is None:
-            self.world_size = dist.get_world_size()
-
-        if rank is None:
-            self.rank = dist.get_rank()
+        self.world_size = dist.get_world_size() if world_size is None else world_size
```
csukuangfj, line 100: I find that some people is using
```python
self.world_size = world_size or dist.get_world_size()
self.rank = rank or dist.get_rank()
```
which is shorter I think, though the current approach is also OK.

danpovey:
That evaluates it in a Boolean context, which I think is less safe:

```

>>> a = False

>>> b = a or 2

>>> b

2

>>>

```

lhotse/dataset/sampling.py
pzelasko: it looks fancy, but in this context it has a potential for a bug when `rank` is set to 0 but `dist.get_rank()` returns something else

PR #229, opened 2021-03-13
On-the-fly feature extraction support in PyTorch datasets

pzelasko:
This is a feature that many people were interested in during the Kaldi town halls.

To avoid re-writing the datasets for each option of single/multi channel audio/precomputed features/on-the-fly features, I introduced something called `InputStrategy`. It takes a `CutSet` as input and converts it to collated batch of features/audio. It also determines the supervisions start/end in the right domain (samples/frames). 

Currently we only support single-channel, but multi-channel will follow.

I've also added a `PerturbSpeed` transform that might be more convenient for on-the-fly computation (it applies the effect to cuts in a `CutSet` with a given probability).


lhotse/dataset/input_strategies.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    def __call__(self, cuts: CutSet) -> Tuple[torch.Tensor, torch.IntTensor]:
+        raise NotImplementedError()
+
+    def supervision_intervals(self, cuts: CutSet) -> Dict[str, torch.Tensor]:
+        """
+        Returns a dict that specifies the start and end bounds for each supervision,
+        as a 1-D int tensor.
+
```
danpovey, line 26: Would be nice to have more clarity on the string and the tensor in terms of what they represent.  E.g. how many elements does the tensor have, and what is its meaning?

pzelasko:
I made some more changes here to update the most of existing dataset classes and add docs; I think it is mostly ready but I want to try it in Snowfall before I merge. I think I should also describe the `InputStrategy` thing in the general Lhotse docs.

pzelasko:
This is a little bit of a scope creep, but I have also added an option for `Sampler`s to use max duration/samples/frames as the batching criterion (as some cuts might not have frames now) and added a section in the docs about this PR. @danpovey could you take a look if it's clear?

danpovey (approved):
Looks good to me!

PR #230, opened 2021-03-15
Refactor Standardize to GlobalMVN

janvainer:
Compute stats from CutSet, or load stats from file.
Save stats to disk with `torch.save(global_mvn.state_dict())`.

Standardize is refactored into `GlobalMVN`. It is a torch Module now, so it should be easier to use it as the first layer in neural nets.
It can be initialized `from_cuts` or `from_file` and it internally uses `torch.state_dict` and `torch.load`. Since it accepts mean and std as arguments to init, `Module.load_state_dict` cannot be easily used right now, but it would be nice so any suggestions how to handle this are welcome. It would make the module fully compatible with standard neural layers in pytorch.

### DIFF
--- a/lhotse/dataset/feature_transforms.py
+++ b/lhotse/dataset/feature_transforms.py
@@ -1,6 +1,6 @@
 import torch
 
-from typing import Union
+from typing import Union, Optional
 
 from lhotse import CutSet
 from lhotse.utils import Pathlike
@@ -11,12 +11,13 @@ class GlobalMVN(torch.nn.Module):
 
     def __init__(self, feature_dim: int):
         super().__init__()
+        self.feature_dim = feature_dim
         self.register_buffer("norm_means", torch.zeros(feature_dim))
         self.register_buffer("norm_stds", torch.ones(feature_dim))
 
     @classmethod
-    def from_cuts(cls, cuts: CutSet) -> "GlobalMVN":
-        stats = cuts.compute_global_feature_stats()
+    def from_cuts(cls, cuts: CutSet, max_cuts: Optional[int] = None) -> "GlobalMVN":
+        stats = cuts.compute_global_feature_stats(max_cuts)
         stats = {name: torch.as_tensor(value) for name, value in stats.items()}
         feature_dim, = stats["norm_means"].shape
         global_mvn = cls(feature_dim)
--- a/lhotse/dataset/feature_transforms.py
+++ b/lhotse/dataset/feature_transforms.py
@@ -1,15 +1,30 @@
 import torch
 
+from typing import Union
+
 from lhotse import CutSet
+from lhotse.utils import Pathlike
 
 
-class Standardize:
-    def __init__(self, cuts: CutSet):
+class GlobalMVN(torch.nn.Module):
+    """Apply global mean and variance normalization"""
+
+    def __init__(self, norm_means, norm_stds):
+        super().__init__()
+        self.register_buffer("norm_means", torch.as_tensor(norm_means))
+        self.register_buffer("norm_stds", torch.as_tensor(norm_stds))
+
+    @classmethod
+    def from_cuts(cls, cuts: CutSet) -> "GlobalMVN":
         stats = cuts.compute_global_feature_stats()
-        self.norm_means = stats["norm_means"]
-        self.norm_stds = stats["norm_stds"]
+        return cls(stats["norm_means"], stats["norm_stds"])
 
-    def __call__(self, features: torch.Tensor) -> torch.Tensor:
+    @classmethod
+    def from_file(cls, stats_file: Pathlike) -> "GlobalMVN":
+        stats = torch.load(stats_file)
+        return cls(stats["norm_means"], stats["norm_stds"])
+
+    def forward(self, features: torch.Tensor) -> torch.Tensor:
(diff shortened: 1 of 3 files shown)


lhotse/dataset/feature_transforms.py
```diff
@@ -1,15 +1,30 @@
...
 from lhotse import CutSet
+from lhotse.utils import Pathlike
+
 
+class GlobalMVN(torch.nn.Module):
+    """Apply global mean and variance normalization"""
 
-class Standardize:
-    def __init__(self, cuts: CutSet):
+    def __init__(self, norm_means, norm_stds):
```
pzelasko, line 12: I think it's more "pytorch-y" if you give it a dimension arg (`feature_dim: int`); then it should enable using `load_state_dict` (and you can adjust `from_cuts` to use that too);

Can you add `to_file()` too? That will make it convenient to dump and load later.

janvainer: And initialize the buffers to zeros and ones? So something like

```
def __init__(self, feature_dim: int):
    self.register_buffer("norm_means", torch.zeros(feature_dim))
    ...

@classmethod
def from_cuts(cls, cuts):
    # compute stats
    global_mvn = cls(feature_dim_that_we_retrieved_from_cuts)
    global_mvn.norm_means = stats["norm_means"]
    ...
    return global_mvn
```
Yup will add `to_file` too

pzelasko: Looks good to me

janvainer: Please dont merge yet, i will add one more test tomorrow.

lhotse/dataset/feature_transforms.py
```diff
@@ -1,15 +1,40 @@
...
 
+    def __init__(self, feature_dim: int):
+        super().__init__()
+        self.register_buffer("norm_means", torch.zeros(feature_dim))
+        self.register_buffer("norm_stds", torch.ones(feature_dim))
 
-class Standardize:
-    def __init__(self, cuts: CutSet):
+    @classmethod
+    def from_cuts(cls, cuts: CutSet) -> "GlobalMVN":
```
pzelasko, line 18: Please add a parameter `max_cuts`, for large datasets you probably don't need to go through everything to get a reasonable estimate (`compute_global_feature_stats()` supports it already).

test/dataset/test_feature_transforms.py
```diff
@@ -0,0 +1,20 @@
+import tempfile
+import torch
+
+from lhotse import CutSet
+from lhotse.dataset.feature_transforms import GlobalMVN
+
+
+def test_global_mvn_initialization_and_stats_saving():
+    cuts = CutSet.from_json('test/fixtures/ljspeech/cuts.json')
+    tf = tempfile.NamedTemporaryFile()
```
pzelasko, line 10: use `with tempfile.NamedTemporaryFile() as tf:` instead

test/dataset/test_feature_transforms.py
```diff
@@ -0,0 +1,20 @@
+import tempfile
+import torch
+
+from lhotse import CutSet
+from lhotse.dataset.feature_transforms import GlobalMVN
+
+
+def test_global_mvn_initialization_and_stats_saving():
```
pzelasko, line 8: it'd be good to assert the shapes too just to make sure we're using the right object types (helps safe-guard against accidental comma's in returns etc., happened to me more than once in Python)

lhotse/dataset/feature_transforms.py
janvainer: Great! adding

test/dataset/test_feature_transforms.py
janvainer: yes

lhotse/dataset/feature_transforms.py
janvainer: Should be ready now.

test/dataset/test_feature_transforms.py
```diff
@@ -0,0 +1,42 @@
...
+@pytest.mark.parametrize(
+    "in_tensor", [torch.ones(10, 40), torch.ones(2, 10, 40)]
+)
+def test_global_mvn_shapes(global_mvn, in_tensor):
+    assert global_mvn(in_tensor).shape == in_tensor.shape
+    assert global_mvn.inverse(in_tensor).shape == in_tensor.shape
+
+
+@pytest.mark.parametrize(
+    "in_tensor", [torch.ones(10, 40), torch.ones(2, 10, 40)]
```
pzelasko, line 37: I'm not sure I understand why it works with both of these shapes. Could you explain?

janvainer: I guess its torch broadcasting. Try
```
(torch.randn(5) - torch.randn(1, 2, 5)).shape
(torch.randn(5) - torch.randn(2, 5)).shape
```

janvainer: Or should it not work?

pzelasko: OK I'll check it out later. Merging!

janvainer: NP. I tested it now and it returns the same result even when I unsqueeze the smaller tensor.

PR #233, opened 2021-03-16
[experimental] Downloading audio from HTTP/S3/GCP/Azure URLs upon request

pzelasko:
I can only test the HTTP(s) downloading since I don't have access to any S3/GCP/Azure account at this time (hence [experimental] tag) - help in testing from any interested parties will be appreciated!

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -58,14 +58,15 @@ class AudioSource:
                 # TODO(pzelasko): How should we support chunking for commands?
                 #                 We risk being very inefficient when reading many chunks from the same file
                 #                 without some caching scheme, because we'll be re-running commands.
-                raise ValueError("Reading audio chunks from command AudioSource type is currently not supported.")
+                raise ValueError("Reading audio chunks from 'command' AudioSource type is currently not supported.")
             source = BytesIO(run(self.source, shell=True, stdout=PIPE).stdout)
             samples, sampling_rate = read_audio(source)
 
         elif self.type == 'url':
             if offset != 0.0 or duration is not None:
                 # We won't support chunking for URLs, as that could generate a lot of unwanted network traffic.
-                raise ValueError("Reading audio chunks from command AudioSource type is currently not supported.")
+                # We might consider some caching scheme both for this and the command scheme though.
+                raise ValueError("Reading audio chunks from 'url' AudioSource type is currently not supported.")
             try:
                 from smart_open import smart_open
             except ImportError:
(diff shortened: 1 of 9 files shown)


lhotse/audio.py
```diff
@@ -38,37 +38,53 @@ class AudioSource:
...
             source = BytesIO(run(self.source, shell=True, stdout=PIPE).stdout)
+            samples, sampling_rate = read_audio(source)
 
-        with warnings.catch_warnings():
-            warnings.simplefilter('ignore')
-            samples, sampling_rate = read_audio(source, offset=offset_seconds, duration=duration_seconds)
-
-        # explicit sanity check for duration as librosa does not complain here
-        if duration_seconds is not None:
+        elif self.type == 'url':
```
csukuangfj, line 64: The comment at line 33 also needs to be updated.

lhotse/audio.py
```diff
@@ -38,37 +38,53 @@ class AudioSource:
...
+            if offset != 0.0 or duration is not None:
+                # We won't support chunking for URLs, as that could generate a lot of unwanted network traffic.
+                raise ValueError("Reading audio chunks from command AudioSource type is currently not supported.")
```
csukuangfj, line 68: from url AudioSource type?

lhotse/audio.py
pzelasko: Thanks!

pzelasko:
For anybody who stumbles upon it and wants to check it with clouds: I am not sure if `smart_open` requires to be passed a special session object to access cloud storage for publicly available URLs - but it surely will for private clouds. We can add support for setting that, but I'll need somebody to test it.

test/cut/test_cut_set.py
```diff
@@ -193,6 +193,15 @@ def test_trim_to_supervisions_mixed_cuts():
     assert cut.supervisions[0].id == 'sup4'
 
 
+def is_pandas_installed():
```
csukuangfj, line 196: How about using the approach from torchaudio?
https://github.com/pytorch/audio/blob/6bad3a66a7a1c7cc05755e9ee5931b7391d2b94c/torchaudio/_internal/module_utils.py#L9
```python
def is_module_available(*modules: str) -> bool:
    r"""Returns if a top-level module with :attr:`name` exists *without**
    importing it. This is generally safer than try-catch block around a
    `import X`. It avoids third party libraries breaking assumptions of some of
    our tests, e.g., setting multiprocessing start method when imported
    (see librosa/#747, torchvision/#544).
    """
    return all(importlib.util.find_spec(m) is not None for m in modules)
```

It is more general.

pzelasko: Great! Cool find.

PR #237, opened 2021-03-17
Use HDF5 as the default storage backend for features

pzelasko:
(this is the last PR I'm making today)

I think the HDF5 storage backend is pretty stable - never had issues with it in different projects. I remember swapping this with the file-per-array backend in snowfall and observing the effects, I think there was not too much difference with batch loading speed. However, for the feature extraction with multiple parallel jobs, there was a good speedup with HDF5 because of much fewer filesystem-related syscalls (when we create a lot of dirs and files for something like the full LibriSpeech, it becomes the biggest bottleneck).


danpovey (commented):
Looks OK to me!  If this causes a problem I guess we'll discover it pretty soon.

danpovey (approved):

PR #238, opened 2021-03-17
BABEL recipe improvements

pzelasko:


pzelasko:
This seems to work for all BABEL languages now, I might extend it going forward but merging this for now.

PR #241, opened 2021-03-20
Fix CutMix not adding enough noise in long cuts

### LINKED ISSUE #236: CutMix might not be covering the whole sample with additive noise
Found in #234

pzelasko:
Fixes #236 

Now everything seems to work as intended; @danpovey I expect today's bugfixes might give a little improvement to snowfall recipes speed and WER (especially if we use duration_factor=2, I think the default is 1)

image

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -238,15 +238,16 @@ class Recording:
         else:
             channels = frozenset(channels)
 
+        offset_sp, duration_sp = self._adjust_for_speed_perturbation(offset, duration)
+
         samples_per_source = []
         for source in self.sources:
             # Case: source not requested
             if not channels.intersection(source.channels):
                 continue
-            offset, duration = self._determine_offset_and_duration(offset, duration)
             samples = source.load_audio(
-                offset=offset,
-                duration=duration,
+                offset=offset_sp,
+                duration=duration_sp,
             )
 
             # Case: two-channel audio file but only one channel requested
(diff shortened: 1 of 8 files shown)


danpovey:
Thanks!
Do I understand correctly from the picture that it is putting multiple supervision objects in a single sequence?
I believe the attention masking code currently in Lhotse does not deal with this correctly.
Dan

pzelasko:
Yeah it does concatenate the cuts (with supervisions) here. I’ll double check but could you be more specific which code do you mean? For K2 I believe we are not providing masks, just start frame and num frames for each supervision (drawn as red boxes here). In the latter box it mixes in MUSAN which is why there is speech outside of the box.

danpovey:
The issue is not k2 but the function 'encoder_padding_mask' in transformer.py.
It doesn't seem to deal correctly with having multiple cuts in the same sequence.

In fact, it's not clear whether we should really be using this mask in cases where the data outside the cuts has augmentations.  I am experimenting with turning off the mask some percentage of the time.

pzelasko:
Oh, got it. I think it makes sense to turn that mask off entirely with what we’re doing here. It will make even more sense with conversational speech where multiple supervisions will come from the same recording.

danpovey:
When training the model from scratch with the configuration in master, I'm getting WER 6.96% / 19.69% with test-clean/test-other, which is about right.
Now I am trying a run where, for training, I only use the mask half the time.
Something odd is, right now I'm not getting the valid_objf I was getting before; it's 0.23 instead of 0.15, while objf is still 0.10.

danpovey:
.. Incidentally, when I remove the mask for decoding, from a model that was trained with the mask, WER only degrades from 6.96% to 7.25%  and from 19.69% to 20.75%.

pzelasko:
@danpovey the dev loss increase probably comes from the fact that I add `dev-other` to the dev set.

pzelasko:
FYI I trained a model where I removed the encoder padding mask entirely (I made that function return `None`), it gave me
```
2021-03-23 15:55:29,939 INFO [mmi_att_transformer_decode.py:329] [test-clean] %WER 6.95% [3656 / 52576, 497 ins, 316 del, 2843 sub ]
2021-03-23 15:56:13,950 INFO [mmi_att_transformer_decode.py:329] [test-other] %WER 19.31% [10107 / 52343, 1220 ins, 1001 del, 7886 sub ]
```
it seems to have either very mild or no effect at all.

PR #246, opened 2021-03-24
SpecAugment implementation

pzelasko:
It has some issues for now; it is not padding-aware atm, so the time masks and warps can happen in the padded areas. It's probably not a big issue for how we're training ASR in K2, but it might be unexpected for other users, so I might have to fix it.

# Examples:

## Regular batches (p=1.0)
image

## Regular batches (p=0.5)
image

## With cut concatenation (p=1.0)
image


danpovey:
Cool!
I don't see a reason why it needs to be padding-aware...

pzelasko:
Actually I guess you're right. If somebody doesn't like it they can use the BucketingSampler and the issue will be gone anyway.

pzelasko:
@danpovey FYI I am getting slightly better results with this, test-clean is not affected, but test-other seems to improve a bit (I think the baseline number is around 19.2% - 19.3%)

```
2021-03-25 21:52:23,645 INFO [mmi_att_transformer_decode.py:329] [test-clean] %WER 6.93% [3645 / 52576, 529 ins, 308 del, 2808 sub ]
2021-03-25 21:53:10,674 INFO [mmi_att_transformer_decode.py:329] [test-other] %WER 18.53% [9697 / 52343, 1136 ins, 929 del, 7632 sub ]
```

I'll try a bunch of configurations and commit the best I find to snowfall.

danpovey:
Great, yes!  I haven't seen a number that good on test-other.

xiaohui-zhang:
@pzelasko do you remember whether it makes a difference regarding masking each utterance w/ per-utterance mean or a global mean computed from the batch? I guess you found it matters and then made the choice of using per-utt mean sacrificing some parallelism (espnet and speechbrain use global mean)

pzelasko:
I don’t think that I experimented with that. Per-utterance mean made sense to me at the time since it would always give 0 after mean normalization.

xiaohui-zhang:
cool thx @pzelasko

PR #252, opened 2021-03-26
Add LIbrosaFbank consistent with various TTS applications

janvainer:
Add librosa feature extractor. Not sure if the librosa import is ok this way?

### DIFF
--- a/lhotse/features/librosa_fbank.py
+++ b/lhotse/features/librosa_fbank.py
@@ -1,17 +1,8 @@
 from dataclasses import asdict, dataclass
 
 import numpy as np
-from lhotse import FeatureExtractor
-from lhotse.features.base import register_extractor
-from lhotse.utils import Seconds, compute_num_frames, is_module_available
-
-
-if is_module_available("librosa"):
-    import librosa
-else:
-    raise ImportError(
-        "Librosa is not installed. Please install librosa before using LibrosaFbank extractor."
-    )
+from lhotse.features.base import register_extractor, FeatureExtractor
+from lhotse.utils import Seconds, compute_num_frames, is_module_available, LOG_EPSILON
 
 
 @dataclass
@@ -29,6 +20,26 @@ class LibrosaFbankConfig:
     num_mel_bins: int = 80
     fmin: int = 80
     fmax: int = 7600
+
+
+def pad_or_truncate_features(
+        feats,
+        expected_num_frames: int,
+        abs_tol: int = 1,
+        pad_value: float = LOG_EPSILON
+):
+    frames_diff = feats.shape[0] - expected_num_frames
+
+    if 0 < frames_diff <= abs_tol:
+        feats = feats[:expected_num_frames]
+    elif -abs_tol <= frames_diff < 0:
+        feats = np.pad(feats, ((0, -frames_diff), (0, 0)), mode="constant", constant_values=LOG_EPSILON)
+    elif abs(frames_diff) > abs_tol:
+        raise ValueError(
+            f"Expected {expected_num_frames} source_feats; feats.shape[0] = {feats.shape[0]}"
+        )
+
+    return feats
 
 
 def logmelfilterbank(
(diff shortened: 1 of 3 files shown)


lhotse/features/librosa_fbank.py
```diff
@@ -0,0 +1,110 @@
+from dataclasses import asdict, dataclass
+
+import numpy as np
+from lhotse import FeatureExtractor
+from lhotse.features.base import register_extractor
+from lhotse.utils import Seconds, compute_num_frames, is_module_available
+
+
+if is_module_available("librosa"):
```
pzelasko, line 9: I'd put this thing inside `logmelfilterbank` function instead, so that the class can be visible in global `lhotse` namespace without import errors, and only throws the error if somebody uses it without installing librosa

lhotse/features/librosa_fbank.py
```diff
@@ -0,0 +1,110 @@
...
+    mel_basis = librosa.filters.mel(sampling_rate, fft_size, num_mel_bins, fmin, fmax)
+
+    feats = np.log10(np.maximum(eps, np.dot(spc, mel_basis.T)))
+
+    expected_num_frames = compute_num_frames(
+        duration=len(audio) / sampling_rate,
+        frame_shift=hop_size / sampling_rate,
+        sampling_rate=sampling_rate,
+    )
+    if feats.shape[0] > expected_num_frames:
```
pzelasko, line 86: Can we re-write this conditional so that it allows some small absolute difference in frame counts (e.g. +/- 1 or 2 frames), and pads with low energy/truncates to make it compatible with `expected_num_frames`, and raises the error otherwise? Currently I think it can let some real errors pass through.

lhotse/features/librosa_fbank.py
```diff
@@ -0,0 +1,110 @@
...
+    elif feats.shape[0] < expected_num_frames:
+        raise ValueError(
+            f"Expected {expected_num_frames} source_feats; feats.shape[0] = {feats.shape[0]}"
+        )
+
+    return feats
+
+
+@register_extractor
+class LibrosaFbank(FeatureExtractor):
```
pzelasko, line 97: I think you could copy `mix` and `compute_energy` from the `Fbank` class, this way it will enable feature-domain mixing for this feature extractor.

janvainer: Done

lhotse/features/librosa_fbank.py
janvainer: Done + added test

lhotse/features/librosa_fbank.py
janvainer: Done

lhotse/features/librosa_fbank.py
pzelasko: Can we have some description of this class and how is it different from `Fbank`?

test/features/test_librosa_fbank.py
```diff
@@ -0,0 +1,34 @@
+import pytest
```
pzelasko, line 1: I wonder whether we should also have a test that computes the features using `LibrosaFbank` for a number of different lengths of a waveform.

janvainer: OK, The test checks that the length of the feature follows the general formula for convolution output size with +-1 frame tolerance.

lhotse/features/librosa_fbank.py
janvainer: I added short description. Not sure what else to add. Any ideas?

test/features/test_librosa_fbank.py
pzelasko: I meant a test that actually runs an object LibrosaFbank class. Just to make sure we are able to detect whenever something gets broken inside due to somebody modifying it. Right now you can see codecov complaining that the code is never run.

janvainer: I am not sure what you mean tbh. Do you want to initialize the class, but not call execute? The test can not be executed unless we install librosa in the CI. I see that the codecov is failing, but not sure what should be changed.

pzelasko: I'm gonna merge for now since it is unlikely to break anything. 

I don't think I explained myself well. I meant that we should probably install librosa as part of CI and run the code that imports it (make the test conditional on its availability, see `pytest.mark.skipif`). There are tests of others feature extractor classes and `LibrosaFbank` should be included in them. But it is not urgent and we can add these later.

pzelasko: If you have more time to spend on this then feel free to create a follow-up PR.

PR #259, opened 2021-04-09
add multi thread to process AIShell data

pingfengluo:
add multi thread to process AIShell data and try to make it faster

### DIFF
--- a/lhotse/recipes/aishell.py
+++ b/lhotse/recipes/aishell.py
@@ -70,7 +70,7 @@ def prepare_aishell(
         output_dir = Path(output_dir)
         output_dir.mkdir(parents=True, exist_ok=True)
     transcript_path = corpus_dir / 'data_aishell/transcript/aishell_transcript_v0.8.txt'
-    transcript_dict = {}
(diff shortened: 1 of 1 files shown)


lhotse/recipes/aishell.py
```diff
@@ -74,46 +77,47 @@ def prepare_aishell(
...
-                recording_id=idx,
-                start=0.0,
-                duration=recording.duration,
-                channel=0,
-                language='Chinese',
-                speaker=speaker,
-                text=text.strip()
-            )
-            supervisions.append(segment)
+    with ThreadPoolExecutor(num_jobs) as ex:
```
pzelasko, line 80: The `ex` object is never used here, so nothing actually gets processed in parallel. You’d have to use either `ex.submit` or `ex.map` to leverage the thread pool (I think there’s an example in the librispeech recipe).

pingfengluo: ok. thank you.

pzelasko:
Did you test it after the changes? Let me know and I'll merge it.

PR #262, opened 2021-04-10
Add JSONL support

pzelasko:
This could be a bit more efficient than reading large JSONs; but I haven't measured the speed difference yet. I have some ideas for speeding-up the reading of larger (5 digit number of hours) datasets, will try them out eventually.

@jimbozhang this could be helpful in your current efforts.

### DIFF
--- a/docs/corpus.rst
+++ b/docs/corpus.rst
@@ -1,12 +1,12 @@
 Representing a corpus
 =====================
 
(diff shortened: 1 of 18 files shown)


jimbozhang:
Yes, I think JSONL should be faster for large dataset. Maybe we need to write a converting tool (JSON to JSONL) and put it to https://github.com/SpeechColab/GigaSpeech/tree/main/utils.

PR #265, opened 2021-04-11
Fix find segments float rounding issues

### LINKED ISSUE #263: Loading CutSet from manifest may drop supervisions
Calling `CutSet.from_manifests()` on pre-computed manifests may result in dopping supervisions. 
Steps to reproduce:

```
ROOT = "."
MANIFESTS = $(ROOT)/LJSpeech-1.1/lhotse_manifests
FEATS = $(ROOT)/LJSpeech-1.1/feats

lhotse obtain ljspeech $(ROOT)
lhotse prepare ljspeech $(ROOT)/LJSpeech-1.1/ $(MANIFESTS)

# Extract LibrosaFbank features
lhotse feat write-default-config $(MANIFESTS)/librosa_config.yaml -f librosa-fbank
lhotse feat extract \
$(MANIFESTS)/recordings.json \
$(FEATS) \
-f $(MANIFESTS)/librosa_config.yaml

# store cuts
lhotse cut simple $(MANIFESTS)/cuts.json \
-r $(MANIFESTS)/recordings.json \
-f $(FEATS)/feature_manifest.json.gz \
-s $(MANIFESTS)/supervisions.json
```
The stored cuts don't have supervisions. The problem is that the ends of the feature segments and ends of the supervision segments do not match, because Supervision.end [rounds to 8 digits](https://github.com/lhotse-speech/lhotse/blob/61177e48cbc2e53f83541ed8b301d4d4ab1b54da/lhotse/supervision.py#L27), but the Features.end does not.

The SupervisionSegment.find then does not match features with supervisions correctly in [SupervisionSet.find()](https://github.com/lhotse-speech/lhotse/blob/61177e48cbc2e53f83541ed8b301d4d4ab1b54da/lhotse/cut.py#L1286). 

I tried to disable the rounding, but it triggers the `test/cut/test_cut_truncate.py` test. Quick fix would be to round digits inside the find function or use some tolerance, but I am not sure if this would not break something else. WYT @pzelasko ?

pzelasko:
Resolves #263 

@janvainer

### DIFF
--- a/lhotse/supervision.py
+++ b/lhotse/supervision.py
@@ -194,7 +194,8 @@ class SupervisionSet(Serializable, Sequence[SupervisionSegment]):
             channel: Optional[int] = None,
             start_after: Seconds = 0,
             end_before: Optional[Seconds] = None,
-            adjust_offset: bool = False
+            adjust_offset: bool = False,
+            tolerance: Seconds = 0.001
     ) -> Iterable[SupervisionSegment]:
         """
         Return an iterable of segments that match the provided ``recording_id``.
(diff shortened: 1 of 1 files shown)


janvainer:
Cool, thanks!

PR #268, opened 2021-04-13
Update the examples of libirispeech (full) and ami

jimbozhang:


examples/librispeech/librispeech.ipynb
```diff
@@ -119,20 +104,16 @@
...
-    "# The issue seems to happen only in a Jupyter notebook on Mac OS X, hence the work around below.\n",
-    "if use_data_augmentation:\n",
-    "    num_jobs = 1\n",
-    "else:\n",
-    "    num_jobs = os.cpu_count()\n",
-    "    torch.set_num_threads(1)\n",
-    "    torch.set_num_interop_threads(1)"
+    "augment_fn = SoxEffectTransform(effects=[\n",
+    "   ['reverb', 50, 50, RandomValue(0, 100)],\n",
+    "   ['speed', RandomValue(0.9, 1.1)],\n",
```
pzelasko, line 114: I strongly advise not to use `speed` + `rate` this way because it change the audio array length, but doesn't update the supervisions to match it. The preferred way is to use `cut_set.perturb_speed(0.9)`, and if you want to combine raw and augmented versions, `cut_set = cut_set + cut_set.perturb_speed(0.9) + cut_set.perturb_speed(1.1)`

examples/librispeech/librispeech.ipynb
```diff
@@ -144,23 +125,28 @@
...
+    "            cut_set = CutSet.from_manifests(\n",
+    "                recordings=manifests['recordings'],\n",
+    "                supervisions=manifests['supervisions']\n",
+    "            ).compute_and_store_features(\n",
+    "                extractor=Fbank(),\n",
+    "                storage_path=f'{output_dir}/feats_{partition}',\n",
+    "                storage_type=LilcomFilesWriter,\n",
+    "                augment_fn=augment_fn,\n",
+    "                executor=ex\n",
+    "            ).pad()\n",
```
pzelasko, line 147: `.pad()` is no longer needed here, dataset classes should take care of it

examples/librispeech/librispeech.ipynb
```diff
@@ -191,6 +187,19 @@
    "cell_type": "code",
    "execution_count": 10,
    "metadata": {},
+   "outputs": [],
+   "source": [
+    "sampler = SingleCutSampler(cutset_dev, shuffle=False, max_samples=16)\n",
```
pzelasko, line 192: `max_samples=16` is a limit on the audio samples, which is extremely low; I think we might need to find a better name to avoid confusion of "audio samples" with "samples in a batch"...

pzelasko: (you probably wanted to specify `max_cuts=16`)

pzelasko: but actually, it's more advisable to specify e.g. `max_duration=100.0` (seconds)

examples/librispeech/librispeech.ipynb
```diff
@@ -231,31 +240,30 @@
...
   {
    "cell_type": "code",
-   "execution_count": 17,
+   "execution_count": 12,
    "metadata": {},
    "outputs": [],
    "source": [
-    "train_dl = torch.utils.data.DataLoader(cuts_train, batch_size=32, shuffle=True)\n",
-    "dev_dl = torch.utils.data.DataLoader(cuts_dev, batch_size=32)"
+    "train_dl = torch.utils.data.DataLoader(dataset_train, batch_size=32, shuffle=True)\n",
```
pzelasko, line 247: `batch_size` has to be `None`

pzelasko: also the train sampler is not specified

examples/librispeech/mini_librispeech.ipynb
```diff
@@ -229,7 +229,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "train_dl = torch.utils.data.DataLoader(dataset_train, batch_size=None, shuffle=True)\n",
+    "train_dl = torch.utils.data.DataLoader(dataset_train, batch_size=32, shuffle=True)\n",
```
pzelasko, line 232: same comments as in the full librispeech recipe

examples/librispeech/librispeech.ipynb
csukuangfj, line 247: I think there is no point in building a dataloader from a `K2SpeechRecognitionDataset` without specifying a sampler.

The resulting dataloader is useless since `__getitem__` of `K2SpeechRecognitionDataset`
accepts a list of `str`, not a single `int`.

pzelasko:
Thanks!

Issue #43, opened 2020-07-02
Reading SPH files directly from Python


pzelasko:
Current mechanism is very similar to how Kaldi does it - we can leverage external shell tools to pipe the result and read it in Lhotse. 

To do that, one has to create a RecordingSet (recording manifest), where each Recording has an AudioSource of type ‘command’. Then, put something like `sph2pipe file.sph -` in there (the pipe symbol at the end is not needed), and when “load_audio()” is called somewhere in Lhotse, it will take care of spawning a subprocess and reading the data from it. 

We could support for reading SPH files directly in Python with https://pypi.org/project/sphfile/ (sounds like a good first issue)

PR #269, opened 2021-04-13
Fix test for sphere files.

csukuangfj:
As the test shows, `soundfile` can handle `.sph` files correctly.

Should we remove the dependency on `sphfile` from `requirements.txt`? (It was introduced in https://github.com/lhotse-speech/lhotse/issues/43)

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -30,7 +30,7 @@ class AudioSource:
     """
     AudioSource represents audio data that can be retrieved from somewhere.
     Supported sources of audio are currently:
-    - 'file' (formats supported by librosa, possibly multi-channel)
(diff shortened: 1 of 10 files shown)


pzelasko:
Thanks! Yes, we can remove the `sphfile` dependency; we should also remove `Recording.from_sphere()` (and actually `Recording.from_wav()` too since it's deprecated). We'd need to scan the codebase for usage of these methods and replace them with `from_file()`. Since you opened the PR would you like to do it?

csukuangfj:
@pzelasko 

OK. `from_sphere` is replaced by `from_file`.

`from_wav` turns out to be never used.

pzelasko:
Cool, thanks!

PR #272, opened 2021-04-14
Randomized smoothing augmentation

pzelasko:
image


danpovey:
Surely you would want to do this with some small probability rather than all the time?
Otherwise the nnet wouldn't be able to deal with non-noisy  data..

pzelasko:
In the default mode it’s sampling the magnitude of the noise so it won’t always be as profound as in the picture but it’s a good point, I’ll add a probability.

Issue #271, opened 2021-04-13
Features after SpecAugment go outside of supervisions


pzelasko:
Example:

![image](https://user-images.githubusercontent.com/15930688/114607995-320bcd00-9c6b-11eb-87c6-73744ef922db.png)

This is due to the "time warping" component of SpecAugment. I think we'll need to apply it before the padding instead. Interestingly, it improves the results on the 100h subset of LibriSpeech nevertheless.

danpovey:
Cool!

PR #275, opened 2021-04-16
Add support for supervision segments in SpecAugment

pzelasko:
In this PR, SpecAugment has an additional input of supervision segments in the same format as K2's DenseFsaVec. This solves the issue #271 where spec-augmented features go out of supervision segments, by applying spec augment to slices of the feature matrices. 

I think the transformed batches are very interesting:

image

I'll see how/if it affects ASR training.

### DIFF
--- a/lhotse/dataset/input_strategies.py
+++ b/lhotse/dataset/input_strategies.py
@@ -34,8 +34,9 @@ class InputStrategy:
         .. code-block:
 
             {
+                "sequence_idx": tensor(shape=(S,)),
                 "start_frame": tensor(shape=(S,)),
-                "num_frames": tensor(shape=(S,))
+                "num_frames": tensor(shape=(S,)),
             }
 
         or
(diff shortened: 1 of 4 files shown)


danpovey:
Why is it necessary to avoid the zeroed segments crossing supervision boundaries?

pzelasko:
It is not, but it would have further complicated the code. It’s the time stretching that needs to be done within supervisions, otherwise the features can be stretched outside their bounding boxes.

pzelasko:
Alright, I modified it to warp in supervisions and mask everywhere. Note that masks may be applied to different examples than the time-warped ones. It is really not defined how exactly we should do it since the original paper did not consider concatenating cuts ;)  

Looks like this now:

image

pzelasko:
It didn't affect LibriSpeech results in a significant way (4.99/10.21 [new] for 5.02/10.16 [old]) for averaged model after 6 epochs, but it seems technically correct to do it this way so I'll merge it.

danpovey:
Great!

pzelasko:
BTW I realized why it worked before -- since our models see all the frames outside of supervision segments, they were able to incorporate the frames that went outside of the supervisions into recognition.

PR #279, opened 2021-04-20
fix l2arctic dataset prep when output_dir is not None

oplatek:


lhotse/recipes/l2_arctic.py
```diff
@@ -162,6 +163,8 @@ def prepare_l2_arctic(
     }
 
     if output_dir is not None:
+        output_dir = Path(output_dir)
+        makedirs(output_dir, exist_ok=True)
```
pzelasko, line 167: FYI: Path.mkdir(...)

csukuangfj: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

csukuangfj: Turns out the two are equivalent.

See

### pathlib.mkdir
https://github.com/python/cpython/blob/2a36b09ce7cd67503893412b53295717d66fee19/Lib/pathlib.py#L1221


```python
    def mkdir(self, mode=0o777, parents=False, exist_ok=False):
```

https://github.com/python/cpython/blob/2a36b09ce7cd67503893412b53295717d66fee19/Lib/pathlib.py#L361

```python
    mkdir = os.mkdir
```

### os.makedirs

https://github.com/python/cpython/blob/2a36b09ce7cd67503893412b53295717d66fee19/Lib/os.py#L200

```python
def makedirs(name, mode=0o777, exist_ok=False):
   ...
    try:
        mkdir(name, mode)
```

pzelasko: Yeah. My comment was just about consistency (we typically use `pathlib` which is a bit more high-level than  `os`/`os.path` in Lhotse)

PR #282, opened 2021-04-22
MLS recipe and reading OPUS files with an optional audioread dependency

pzelasko:


pzelasko:
I am merging this but will probably revisit MLS data prep at some point - it works but is too very slow (despite parallelism), it should at least store some intermediate results while processing.

PR #283, opened 2021-04-26
The GigaSpeech recipe

jimbozhang:
This is just a draft for https://github.com/k2-fsa/snowfall/issues/148.

This speechcolab recipe depends on the `PySpeechColab` project, which is designed as a set of common tools for speech tasks. We will use the `speechcolab` package to download the GigaSpeech data in this recipe.

The `PySpeechColab` project is under development.

### DIFF
--- a/lhotse/recipes/gigaspeech.py
+++ b/lhotse/recipes/gigaspeech.py
@@ -1,0 +1,10 @@
+from speechcolab.datasets import gigaspeech
+
+
+def download_gigaspeech(
+        target_dir: Pathlike = '.',
+        dataset_parts: Optional[Union[str, Sequence[str]]] = "XS",
+        force_download: Optional[bool] = False
+) -> None:
+    gigaspeech_data = gigaspeech.GigaSpeech(target_dir)
+    gigaspeech_data.download(dataset_parts, force_download)
--- a/lhotse/recipes/__init__.py
+++ b/lhotse/recipes/__init__.py
@@ -5,6 +5,7 @@ from .broadcast_news import prepare_broadcast_news
 from .callhome_egyptian import prepare_callhome_egyptian
 from .callhome_english import prepare_callhome_english
 from .cmu_arctic import download_cmu_arctic, prepare_cmu_arctic
+from .gigaspeech import download_gigaspeech, prepare_gigaspeech
 from .heroico import prepare_heroico
 from .l2_arctic import prepare_l2_arctic
 from .librimix import prepare_librimix
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ pyyaml>=5.3.1
 torch>=1.7.1
 torchaudio>=0.7.1
 tqdm
+speechcolab


lhotse/recipes/gigaspeech.py
```diff
@@ -0,0 +1,10 @@
+from speechcolab.datasets import gigaspeech
+
+
+def download_gigaspeech(
+        target_dir: Pathlike = '.',
+        dataset_parts: Optional[Union[str, Sequence[str]]] = "XS",
+        force_download: Optional[bool] = False
+) -> None:
```
pzelasko, line 8: Can we handle the `speechcolab` dependency like the other optional deps, see e.g. the import of `librosa` here:

https://github.com/lhotse-speech/lhotse/blob/274cf876ee1e31ec8a837cbd5dc3b6f6bfdf2b79/lhotse/features/librosa_fbank.py#L85

The point is to try to import, if it fails then raise an exception and say "you can fix this with `pip install X`". Then we can remove the dep from `requirements.txt`.

lhotse/recipes/gigaspeech.py
```diff
@@ -0,0 +1,14 @@
...
+from lhotse.utils import Pathlike, is_module_available
+
+
+def download_gigaspeech(
+        target_dir: Pathlike = '.',
+        dataset_parts: Optional[Union[str, Sequence[str]]] = "XS",
+        force_download: Optional[bool] = False
+) -> None:
+    if is_module_available('speechcolab'):
+        from speechcolab.datasets import gigaspeech
```
pzelasko, line 11: I think the error message will be quite cryptic in this case (undefined symbol `gigaspeech`) -- how about:

```python
if not is_module_available('speechcolab'):
    raise ImportError('To download GigaSpeech, please install optional dependency: pip install speechcolab')
```

jimbozhang:
I just committed a draft implement of `lhotse.recipes.gigaspeech` in https://github.com/lhotse-speech/lhotse/pull/283/commits/78dc60538e97ec7254ee302313450aed1eae8244
Could you please take a look at it and give some suggestions? @pzelasko @chenguoguo

lhotse/recipes/gigaspeech.py
```diff
@@ -1,14 +1,37 @@
...
+                      'speechcolab')
+else:
+    from speechcolab.datasets.gigaspeech import GigaSpeech
 
-    gigaspeech_data = gigaspeech.GigaSpeech(target_dir)
-    gigaspeech_data.download(dataset_parts, force_download)
+
+def prepare_gigaspeech(
+        gigaspeech: GigaSpeech,
+        subset: str
```
pzelasko, line 16: how about we add `dataset_parts` like the ones in LibriSpeech with a default of 'auto' or 'all'

https://github.com/lhotse-speech/lhotse/blob/78dc60538e97ec7254ee302313450aed1eae8244/lhotse/recipes/librispeech.py#L69

and return a dict of:

```
{
  "XS": {"recordings": ..., "supervisions": ...},
  "test": {...},
  ...
}
```

lhotse/recipes/gigaspeech.py
```diff
@@ -1,14 +1,37 @@
...
+if not is_module_available('speechcolab'):
+    raise ImportError('To process the GigaSpeech corpus, please install optional dependency: pip install '
+                      'speechcolab')
+else:
+    from speechcolab.datasets.gigaspeech import GigaSpeech
 
-    gigaspeech_data = gigaspeech.GigaSpeech(target_dir)
-    gigaspeech_data.download(dataset_parts, force_download)
+
+def prepare_gigaspeech(
```
pzelasko, line 14: We should also add an `output_dir` arg that behaves like in the other recipes

lhotse/recipes/gigaspeech.py
```diff
@@ -1,14 +1,37 @@
-from typing import Optional, Sequence, Union
```
pzelasko: Would be good to start this file with a docstring description of GigaSpeech.

lhotse/recipes/gigaspeech.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Dict, Optional, Sequence, Union
+
+from lhotse.audio import Recording, RecordingSet
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike, is_module_available
+
+if is_module_available('speechcolab'):
+    from speechcolab.datasets.gigaspeech import GigaSpeech
+else:
+    raise ImportError('To process the GigaSpeech corpus, please install optional dependency: pip install speechcolab')
```
mthrok, line 11: If I read this correctly, this makes `gigaspeech` hard requirement even when I don't intend to use this recipe. This module is imported in `recipes.__init__.py`. so this `ImportError` will be always raised if I don't have `gigaspeech`.

jimbozhang: Thanks for pointing this out. How about I check that module in the `prepare_gigaspeech()` function?

csukuangfj: How about
```python
if is_module_available('speechcolab'):
  import importlib
  speechcolab = importlib.import_module('speechcolab')
  GigaSpeech = speechcolab.datasets.gigaspeech.GigaSpeech
else:
  raise ImportError('.....')
```

pzelasko: Thanks for noticing that. It should be a local import inside prepare_gigaspeech, then import lib is not needed. It works for other recipes and optional functionality.

lhotse/recipes/gigaspeech.py
```diff
@@ -1,41 +1,79 @@
...
+        audio: Any,
+        root_path: Path
+) -> Optional[Tuple[Recording, List[SupervisionSegment]]]:
+    recording = Recording.from_file(root_path / audio['path'], recording_id=audio['aid'])
+    segments = []
+    for seg in audio['segments']:
+        segments.append(SupervisionSegment(id=seg['sid'],
+                                           recording_id=audio['aid'],
+                                           start=Seconds(seg['begin_time']),
+                                           duration=Seconds(seg['end_time'] - seg['begin_time']),
```
pzelasko, line 74: you might want to round the duration with sth like `round(…, ndigits=8)`, otherwise there will be entries like 17.0000000002

lhotse/recipes/gigaspeech.py
```diff
@@ -1,41 +1,79 @@
...
+                supervisions += segments
+
+            manifests[part] = {
+                'recordings': RecordingSet.from_recordings(recordings),
+                'supervisions': SupervisionSet.from_segments(supervisions)
+            }
+
+            if output_dir is not None:
+                manifests[part]['recordings'].to_json(output_dir / f'recordings_{part}.json')
+                manifests[part]['supervisions'].to_json(output_dir / f'supervisions_{part}.json')
```
pzelasko, line 59: can we save those as JSONL instead? You can use `to_file(… / …jsonl)`

jimbozhang:
The feature extraction of this recipe fails on some opus-format audios. It turns out that `SoundFile` can not open such format audios. I extracted the related codes and illustrated them at: 
https://colab.research.google.com/drive/18xr9JBzOp3VfUpYNc0kX0fYbmT3c_t3p

where the audio could be downloaded from:
https://drive.google.com/file/d/1AyfS7MeoClhTrtsAJw2CkVfWb8xoUwNF/view?usp=sharing

@dophist suggested that it should be a bug of `SoundFile`, which has been discussed in: https://github.com/bastibe/python-soundfile/issues/252

I'd like to hear the suggestions from @pzelasko 😐

pzelasko:
I’m aware, for that reason I recently added audioread support specifically to be able to read OPUS. It’s a core dependency now, and everything you can read with librosa you can also read with Lhotse https://github.com/lhotse-speech/lhotse/blob/master/requirements.txt#L1  Are you using the most recent Lhotse?

jimbozhang:
> I’m aware, for that reason I recently added audioread support specifically to be able to read OPUS. It’s a core dependency now, and everything you can read with librosa you can also read with Lhotse https://github.com/lhotse-speech/lhotse/blob/master/requirements.txt#L1 Are you using the most recent Lhotse?

Thanks @pzelasko . I just tried it. For the audios that were read by `SoundFile`, everything goes well. But for the audios that have to be read by `audioread`, say, [POD0000000480.opus](https://drive.google.com/file/d/1AyfS7MeoClhTrtsAJw2CkVfWb8xoUwNF/view?usp=sharing), a `ValueError` exception raises:

```
ValueError: The number of declared samples in the recording diverged from the one obtained when loading audio.
This could be internal Lhotse's error or a faulty transform implementation.
Please report this issue in Lhotse and show the following:
diff=-100616288, audio.shape=(1, 150924432),
recording=Recording(id='POD0000000480',
sources=[AudioSource(type='file', channels=[0], source='data/GigaSpeech/audio/podcast/P0005/POD0000000480.opus')], sampling_rate=16000, num_samples=50308144, duration=3144.259, transforms=None)
```

pzelasko:
Hmm, where do you have `num_samples` from for that recording? It seems like it's loading 3x more audio samples than declared in the manifest.

pzelasko:
... it could be possible that this OPUS recording might have 48kHz sampling rate and not 16kHz (can you double check?)

jimbozhang:
> ... it could be possible that this OPUS recording might have 48kHz sampling rate and not 16kHz (can you double check?)

Ah yes it is 16kHz actually. Thanks for your help 👍

pzelasko:
Cool. I might improve the error message to also check the sampling rate in these cases.

jimbozhang:
Hmm... I have been confused about the sample rate of [that audio](https://drive.google.com/file/d/1AyfS7MeoClhTrtsAJw2CkVfWb8xoUwNF/view?usp=sharing). Some tools (eg. oceanaudio) say its sample rate is 16000, some others (eg. audacity, vlc, soxi) say it's 48000.

I found this page: https://stackoverflow.com/questions/39186282/ffmpeg-not-honoring-sample-rate-in-opus-output
It seems opus files are decoded at 48kHz as possible, with the original sample rate being ignored.

pzelasko:
Yes, I ran into that issue with OPUS previously. I ended up resampling it to 16kHz myself (`recordings.resample(16000)`) and I extended resampling to cuts for the very same reason (`cuts.resample(16000)`) to be able to do it on already-prepared data. 

Unless somebody knows a work-around to force a different OPUS decoder behavior, it is what it is...

lhotse/audio.py
```diff
@@ -93,6 +94,13 @@ def load_audio(
         else:  # self.type == 'file'
             samples, sampling_rate = read_audio(source, offset=offset, duration=duration)
 
+        # A useful case of resampling here is for the opus-format audios. Opus-format audio would be decoded at 48kHz
+        # by force, with the original sampling rate being ignored.
+        if target_sampling_rate is not None and sampling_rate != target_sampling_rate:
```
pzelasko, line 99: We shouldn’t do it this way. The changes to audio.py are duplicating existing Lhotse’s functionality (`recordings.resample(target_sampling_rate)`), and in a way that is incompatible with the rest of the code base (how would e.g. Cut know that it has to specify a target_sampling_rate when load_audio() is called on it?)

lhotse/recipes/gigaspeech.py
```diff
@@ -65,7 +66,13 @@ def parse_utterance(
         audio: Any,
         root_path: Path
 ) -> Optional[Tuple[Recording, List[SupervisionSegment]]]:
-    recording = Recording.from_file(root_path / audio['path'], recording_id=audio['aid'])
+    recording = Recording(id=audio['aid'],
+                          sources=[AudioSource(type='file',
+                                               channels=list(range(int(audio['channels']))),
+                                               source=f'{root_path}/{audio["path"]}')],
+                          num_samples=round(int(audio['sample_rate']) * Seconds(audio['duration']), ndigits=8),
+                          sampling_rate=audio['sample_rate'],
```
pzelasko, line 74: For OPUS files you might want to "hardcode" 48kHz here, and add a line afterwards `recording = recording.resample(16000)`

docs/conf.py
```diff
@@ -20,7 +20,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'lhotse'
-copyright = '2020, Lhotse development team'
+copyright = '2020-2021, Lhotse development team'
```
pzelasko, line 23: ❤️

lhotse/recipes/gigaspeech.py
```diff
@@ -0,0 +1,89 @@
...
+) -> Optional[Tuple[Recording, List[SupervisionSegment]]]:
+    # Opus-format audio would be decoded at 48kHz by force, with the original sampling rate being ignored.
+    opus_decoding_sample_rate = 48000
+
+    recording = Recording(id=audio['aid'],
+                          sources=[AudioSource(type='file',
+                                               channels=list(range(int(audio['channels']))),
+                                               source=f'{root_path}/{audio["path"]}')],
+                          num_samples=round(opus_decoding_sample_rate * Seconds(audio['duration']), ndigits=8),
+                          sampling_rate=opus_decoding_sample_rate,
```
pzelasko, line 77: just double checking -- everything in GigaSpeech is encoded in OPUS, right?

pzelasko:
Thanks a lot, it looks good! The test fails seems to be just a flake. LGTM!

pzelasko:
Hey @jimbozhang, the "speechcolab" dependency doesn't seem to be possible to install with pip. How can I run this recipe?

jimbozhang:
> Hey @jimbozhang, the "speechcolab" dependency doesn't seem to be possible to install with pip. How can I run this recipe?

As it hasn't been released officially, you can install it from TestPyPI for now:

```
pip install -i https://test.pypi.org/simple/ speechcolab
```

@chenguoguo will push a stable release to PyPI soon.

PR #288, opened 2021-04-28
Lazy manifests usable with a non-shuffled sampler + dataset

pzelasko:
With these changes (a utility for format conversion and one more optimization) I'm able to use the cuts in the mmapped arrow format for training full LibriSpeech in snowfall without a noticeable performance penalty. It requires setting ``check_inputs`` in K2 ASR dataset to `False` to avoid costly double iteration over the inputs though.


danpovey (commented):
LGTM!

PR #289, opened 2021-04-29
Automatically inferring the schema for direct Arrow serialization

pzelasko:
This is not on the critical path (mostly useful for convenience) but I was curious to learn more about how Arrow works.


pzelasko:
... I also changed the samplers to "auto-discover" that a cut manifest is lazy, in which case they will not provide "len()" operator, which causes to iterate the whole sampler and may incur a big lag at the training start (some frameworks like pytorch lightning query the len() to show progress bars)

PR #291, opened 2021-04-29
add git commit hash to version in form X.Y.Z.dev-gitgitsha-indexclean

oplatek:
Hi there!

I added git SHA to the version in `setup.py` in the format X.Y.Z.dev-SHA-dirty or X.Y.Z.dev-SHA-clean
The clean option is added when the git index is clean (on all releases).

I find it particularly useful if Lhotse is installed via `pip install git+https://github.com/lhotse-speech/lhotse`

```
$ pip list | grep lhotse   # see the second column for the version example
lhotse                        0.6.0.dev-2840393-clean /fsx/home/oplatek/code/lhotse
```


pzelasko:
Thanks, looks great!

PR #292, opened 2021-05-03
Pickling support for LazyDict + minor fixes

pzelasko:
This solves an issue that is hard to notice unless one uses very large manifests -- when the mmapped file gets pickled for transfer to dataloader's worker processes, Python reads the whole file and tries to pickle it which blows up RAM. I changed the pickling behaviour for that class to only transfer the path to the file and re-open it in the new process. After this change, the training takes approx. 5-6GB CPU RAM per GPU, and I'm able to run snowfall training on full MLS with 4xGPU and on-the-fly feature extraction with no I/O or memory issues (with shuffle=False).


danpovey:
Cool!

PR #297, opened 2021-05-06
CMU and CSLU Kids speech recipes

desh2608:
**Corpora:**
- CMU Kids (LDC97S63)
- CSLU Kids (LDC2007S18).

**Documentation:**
The _corpora_ list in the documentation is changed into a table which looks like the following:

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,9 +33,10 @@ release = '0.1'
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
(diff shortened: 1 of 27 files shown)


pzelasko:
Thanks, that looks great! Could you add a field in that table that says whether the corpus is freely available? (e.g. Free: ✅ / ❌)

pzelasko:
Anyway I'll merge -- we can follow up with this kind of improvements later.

PR #301, opened 2021-05-07
Mtedx

m-wiesner:


lhotse/recipes/mtedx.py
```diff
@@ -0,0 +1,237 @@
...
+
+def download_and_untar(
+    target_dir: Pathlike = '.',
+    force_download: Optional[bool] = False
+    language: str = 'fr',
+) -> None:
+    target_dir = Path(target_dir)
+    target_dir.mkdir(parents=True, exist_ok=True)
+    tar_path = target_dir / f'{language}-{language}.tgz'
+    urlretreive_progress(
```
pzelasko, line 72: typo: should be `urlretrieve_progress`

lhotse/recipes/mtedx.py
```diff
@@ -0,0 +1,237 @@
...
+    'el': 'Greek',
+    'ar': 'Arabic',
+    'de': 'German',
+}
+
+
+def download_and_untar(
+    target_dir: Pathlike = '.',
+    force_download: Optional[bool] = False
+    language: str = 'fr',
```
pzelasko, line 67: could we change this to accept either:
- a list of languages; or
- a single language string; or
- a string "all" (default)?

lhotse/recipes/mtedx.py
```diff
@@ -0,0 +1,237 @@
...
+from cytoolz import sliding_window
+
+import tarfile
+import shutil
+from lhotse import Recording, RecordingSet, SupervisionSegment, SupervisionSet, validate_recordings_and_supervisions
+from lhotse.qa import remove_missing_recordings_and_supervisions, trim_supervisions_to_recordings
+from lhotse.utils import Pathlike, urlretrieve_progress
+import regex as re2
+from functools import partial
+import unicodedata
```
pzelasko, line 38: I don't think `unicodedata` and `regex` are direct lhotse's dependencies -- could you use `lhotse.utils.is_module_available` to check that they are installed raise an error saying to install them otherwise?

pzelasko (commented):
Thanks, looks great! 

Could you add a CLI mode for it (see `lhotse/bin/modes/recipes`) and add an entry in the table in `docs/corpora.rst`?

lhotse/recipes/mtedx.py
m-wiesner: I have used importlib in the past to do this. You just try to import and if it fails you can install with pip. Is that ok with you?

desh2608: I think `unicodedata` is part of the [Python Standard Library](https://docs.python.org/3/library/unicodedata.html). `regex` is a third-party library, so perhaps it may be better to replace it with `re` (unless there is some functionality not available in `re`)?

jtrmal: I think python's re doesn't know unicode character classes  -- e.g. ``\p{Zs}``

m-wiesner: Yenda is right, and I used it just to be able to to \p{Zs} in fact.

pzelasko: It's fine -- just add a check like this for example: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/features/librosa_fbank.py#L85-L90

docs/corpus.rst
```diff
@@ -190,6 +190,8 @@ a CLI tool that create the manifests given a corpus directory.
     - :func:`lhotse.recipes.prepare_ljspeech`
   * - MiniLibriMix
     - :func:`lhotse.recipes.prepare_librimix`
+  * - MTEDx
+    - :func:`lhotse.recipes.prepare_single_mtdex_language`
```
desh2608, line 194:
```suggestion
    - :func:`lhotse.recipes.prepare_mtedx`
```

m-wiesner: Thanks for catching that.

lhotse/recipes/mtedx.py
```diff
@@ -0,0 +1,349 @@
...
+###############################################################################
+# All remaining functions are just helper functions, mainly for text
+# normalization and parsing the vtt files that come with the mtedx corpus
+###############################################################################
+
+# Prepare data for a single language
+def prepare_single_mtedx_language(
+    corpus_dir: Pathlike,
+    output_dir: Optional[Pathlike] = None,
+    language: str = 'language',
```
pzelasko, line 173: The default argument choice seems weird, if we treat it as an internal function maybe it’s better to remove the default value?

pzelasko (commented):
Thanks! Just a couple of comments and LGTM!

lhotse/recipes/mtedx.py
```diff
@@ -0,0 +1,349 @@
...
+from cytoolz import sliding_window
+
+import tarfile
+import shutil
+from lhotse import Recording, RecordingSet, SupervisionSegment, SupervisionSet, validate_recordings_and_supervisions
+from lhotse.qa import remove_missing_recordings_and_supervisions, trim_supervisions_to_recordings
+from lhotse.utils import Pathlike, urlretrieve_progress, is_module_available
+import unicodedata
+
+if not is_module_available('regex'):
```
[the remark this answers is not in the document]
pzelasko: That’s why the tests are failing BTW (good tests 😁)

lhotse/recipes/mtedx.py
m-wiesner: Sure. There was one other point which I wanted to ask about. The dumped json files for arabic for instance print the unicode values using the \u00... format, rather than something readable, which might be of interest. If you use the ascii_only=False flag on json.dump then you can read the output, but maybe there was a reason for serializing with ascii only characters?

pzelasko: Hmm theoretically we’re in the 21st century and everything should be ok if we just write Unicode in the manifests. Yeah… sure. 🤥

I’d suggest to allow writing Unicode by default and if somebody needs to turn it off, they can always manually encode the text values. Or we can add flags to customize this behavior later. WDYT?

pzelasko: JSON RFC https://www.ietf.org/rfc/rfc4627.txt

`3.  Encoding

   JSON text SHALL be encoded in Unicode.  The default encoding is
   UTF-8. `

m-wiesner: I guess I don't really know what SHALL means here. At some point in the future, or just recommended? I think as long as there aren't cases where writing and then reading cause some corruption of the data. it's preferable to be able to read the manifest. The easiest change is just setting ensure_ascii=False in the save_to_json file and then always doing that, but I don't know if that is back-compatible or if that causes other problems.

pzelasko: I agree with you. I think it is unlikely to break anything.

m-wiesner: Do you want that update in this PR, or in a separate commit? I guess we might want to change jsonl as well? I'm not as familiar with the other formats.

pzelasko: Yeah, a follow-up PR is a good idea, let's change JSONL as well (and YAML too if they allow it -- it's only for didactic purposes so it'd be good to have it there too)

PR #303, opened 2021-05-11
Fixed AMI CLI mode

desh2608:
* Add options to make AMI data prep from CLI consistent with python usage.
* Minor bug fix.


pzelasko:
Thanks!

Issue #298, opened 2021-05-06
Alignments in supervision segments


pzelasko:
We should start supporting alignments in Lhotse. I think that the following scheme is expressive enough:

```python

class AlignmentItem(NamedTuple):
  symbol: str
  start: Seconds
  duration: Seconds

SupervisionSegment(
  id='...',
  start=...,
  duration=...,
  text='...',
  ...
  alignment={
    'type': 'word',
    'items': [
      ['word', 0.0, 0.5],
      ['alignment', 0.5, 0.7],
      ...
    ]
  }
)
```

The types would be entirely user-dependent, e.g. it could be "phones", "subwords", "letters", etc. I'm not sure if any other fields are needed besides start and duration...

pzelasko:
... we'd also add functions such as `from_ctm` and `to_ctm` to make it easy to work with standard formats.

danpovey:
Cool!
yes that looks good to me.

desh2608:
In some recipes like AMI, the `SupervisionSegment` themselves are words (since the official annotation provides word-level alignments). How should the `alignments` look like in that case?

P.S.: I can implement the alignment and the CTM conversion thing since I also need it for OVAD.

desh2608:
I believe Kaldi creates segments based on punctuation boundaries (at least in the AMI recipe). I already have an option in the Lhotse recipe for specifying `max-pause` between words to merge them into one supervision segment, so we could just use that.

Are the `AlignmentItem` start times w.r.t. the `SupervisionSegment` start?

PR #304, opened 2021-05-11
[WIP] Adding alignments to supervision

desh2608:
This PR will add alignments to SupervisionSegment, as discussed in https://github.com/lhotse-speech/lhotse/issues/298.


lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
-from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Sequence
+from typing import Any, Callable, Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence
 
 from lhotse.serialization import Serializable
 from lhotse.utils import Seconds, asdict_nonull, exactly_one_not_null, fastcopy, \
     ifnone, index_by_id_and_check, \
     perturb_num_samples, split_sequence
 
 
+@dataclass(frozen=True, unsafe_hash=True)
```
pzelasko, line 12: when it's derived from a NamedTuple, the dataclass part is not needed (I think it would actually be in conflict)

lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
+    start: Seconds
+    duration: Seconds
+
+    @property
+    def end(self) -> Seconds:
+        return round(self.start + self.duration, ndigits=8)
+
+    def with_offset(self, offset: Seconds) -> 'AlignmentItem':
+        """Return an identical ``AlignmentItem``, but with the ``offset`` added to the ``start`` field."""
+        return fastcopy(self, start=round(self.start + offset, ndigits=8))
```
pzelasko, line 24: I think `fastcopy` is not going to work on a named tuple, since it has only 3 members I think it's better to explicitly instantiate a new named tuple (i.e. `AlignmentItem(self[0], round(...), self[2])` - IIRC indexing is faster than named attr access and this code is potentially executed many times on large supervision/cut sets)

lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
+
+    def perturb_speed(
+            self,
+            factor: float,
+            sampling_rate: int,
+    ) -> 'AlignmentItem':
+        """
+        Return a ``AlignmentItem`` that has time boundaries matching the
```
pzelasko, line 32: a -> an

lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
+        recording/cut perturbed with the same factor. See `perturb_speed()` method
+        in `SupervisionSegment` for details.
+        """
+        start_sample = round(self.start * sampling_rate)
+        num_samples = round(self.duration * sampling_rate)
```
pzelasko, line 37: I see this is a copy of supervision segment's version -- I think it should have been `compute_num_samples(...)` (`from lhotse.utils`) instead of explicit multiplication and rounding, as that method handles some edge cases better. It's probably going to work anyway, but it's more consistent with the rest of the codebase. Could you change it?

pzelasko (commented):
Looks good! I left some comments and suggestions. 

Before you go on with adding more stuff, I suggest you add/extend unit tests to cover the new code paths that are being introduced. In particular you might want to see `lhotse.testing.dummies` (https://github.com/lhotse-speech/lhotse/blob/master/lhotse/testing/dummies.py) and add alignments to supervision segments to see how many existing tests will break (if any). You can also extend `lhotse.testing.fixtures` with it which performs randomized testing with PyTorch datasets (https://github.com/lhotse-speech/lhotse/blob/master/lhotse/testing/fixtures.py#L81). 

Your questions:

> I don't fully understand what you meant about using __post_init__ in SupervisionSegment. Could you explain a little more?

When you store the manifest in JSON/sth else, it will convert `AlignmentItem` to a 3-element list. When we read that JSON, we need to convert these 3-element lists back to `AlignmentItem`s. We can do it in two ways: one is using a special member in dataclasses that is called `__post_init__`, which is called after `__init__` (this allows to use default `__init__` in dataclasses and still customize it). But I think there's actually a better way to do it -- I think there is a method called `from_dict` in supervision segment that you can adjust instead.

> Where should from_ctm and to_ctm be implemented? CTM word supervisions are usually at the recording level, so a CTM would more naturally be associated with a SupervisionSet I think.

Yeah `SupervisionSet` makes sense to me. Actually if we read with `from_ctm`, we would not have the speaker etc. information anymore -- maybe we can instead make it sth like `add_ctm_info` (not sure how to name it well...) that adds alignments to an existing object? Or we can read and merge two supervision sets. Or maybe there is another way -- it's up to you.

> Some functions in cuts.py which prepare supervision masks can be modified to optionally use the alignments for mask generation. This would be particularly useful for VAD training etc.

Makes sense!

> What's a natural workflow for adding alignments? Should they be included in Lhotse recipes, or do users get to add them to the prepared supervisions manifest?

I think users will typically extend existing supervision manifests, as not too many corpora provide alignments.

danpovey:
Guys, I haven't looked at this in detail, but I do want to mention something that's been on my mind recently RE snowfall and (eventual) Icefall.

Firstly, the timeline for designing icefall is basically in the next few weeks we'll be figuring out how to start, and we'd like Icefall to be "officially releasable" in time for the Intespeech tutorial around September 1st.  Think of Icefall as a "properly designed" version of snowfall.

Anyway, the issue related to this is, I'd like to be able to deal with things like alignments in Icefall/Snowfall, e.g. for purposes of building trees and the like, or for segmenting training data, data-cleaning and analysis, and so on.  I'm not saying the format we'd use for such things would necessarily be identical to the the way they are represented in Lhotse.  But a workflow that I think might end up being fairly common, is we train some fairly basic neural net model that only sees limited acoustic context (like a TDNN), for purposes of data alignment.    If we're using a model where there is a phone for silence or at least a unit that represents "end-of-word", it may be useful to have a blank-free topology here.  I don't believe we currently have code for such a topology in snowfall.

And we might want some way to represent those data alignments.  E.g. one simple representation might just be the label sequences, indexed somehow by utterance-id.  There are several different label-sequences that might be relevant here, depending on the type of system: the ilabel from the model, the phone-label "without repetitions", which we could store as a separate attribute on the graphs by using the "inner_labels=xxx" arg to compose in the appropriate stage of graph creation; and the olabel which is the word label.  One possibility is to just store these as a dict indexed by utterance-id and then by 'ilabel', 'olabel' and 'phone_label' (for phone labels without repetitions) or something like that, and store it as a .pt file with torch.save().

pzelasko:
That makes sense to me. In another project, where I prototyped some of the alignment-related stuff, I'm using this CTM-like format together with methods to convert to/from a frame-level alignment (int sequence). It handles conversions such as 10ms - 12ms frame shift etc. I think it's going to play well with what you're planning for Icefall.

danpovey:
Something else the alignments might be useful for, if we can get word-level alignments, is in designing BPE-type units.  It's desirable to have units where words that have a short pronunciation also have a short representation, so we can keep the frame rate slow.

lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
+from typing import Any, Callable, Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence
 
 from lhotse.serialization import Serializable
 from lhotse.utils import Seconds, asdict_nonull, exactly_one_not_null, fastcopy, \
     ifnone, index_by_id_and_check, \
     perturb_num_samples, split_sequence
 
 
+@dataclass(frozen=True, unsafe_hash=True)
+class AlignmentItem(NamedTuple):
```
csukuangfj, line 13: If `frozen` is True, is there a need to inherit from a tuple?
It is not modifiable any way when `frozen` is True.

lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
+        return fastcopy(self, start=round(self.start + offset, ndigits=8))
+
+    def perturb_speed(
+            self,
+            factor: float,
+            sampling_rate: int,
+    ) -> 'AlignmentItem':
+        """
+        Return a ``AlignmentItem`` that has time boundaries matching the
+        recording/cut perturbed with the same factor. See `perturb_speed()` method
```
csukuangfj, line 33:
```
See :meth:`SupervisionSegment.perturb_speed`
```
is better as it will create a cross-reference link to that method in the generated documentation.

lhotse/supervision.py
```diff
@@ -1,14 +1,68 @@
...
+        num_samples = round(self.duration * sampling_rate)
+        new_start = perturb_num_samples(start_sample, factor) / sampling_rate
+        new_duration = perturb_num_samples(num_samples, factor) / sampling_rate
+        return fastcopy(
+            self,
+            start=new_start,
+            duration=new_duration
+        )
+
+    def trim(self, end: Seconds, start: Optional[Seconds] = 0) -> 'AlignmentItem':
```
csukuangfj, line 46: Why not to use
```
start: Seconds = 0
```
?

The code does not handle the case when `start is None`.

`Optional[Seconds]` means `Union[Seconds, None]` but the code does not support `start == None`.

lhotse/supervision.py
pzelasko: I thought about using tuples to reduce the memory overhead and the serialization overhead (store it as a list `['word', 0, 0.7]` rather than dict `{'symbol': 'word', 'start': 0, 'duration': 0.7}`). I did a simple check in jupyter and the named tuple variant is 2x faster to create and consumes 2x less memory...

pzelasko: (I am concerned about this because there could be a lot of these objects around)

lhotse/supervision.py
```diff
@@ -6,61 +6,57 @@
...
-        Return a ``AlignmentItem`` that has time boundaries matching the
-        recording/cut perturbed with the same factor. See `perturb_speed()` method
-        in `SupervisionSegment` for details.
+        Return an ``AlignmentItem`` that has time boundaries matching the
+        recording/cut perturbed with the same factor. See :meth:`SupervisionSegment.perturb_speed` 
+        for details.
         """
-        start_sample = round(self.start * sampling_rate)
-        num_samples = round(self.duration * sampling_rate)
+        start_sample = round(self[1] * sampling_rate)
```
pzelasko, line 40: if you're using compute num samples below, do so here as well

lhotse/supervision.py
```diff
@@ -113,7 +109,7 @@ def perturb_speed(
         :return: a modified copy of the current ``Recording``.
         """
         start_sample = round(self.start * sampling_rate)
```
pzelasko, line 111: same as above

lhotse/testing/dummies.py
```diff
@@ -34,14 +34,27 @@ def dummy_recording(unique_id: int) -> Recording:
...
+        duration=duration
+    )
 
-def dummy_supervision(unique_id: int, start: float = 0.0, duration: float = 1.0, text: str = "irrelevant") -> SupervisionSegment:
+def dummy_supervision(
+    unique_id: int,
+    start: float = 0.0,
+    duration: float = 1.0,
+    text: str = "irrelevant",
+    alignment: Optional[Dict[str, List[AlignmentItem]]] = None
```
pzelasko, line 49: I suggest to use `dummy_alignment()` as the default value, which will let you leverage a lot of existing unit tests for this

lhotse/utils.py
```diff
@@ -66,6 +69,8 @@ def non_null_dict_factory(collection):
         for key, val in d.items():
             if val is None:
                 remove_keys.append(key)
+            if convert_named_tuples and isinstance(val, tuple) and hasattr(val, '_fields'):
+                d[key] = val._asdict()
```
pzelasko, line 73: I think we should serialize named tuples (specifically, alignment) as lists rather dicts to save the space in the manifests.

desh2608: I have made this change in the latest commit. Some JSONL serialization tests are failing. I'm looking into that now.

desh2608: It seems this is because we convert the NamedTuple to a list, which now has 2 `float` types and an `str` type. The `pyarrow.json` cannot read lists with such mixed types. Do you have any suggestions to handle this?

pzelasko: Got it. In that case I think what you suggested makes more sense than trying to hack it somehow. People using alignments in supervisions might want to use gzip compression.

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
     def from_dicts(data: Iterable[Dict]) -> 'SupervisionSet':
         return SupervisionSet.from_segments(SupervisionSegment.from_dict(s) for s in data)
     
-    def add_alignments_from_ctm(ctm) -> 'SupervisionSet':
+    def add_alignments_from_ctm(self, ctm_file: Pathlike, type: str = 'word') -> 'SupervisionSet':
```
pzelasko, line 257: I'd suggest `with_alignments_from_ctm`, `add` suggests we're mutating the original object

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
         
-        """
-        pass
-
+        :param ctm: Path to CTM file.
+        :param type: Alignment type (optional, default = `word`).
+        :return: A new SupervisionSet with AlignmentItem objects added to the segments.
+        """
+        ctm_words = []
+        with open(ctm_file, 'r') as f:
```
pzelasko, line 266:
```suggestion
        with open(ctm_file) as f:
```

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+            for line in f:
+                reco_id, channel, start, duration, symbol = line.strip().split()
+                ctm_words.append((reco_id, channel, float(start), float(duration), symbol))
```
pzelasko, line 269: shouldn't `channel` be an int?

pzelasko: (at least in Lhotse we always map them to ints and expect int, so be careful there)

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+        ctm_words = sorted(ctm_words, key=lambda x:x[0])
```
pzelasko, line 270: is python's `sorted` guaranteed to be stable? otherwise you might want to change the key to be a tuple of `(reco_id, start)`

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+        reco_to_ctm = defaultdict(list, {k: list(v) for k,v in groupby(ctm_words, key=lambda x:x[0])})
+        segments = []
+        for reco_id in reco_to_ctm:
+            segs = [s for s in self if s.recording_id == reco_id]
```
pzelasko, line 274: I think a faster variant is `self.find(recording_id=reco_id)` which internally builds an index and caches it

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+            for seg in segs:
+                alignment = [AlignmentItem(word[4], word[2], word[3]) for word in ctm_words 
+                             if overspans(
+                                 TimeSpan(start=seg.start, end=seg.end),
```
pzelasko, line 278: I think it's OK to just pass `seg` here due to duck typing

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+                                 TimeSpan(word[2], word[2] + word[3])
+                                ) and seg.channel == word[1]
+                            ]
+                segments.append(fastcopy(seg, alignment={type: alignment}))
+        return SupervisionSet.from_segments(segments)
+                          
+
+    def write_ctm(self, ctm_file: Pathlike, type: str = 'word') -> None:
```
pzelasko, line 286: how about `write_alignment_to_ctm`? it's un-ambiguous, as it's also possible to convert "normal" supervisions to a CTM

lhotse/supervision.py
```diff
@@ -251,12 +254,49 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+                reco_id, channel, start, duration, symbol = line.strip().split()
+                ctm_words.append((reco_id, channel, float(start), float(duration), symbol))
+        ctm_words = sorted(ctm_words, key=lambda x:x[0])
+        reco_to_ctm = defaultdict(list, {k: list(v) for k,v in groupby(ctm_words, key=lambda x:x[0])})
+        segments = []
+        for reco_id in reco_to_ctm:
+            segs = [s for s in self if s.recording_id == reco_id]
+            for seg in segs:
+                alignment = [AlignmentItem(word[4], word[2], word[3]) for word in ctm_words 
+                             if overspans(
```
pzelasko, line 277: what should happen if the alignment item overlaps a supervision segment? maybe we should issue some warning about potential mismatch? I'm not sure.

pzelasko (commented):
Thanks @desh2608! I left some comments, could you also add unit tests for the two new methods of import/export to CTM?

pzelasko:
(you can add dummy CTMs to `test/fixtures`, don't forget to add them to git)

lhotse/supervision.py
desh2608: Ideally, yes. But sometimes they are also denoted by `A`, `B`, etc. in CTM files.

lhotse/supervision.py
desh2608: Yes, it is guaranteed (https://docs.python.org/3/library/functions.html#sorted). Also, we don't particularly need the alignments to be sorted by time I suppose? This sorting by `reco_id` is just to enable `groupby`.

pzelasko: We probably don't, but I think it's a reasonable expectation that they are sorted (I know I'd be surprised if they're not). If it's guaranteed then it's cool :)

lhotse/supervision.py
jtrmal: yes, channels do not have to be int

desh2608:
@pzelasko I have made the changes you suggested and also added tests for reading/writing CTM. Perhaps the PR can be merged now if you  (I will try to make the changes to supervision mask generation in Cuts later).

pzelasko:
OK I am merging as it is. We can try them out and adjust or optimize as needed.

Thanks for all the work on this, great job!

lhotse/supervision.py
```diff
@@ -97,13 +175,51 @@ def transform_text(self, transform_fn: Callable[[str], str]) -> 'SupervisionSegm
...
+        return fastcopy(
+            self,
+            alignment={
+                ali_type:[
+                    item.transform(transform_fn=transform_fn)
+                    if ali_type == type else item
+                    for item in ali
+                ]
+                for ali_type, ali in self.alignment.items()
+            } if self.alignment else None
```
csukuangfj, line 203: Line 192 already checks whether `self.alignment` is `None`, so no need to do an extra check here.

lhotse/supervision.py
```diff
@@ -137,7 +253,57 @@ def from_segments(segments: Iterable[SupervisionSegment]) -> 'SupervisionSet':
...
+                for seg in self.find(recording_id=reco_id):
+                    alignment = [AlignmentItem(symbol=word[4], start=word[2], duration=word[3]) for word in reco_to_ctm[reco_id] 
+                                    if overspans(seg, TimeSpan(word[2], word[2] + word[3]))
+                                    and (seg.channel == word[1] or not match_channel)
+                                ]
+                    num_overspanned += len(alignment)
+                    segments.append(fastcopy(seg, alignment={type: alignment}))
+            else:
+                segments.append([s for s in self.find(recording_id=reco_id)])
+        print (segments)
```
csukuangfj, line 287: This kind of debugging code should not be checked into the final commit.

lhotse/supervision.py
```diff
@@ -97,13 +175,51 @@ def transform_text(self, transform_fn: Callable[[str], str]) -> 'SupervisionSegm
...
+            } if self.alignment else None
+        )
 
     def to_dict(self) -> dict:
-        return asdict_nonull(self)
+        """
+        We use the recursive conversion only if alignments are present, since it may
+        potentially be slower due to type checking of member objects.
+        """
+        return asdict_nonull(self) if self.alignment is None else asdict_nonull_recursive(self)
```
csukuangfj, line 211: The `AlignmentItem` class is no longer a subclass of `NamedTuple`, is `asdict_nonull_recursive` really necessary? It makes the code slow, I believe.

Also, the code in `cut.py` should also be updated to remove `asdict_nonull_recursive`.

pzelasko: Good point, @desh2608 could you address Fangjuns comments?

Btw @csukuangfj we should make sure that Lhotse alignments and Snowfall alignments interact well together. If there are any additions/changes that are helpful on Lhotse side, please let us know.

desh2608:
@csukuangfj thanks for your review. I have made suggested changes in https://github.com/lhotse-speech/lhotse/pull/313.

PR #312, opened 2021-06-04
Support for reading and writing features to URLs (HTTP, S3, GCP, Azure, etc.)

pzelasko:
Requires an optional dependency `smart_open`.


pzelasko:
I also added a CLI mode for uploading existing feature manifests to URLs. 

I validated that it works but I guess I have no way to add an automated test to the CI. Merging.

Issue #314, opened 2021-06-06
Save augmented/mixed audio


entn-at:
Hi,
I was wondering if there is currently a way to

1. Create a CutSet
2. Augment it and mix it with noise (CutSet of MixedCut, I guess)
3. Store the transformed audio to disk (and the CutSet with appropriate metadata)

Thus far, it seems the only way would be to create a custom `FeatureExtractor` that just stores the audio samples, but it wouldn't be possible to extract features from the resulting `FeatureSet`.

If it is currently not possible, would this be seen as a welcome enhancement?

Thanks,
Ewald

pzelasko:
Hi Ewald,

Yes, it’s a welcome contribution. I believe this is exactly the same as https://github.com/lhotse-speech/lhotse/issues/150 — if that looks interesting to you, please submit a PR.

entn-at:
Hi Piotr,
thanks, you're right, that issue is exactly the same. I will look into this!

Issue #150, opened 2020-11-24
Feature: AudioCutting feature extractor


janvainer:
Is it possible right now to use CutSet to cut all recordings into separate wav files and dump these files on disk?
If not, would it be a considerable feature?

It could be used in a similar way to feature extraction.
```
recording_set = RecordingSet.from_yaml('audio.yml')
audio_cutter = AudioCutter()

with AudioWriter() as storage:
    builder = FeatureSetBuilder(feature_extractor=audio_cutter, storage=storage)
    feature_set = builder.process_and_store_recordings(
        recordings=recording_set,
        num_jobs=8
    )
```
or the cutset would simply have a function `CutSet.cut_audio(path)` or similar.
The audio cuts could be stored in subfolders of path based eg on recording name. Along with the wav cuts, there would also be a manifest file for the Cuts. Let me know if this would be of interest or if something similar is already possible. Thanks

pzelasko:
That's a cool idea. Since `CutSet` has `compute_and_store_features` already, we could add `compute_and_store_recordings` to it, with the same API (minus minor modifications):

```python
def compute_and_store_recordings(
    self,
    output_dir: Pathlike,
    executor: Optional[Any] = None,
    augment_fn: Optional[AugmentFn] = None
) -> 'CutSet':
```

I think we'll eventually have different classes for `AudioStorage` but at this point it's fine to just support a path to a directory. The `executor` thing is nice because it allows you to choose between no parallelism, multithreading, multiprocessing and distributed processing (e.g. with Dask). I think having `augment_fn` is beneficial in case you want to leverage WavAugment/Torchaudio integration into Lhotse to transform your data before saving.

It should return a `CutSet` that "converted" every `MixedCut` to a simple `Cut`, since the audio files contain mixes after doing that.

WDYT? Want to have a go at it? :)

danpovey:
Might be nice to have one place where "executor" is documented in full, and
have the documentation for individual uses explain it very briefly and
refer to that.
Possibly the same for other concepts.  Just an idea.

pzelasko:
Thanks @danpovey - I opened an issue for that #152

janvainer:
@pzelasko being able to use augmentations and using parallel execution is a really cool idea I think. 
Sure I will give it a try. Just a question - what should be the dir structure for saving? By recordings or perhaps by speakers? The latter would not work if the data had no speaker annotations. So by recording would probably make most sense. We could also do it bu Cuts though.

[...]

PR #316, opened 2021-06-07
Store waveforms of cuts as audio recordings to disk

entn-at:
This PR implements a method `compute_and_store_recordings` in `CutSet` that stores waveforms of cuts as audio recordings (specifically FLAC) to disk. It addresses issues #150 and #314. I added some basic tests.

As proposed in #150, for each cut, a sub-directory will be created that starts with the first three characters of the cut ID. The waveform of the cut is then stored as FLAC audio file in the sub-directory using the cut ID as filename and `.flac` as suffix.


pzelasko:
Looks good to me, thanks! LGTM!

PR #318, opened 2021-06-09
Support for importing Kaldi's feats.scp and reading features directly from scp/ark

pzelasko:


pzelasko:
Seems to work, I might add a specific test for importing/exporting Kaldi data dir in the future.

PR #319, opened 2021-06-10
WIP: Add posteriors to Cut.

csukuangfj:
Required by https://github.com/k2-fsa/snowfall/pull/210

### DIFF
--- a/lhotse/post.py
+++ b/lhotse/post.py
@@ -1,0 +1,73 @@
+# Copyright (c)  2021  Xiaomi Corp.       (author: Fangjun Kuang)
+
+from dataclasses import dataclass
+
+import numpy as np
+
+from lhotse.features.io import FeaturesWriter, get_reader
+from lhotse.utils import uuid4
+
+
+@dataclass(order=True)
+class Posteriors:
+    """
+    Represent the output of the neural network.
+    It contains metadata about how it's stored: storage_type describes "how to read it", for now
+    it supports numpy arrays serialized with np.save, as well as arrays compressed with lilcom;
+    storage_path is the path to the file on the local filesystem.
+    """
+    num_frames: int
+
+    num_dim: int  # model output dim
+
+    # The subsampling factor of the model
+    subsampling_factor: int
+
+    # Storage type defines which features reader type should be instantiated
+    # e.g. 'lilcom_files', 'numpy_files', 'lilcom_hdf5'
+    storage_type: str
+
+    # Storage path is either the path to some kind of archive (like HDF5 file) or a path
+    # to a directory holding files with feature matrices (exact semantics depend on storage_type).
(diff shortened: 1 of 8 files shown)


csukuangfj:
Will add tests soon.

lhotse/post.py
```diff
@@ -0,0 +1,73 @@
...
+class Posteriors:
+    """
+    Represent the output of the neural network.
+    It contains metadata about how it's stored: storage_type describes "how to read it", for now
+    it supports numpy arrays serialized with np.save, as well as arrays compressed with lilcom;
+    storage_path is the path to the file on the local filesystem.
+    """
+    num_frames: int
+
+    num_dim: int  # model output dim
```
pzelasko, line 21: How about one of `num_classes`, `num_hidden`, `num_outputs`, `num_activations`, etc.? `num_dim` suggests to me it's more like `len(shape)`

lhotse/post.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # Load and return the posteriors from the storage
+        return storage.read(self.storage_key)
+
+    @staticmethod
+    def from_dict(data: dict) -> 'Posteriors':
+        return Posteriors(**data)
+
+
+def save_posteriors(
```
pzelasko, line 50: Would it make sense to call this method `from_array()` and make it a `staticmethod` on `Posteriors`? I think it'd be consistent with constructors of other objects in Lhotse.

lhotse/post.py
```diff
@@ -0,0 +1,73 @@
...
+    Store ``posts`` array on disk.
+
+    :param posts: a numpy ndarray (2-D) containing posteriors.
+    :param: subsampling_factor: the subsampling factor of the model.
+    :param storage: a ``FeaturesWriter`` object to use for array storage.
+    :return: An instance of :class:`Posteriors`.
+    """
+    assert posts.ndim == 2
+
+    posts_id = str(uuid4())
```
pzelasko, line 65: you might consider adding a parameter `post_id: Optional[str] = None`, in case somebody wants to search posteriors by e.g. cut IDs; the default behavior of assigning a random ID looks good.

pzelasko (commented):
Thanks, it looks really good to me! 

Some comments (not necessarily for this PR):
- we should support storing posteriors in sparse matrices, which is also something that Kaldi does (see e.g. https://github.com/kaldi-asr/kaldi/blob/master/src/bin/prob-to-post.cc#L28) -- it's not an issue today, but if we want to extend k2 acoustic models to predict senones, we will have thousands of output classes
- for lilcom compression, it would be good to auto-detect that the arrays have compatible/incompatible range of values (e.g. if you store [0-1] values with lilcom it's going to lose a lot of data, but the log-domain values works well)

PR #323, opened 2021-06-26
use whitespace in kaldi field-splitting

dophist:
use whitespace in kaldi field-splitting, sometimes people may use tab as separator to generate kaldi resources


pzelasko:
Thanks!

PR #329, opened 2021-07-03
Adding GigaSpeech recipe CLI, minor GigaSpeech fixes, renaming "lhotse obtain" to "lhotse download"

pzelasko:


pzelasko:
I tested that it works (the download throws not implemented error from speechcolab -- we can adjust it on Lhotse side after speechcolab release if it doesn't work)

@jimbozhang FYI

PR #331, opened 2021-07-06
Adding trim_to_supervisions to Cut; new options to include acoustic context when using trim_to_supervisions

pzelasko:


lhotse/cut.py
```diff
@@ -262,6 +270,108 @@ def plot_features(self):
...
         return plt.matshow(features)
 
+    def trim_to_supervisions(
+            self,
+            keep_overlapping: bool = True,
+            min_duration: Optional[Seconds] = None,
+            context_direction: Literal['center', 'left', 'right', 'random'] = 'center',
+    ) -> List['Cut']:
+        """
+        Splits the current :class:`.Cut` into as many cuts as there supervisions (:class:`.SupervisionSegment`).
```
csukuangfj, line 280:
```suggestion
        Splits the current :class:`.Cut` into as many cuts as there are supervisions (:class:`.SupervisionSegment`).
```

pzelasko:
It seems that with some effort, I got it working with a slow-ish, but mostly reasonable processing time. An example of what it does, using one of GigaSpeech recordings that are very long:

60 seconds of the recording after the first supervision starts (there are many supervisions):
image

Creating 20 second sub-cuts, as many as there are supervisions, each one centered on its corresponding supervision:
image

And the next two cuts:
image

Note that this due to `keep_overlapping=False`, the supervisions that didn't fully fit in the cut were rejected. This makes sure we don't learn bogus alignments in alignment-free training, but still use that speech as the acoustic context. Maybe we can figure out a way to split supervisions based on alignment in these cases in the future, but likely not before the tutorial.

pzelasko:
... the next step is to actually train snowfall models on some GigaSpeech subset -- using "vanilla" supervisions with padding first, and then "contextual" supervisions. Hopefully it's going to show some improvement...

danpovey:
Cool!

PR #334, opened 2021-07-07
Chunked HDF5 feature storage + minor recording fixes + adjust GigaSpeech recipe

pzelasko:
This is to solve an issue with GigaSpeech-like data prep; basically, computing features for a lot of short cuts of long recordings may be very wasteful, if:
a) the whole recording needs to be read each time; or
b) we are running some data augmentation with sox etc. on the recordings.

On the other hand, Lhotse storages we had so far didn't allow for both high compression with lilcom, and reading select chunks of data. This PR fixes it by adding a chunked lilcom storage (for now, just for HDF5).

It is somewhat slower to read (or decompress, I haven't profiled it in detail) whole matrices, but it's faster when we collect chunks. The benchmark below is done on mini-librispeech dev-clean-2, comparing old and new lilcom HDF5 storage readers.

image

The difference in storage size is minimal (36M old vs 37M chunked). Not sure whether to make it default or not -- not changing that for now.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -349,7 +349,7 @@ class Recording:
             by affixing it with "_sp{factor}".
         :return: a modified copy of the current ``Recording``.
         """
-        transforms = self.transforms if self.transforms is not None else []
+        transforms = self.transforms.copy() if self.transforms is not None else []
         transforms.append(Speed(factor=factor).to_dict())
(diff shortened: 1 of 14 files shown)


pzelasko:
This seems to be working well -- merging, might add some other things in follow-up PRs

PR #335, opened 2021-07-08
Wrapper around smart_open to cache transport_params for every call

oplatek:
## Motivation
The `smart_open.open` function when used with default value setups all necessary parameters to access e.g. S3 cloud `uri`.
However, it typically means that it creates a session e.g. `boto3.session` with `client('s3')` for AWS S3 access.
Creating such a session for every call is slow.
If we reused a single session for AWS S3 access for 1000 calls the Wall time was reduced from 44s to 18.9s.

This PR introduces SmartOpen singleton class
which caches transport_params and thus e.g. the `client('s3')` 
In addition, it imports smart_open.open only once and does not import it in every call.

### DIFF
--- a/lhotse/utils.py
+++ b/lhotse/utils.py
@@ -44,20 +44,23 @@ class SmartOpen:
 
     The example demonstrates that instantiating S3 `session.client` once,
     instead using the defaults and leaving the smart_open creating it every time
-    has dramatic performance benefits of 44s vs 18.9s Wall time.
+    has dramatic performance benefits.
 
-    import boto3
-    session = boto3.Session()
-    client = session.client('s3')
-    from lhotse.utils import SmartOpen
+    Example::
 
-    if not slow:  # switch between 44s vs 18.9 Wall time
-        SmartOpen.setup(transport_params=dict(client=client))
-
-    # Simulating SmartOpen usage as in Lhotse data structures: AudioSource, Features, etc.
-    for i in range(1000):
-        SmartOpen.open(s3_url, 'rb') as f:
-            source = f.read()
+        >>> import boto3
+        >>> session = boto3.Session()
+        >>> client = session.client('s3')
+        >>> from lhotse.utils import SmartOpen
+        >>>
+        >>> if not slow:
+        >>>     # Reusing a single client speeds up the smart_open.open calls
+        >>>     SmartOpen.setup(transport_params=dict(client=client))
+        >>>
+        >>> # Simulating SmartOpen usage as in Lhotse data structures: AudioSource, Features, etc.
+        >>> for i in range(1000):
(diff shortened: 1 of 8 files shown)


lhotse/utils.py
```diff
@@ -30,6 +31,47 @@
 _lhotse_uuid: Optional[Callable] = None
 
 
+class SmartOpen:
```
pzelasko, line 34: Could you copy a part of your described motivation to the description of this class? Also if you could add sth like:

    Example of setting up global credentials::

        >>> SmartOpen.setup(…)

I will re use this doc for Lhotse’s documentation later.

pzelasko (commented):
Looks great, thanks! Let’s extend the doc a bit and LGTM

lhotse/utils.py
```diff
@@ -30,6 +31,69 @@
...
+    The example demonstrates that instantiating S3 `session.client` once,
+    instead using the defaults and leaving the smart_open creating it every time
+    has dramatic performance benefits of 44s vs 18.9s Wall time.
+
+    import boto3
+    session = boto3.Session()
+    client = session.client('s3')
+    from lhotse.utils import SmartOpen
+
+    if not slow:  # switch between 44s vs 18.9 Wall time
```
pzelasko, line 54: Sorry to be nit-picking -- this 44s vs 18.9s seems very specific to your use-case, could we remove it?

lhotse/utils.py
```diff
@@ -30,6 +31,69 @@
...
+        SmartOpen.setup(transport_params=dict(client=client))
+
+    # Simulating SmartOpen usage as in Lhotse datastructures: AudioSource, Features, etc.
```
pzelasko, line 57:
```suggestion
    # Simulating SmartOpen usage as in Lhotse data structures: AudioSource, Features, etc.
```

lhotse/utils.py
pzelasko: Also -- can you format it like:

```
Example::
    >>> import boto3
    >>> session = boto3.Session()
    ...
```

This will render well with sphinx.

oplatek: Sure!

oplatek: Can you pls check that it is ok now?

Or just tell me how to generate (and check) the docs myself.

pzelasko: The auto generated docs for this PR are here https://lhotse--335.org.readthedocs.build/en/335/

RTD used to add it as a CI test result in GH but the integration broke when I moved Lhotse to lhotse-speech organization… and I couldn’t fix it. 

Unfortunately it seems that the API reference doesn’t list lhotse.utils so you won’t see it there anyway.. I haven’t gotten to re-working this part of the docs yet.

pzelasko:
Looking good, thanks!

Issue #338, opened 2021-07-15
Streaming feature extraction


pzelasko:
The current feature extraction methods in Lhotse are scaling poorly to long (~1h) inputs and eat up a lot of memory when processing them. The natural solution seems to be "streaming" / "chunking" extraction. Ideally, it would produce identical outputs as the "offline" one. Some thoughts:

- I don't think it's possible to have a truly generic wrapper for chunking the inputs that will work with arbitrary feature extractor (like librosa), sampling rates, and frame shifts -- external libraries might be doing the windowing differently than Kaldi's "snip_edges=False". 
- I really like how Kaldi designed the online feature pipelines, but it's a considerable effort to mirror that in Lhotse and I'm not sure there is real benefit from that (at least at this stage of k2 and snowfall); anyway, this should be designed to work well across all our libraries
- we might end up building a half-baked solution just to overcome these issues for specific sampling rates, frame shifts, and extractors to enable experimenting on large quantities of long recordings (e.g. GigaSpeech XL); I might have a prototype although it feels a bit too hacky to push into the codebase...

@danpovey @csukuangfj any thoughts?

pzelasko:
... also I noticed that the torchaudio sox augmentation such as speed perturbation consumes a lot of memory for longer recordings, not sure if there's anything we can do about it though.

sw005320:
Is there any situation that we need to process the long recording in the training stage (I'm expecting that speed perturbation is only performed in the training stage)?

pzelasko:
I think even Kaldi (as an example of very efficient implementation) has to process the whole recording through sox for speed perturbation (unless you cut the recording explicitly into short audio files for each utterance); although that seems to be done in a streaming/memory-friendly way as I have not experienced OOM with it yet. Consider:

```bash
sp0.9-AUD0000000015 ffmpeg -i /export/corpora5/gigaspeech/audio/audiobook/P0001/AUD0000000015.opus -ar 16000 -f wav pipe:1 \
| sox -t wav - -t wav - speed 0.9 \
| sox --vol 1.98800028353 -t wav - -t wav - |
```

[...]

PR #339, opened 2021-07-17
Special OPUS logic for efficient chunk reads and auto-resampling

pzelasko:
This PR is related to #338 and aims to overcome the problem of long-recording feature extraction by enabling efficient on-the-fly feature extraction for chunks. I've added OPUS reading functions that:
- avoid reading the whole OPUS file into memory just in order to get a chunk;
- avoid OPUS resampling with torchaudio sox_effects by doing it directly in ffmpeg.


pzelasko:
Note: if you have any manifests that refer to OPUS files, it's best to re-generate them after this PR

PR #340, opened 2021-07-19
Avoid archive downloads when the unpacked directory exists 

pzelasko:
(+ export some download method names to lhotse.recipe namespace)

@csukuangfj


pzelasko:
... I thought about building some "CorpusDownloader" class but every corpus does it a little bit differently, I don't think I want to go down that way.

PR #343, opened 2021-07-23
Custom DataLoader with two levels of subprocess workers

pzelasko:
... for the lack of a better name, `LhotseDataLoader`. The main difference between this and `torch.utils.data.DataLoader` is that `LhotseDataLoader` allows to launch subprocesses inside of its workers. This is useful for working with dataset classes which perform dynamic batching and need to perform concurrent I/O to read all the necessary data from disk/network.

This is tested locally with good speed-ups from the inner worker pool, I'll have to test it on the real thing next.

### DIFF
--- a/docs/datasets.rst
+++ b/docs/datasets.rst
@@ -49,11 +49,11 @@ A typical Lhotse's dataset API usage might look like this:
     for batch in dloader:
         ...  # process data
 
-Pre-computed vs. on-the-fly: input strategies
----------------------------------------------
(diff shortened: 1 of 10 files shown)


pzelasko:
As this is unlikely to break anything and tests are passing I'll just merge -- the ``LhotseDataLoader`` works with GigaSpeech and seems to alleviate some of the I/O issues, but I still don't know if it's going to be quirky or not. Users do not have to switch to it, our stuff still works perfectly fine with standard ``DataLoader`` as before.

PR #344, opened 2021-07-23
ZipSampler for batches constructed from different cut sources

pzelasko:
@danpovey I found a simple way to implement what we discussed before. I looked at K2SpeechRecognitionDataset and I think it will produce correct batches, as long as its `cuts` argument is a concatenation of all the cut sets used by all the samplers.

I am considering changing the samplers to return a list of `Cut`s instead of cut ID; then some dataset classes (like our ASR dataset) won't need to store their own `cuts` for anything. It will also make it easier to implement any "streaming" samplers that don't read the whole cut set into memory.

### DIFF
--- a/lhotse/dataset/sampling.py
+++ b/lhotse/dataset/sampling.py
@@ -604,3 +604,75 @@ class BucketingSampler(CutSampler):
             enumerate(zip(self.bucket_samplers, self.depleted))
             if not depleted
         ]
+
+
+class ZipSampler(CutSampler):
+    """
+    :class:`.ZipSampler` takes several samplers as input and concatenates their
+    sampled batch cut IDs together into a single list.
+    It is helpful for ensuring that each batch consists of some proportion of cuts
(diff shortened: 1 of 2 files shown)


lhotse/dataset/sampling.py
```diff
@@ -604,3 +633,75 @@ def _nondepleted_samplers_with_idxs(self):
...
+class ZipSampler(CutSampler):
+    """
+    :class:`.ZipSampler` takes several samplers as input and concatenates their
+    sampled batch cut IDs together into a single list.
+    It is helpful for ensuring that each batch consists of some proportion of cuts
+    coming from different sources.
+
+    The input samplers do not have to provide the same number of batches -- when
+    any of the samplers becomes depleted, the iteration will stop (like with
+    Python's ``zip()`` function).
```
danpovey, line 647: This is nice.  Yes, I think it's OK.

danpovey (approved):

PR #345, opened 2021-07-27
Overhaul of Dataset API

pzelasko:
Two changes, really:
- samplers return `CutSet` instead of `List[str]` of cut IDs; this makes lazy manifest support much easier (dataset class doesn't have to lookup IDs with random access requirements)
- remove apache arrow and rely on sequential reads of JSONL instead (it's much faster and much simpler...)

I will make a parallel snowfall PR to ensure it's compatible with the changes.

### DIFF
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -96,14 +96,14 @@ class K2SpeechRecognitionDataset(torch.utils.data.Dataset):
     def __len__(self) -> int:
         return len(self.cuts)
 
-    def __getitem__(self, cut_ids: List[str]) -> Dict[str, Union[torch.Tensor, List[str]]]:
+    def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List[str]]]:
         """
         Return a new batch, with the batch size automatically determined using the contraints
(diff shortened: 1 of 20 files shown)


lhotse/dataset/speech_recognition.py
```diff
@@ -80,30 +77,20 @@ def __init__(
             Examples: normalization, SpecAugment, etc.
         :param input_strategy: Converts cuts into a collated batch of audio/features.
             By default, reads pre-computed features from disk.
-        :param check_inputs: Should we iterate over ``cuts`` to validate them.
-            You might want to disable it when using "lazy" CutSets to avoid a very long start up time.
         """
         super().__init__()
         # Initialize the fields
-        self.cuts = cuts
```
csukuangfj, line 88: I am trying to print the decoding progress using
`batch_idx / len(self.cuts) * 100`

But `self.cuts` is removed in this pull-request. Is there a way to get the total number of cuts in a `K2SpeechRecognitionDataset`?

pzelasko: In specific cases, yes. You can check in the sampler, but different samplers would need to be checked differently. I’ll see about adding a “num_cuts” attribute.

PR #352, opened 2021-08-03
Padding features instead of audio in OnTheFlyFeatures

### LINKED ISSUE #349: Unnecessary padding to audio samples in OnTheFlyFeatures
See the code below
https://github.com/lhotse-speech/lhotse/blob/81b3a0d842787ec476ead2c255e5be0ee89eb0cb/lhotse/dataset/input_strategies.py#L233

https://github.com/lhotse-speech/lhotse/blob/81b3a0d842787ec476ead2c255e5be0ee89eb0cb/lhotse/dataset/input_strategies.py#L239-L242

The extra padding has the following side effects:
 - (1) Waste CPU during feature extraction.
 - (2) It affects the result of the last feature frame. That means the features of an utterance vary
with respect to the batch size (`--max-duration`).

pzelasko:
Resolves #349 @csukuangfj 

I haven't tested it yet besides running tests locally. I'll need to double-check on real data that there are no substantial differences besides the last frame thing.

### DIFF
--- a/lhotse/dataset/collation.py
+++ b/lhotse/dataset/collation.py
@@ -276,6 +276,20 @@ Helper functions to dispatch jobs to the concurrent executors.
 """
 
 
+def read_audio_from_cuts(
+        cuts: Iterable[Cut], executor: Optional[Executor] = None
+) -> List[torch.Tensor]:
+    map_fn = map if executor is None else executor.map
+    return list(map_fn(_read_audio, cuts))
+
+
+def read_features_from_cuts(
+        cuts: Iterable[Cut], executor: Optional[Executor] = None
+) -> List[torch.Tensor]:
+    map_fn = map if executor is None else executor.map
+    return list(map_fn(_read_features, cuts))
+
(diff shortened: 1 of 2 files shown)


pzelasko:
I think it's OK, the only difference is in the padding values: when we feed zeroes into the feature extractor, the features have log energy values of -15.9424; when we pad directly, the default value in Lhotse is log(1e-10) (-23.0259). I thought that the feature extractor would also use log(1e-10) given `energy_floor=1e-10`, but it turns out that's only true for the extra feature dimension you get with `use_energy=True`. I'm not sure if this discrepancy would really affect anything, so I'm going to merge, and we can always adjust that later if needed.

pzelasko:
Also: on a random single batch (~44 cuts / ~300s of speech) from librispeech, the speedup is about 3x: ~800ms (old version) vs ~280ms (new version), which is pretty cool. Probably with bucketing there would be less difference, but without it it's very useful.

PR #355, opened 2021-08-04
Avoid extra work in compute_and_store_recording + sensible default for open_writer()

pzelasko:


pzelasko:
also an update to docs: API reference shows inherited members as well

Issue #356, opened 2021-08-05
The loss value oscillates at the beginning of each epoch


pkufool:
I am training a model based on this PR https://github.com/k2-fsa/snowfall/pull/219, as Dan's suggestion, I removed the normalization in ctc_loss and attn_loss, and also the clip_grad_norm_ line.

The loss curve is as follows, the value oscillates at the beginning of each epoch:
|total loss|attn_loss|ctc_loss|
|--|--|--|
|![image](https://user-images.githubusercontent.com/11765074/128289888-d41c06fa-7bd7-4dab-9a07-57f4a11a1d70.png)|![image](https://user-images.githubusercontent.com/11765074/128289913-379213e6-bfb6-480f-b91c-9d3bdea556f6.png)|![image](https://user-images.githubusercontent.com/11765074/128289962-a65ec62c-590c-427e-acf8-e6dc3539c23f.png)|

**Note:** I forgot to update the code displaying loss value, it should be divided by batch size.

I think it might be relevant to the sampler method, @csukuangfj  is trying another sampler, will update the result latter.

danpovey:
BTW, I believe this code has num-buckets = 1000, which may be a bit too
high; I think this is about half-full buckets towards the end of an epoch.
Perhaps the bucketing sampler could start combining adjacent buckets once
we get down to partial minibatches?

pzelasko:
Could the oscillation be a „natural” training property when the batches vary a lot in terms of batch size (number of sequences) and sequence lengths? I am not sure what could be done within sampler to change that.

As to the bucketing sampler, combining adjacent buckets seems fairly complex implementation-wise, is it really worth it rather than simply specifying a smaller num-buckets? 

BTW at some point I am planning to add a different version of bucketing that is more usable with larger data (>1k hours) and wouldn’t need to pre-load all cuts into memory. The bucketing would be done as a separate step before training, and each bucket would be stored in a separate cut set manifest. Then, the bucketing samplers would be actually sth like: `sampler = RoundRobinSampler([SingleCutSampler(p, max_duration=300) for p in cut_buckets_path])`

danpovey:
... BTW, perhaps we could have some diagnostic about how much data was
discarded due to partial buckets, and maybe have an option to just discard
data when we'd otherwise be forming a partial bucket?
My concern is that if, at the end of the epoch, the buckets are, on
average, half-empty, we'll get more gradient noise at the end of the epoch
(at least, with Adam, because it will effectively increase the learning
[...]

PR #357, opened 2021-08-05
Samplers: `drop_last` option and `get_report()` diagnostic method

pzelasko:
Following up on the discussion in #356. Here are some examples of usage of `drop_last`, and how the sampler report looks like for dev-clean-2 split of mini_librispeech:

image

@danpovey @pkufool


danpovey:
Great!  @pkufool let's do a run with this, I think it will make a
difference.

Issue #364, opened 2021-08-10
BucketingSampler more randomness?


danpovey:
@pzelasko we are still seeing sawtooth patterns in losses when we use BucketingSampler, even with fewer buckets.
I think it's because the individual buckets are sorted by length.  Is it possible to shuffle somehow within the buckets, or, say, always randomly pick a batch from the front or back of the bucket?  Or perhaps the individual buckets could either be reversed, or not-reversed, randomly or alternately.

csukuangfj:
https://tensorboard.dev/experiment/lFXXFnRvSKK6laA0lBlv8w/#scalars&_smoothingWeight=0
The above tensorboard log is using bucketing sampler with the default number of buckets (30).
You can see that the attention loss jumps at the beginning of each epoch.

pzelasko:
OK, just a sanity check — did you pass in the arguments `shuffle=True` to the constructor, and called `sampler.set_epoch(epoch)` at each epoch?

csukuangfj:
> did you pass in the arguments shuffle=True to the constructor, and called sampler.set_epoch(epoch) at each epoch?

Yes. The related code is shown below:

https://github.com/k2-fsa/icefall/blob/master/icefall/dataset/asr_datamodule.py#L172
```python
        if self.args.bucketing_sampler:
            logging.info("Using BucketingSampler.")
            train_sampler = BucketingSampler(
                cuts_train,
                max_duration=self.args.max_duration,
                shuffle=True,
                num_buckets=self.args.num_buckets,
            )
```

https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/conformer_ctc/train.py#L639
```python
    for epoch in range(params.start_epoch, params.num_epochs):
        train_dl.sampler.set_epoch(epoch)
```

pzelasko:
In that case, I will double-check if the shuffling works as intended in each bucket of the BucketingSampler. 

BTW I didn’t notice the `drop_last=True` option for BucketingSampler in what you posted — was it used?

danpovey:
There is another factor that could cause this sawtooth pattern, which is that the buckets have the same number of samples, but you consume them in batches that depend on the duration, so the shorter buckets will be consumed first.  It might be better to compute the cumulative sum of the durations, and split at percentiles of that, if that is not already what split() there does.

pzelasko:
It doesn’t do it. I can make a PR with that option later and let’s see then.

csukuangfj:
[...]

PR #365, opened 2021-08-10
Add a bucketing method with equal cumulative bucket duration

pzelasko:
Hopefully resolves the issue #364 

@danpovey @csukuangfj please check this out and LMK if it helps with the sawtoothed loss

To use it, add the argument:
```python
sampler = BucketingSampler(
   ...,
   bucket_method='equal_duration',
)
```


pzelasko:
I also double-checked that shuffling works as intended in the BucketingSampler.

csukuangfj:
Thanks! Will try it and post the result here.

PR #367, opened 2021-08-12
add chunked sph file processing

videodanchik:
This MR aims to provide an optimal way for reading SPHERE files, avoiding reading the whole file, when offset and duration are provided. These parameters can be specified directly in `sph2pipe` command which helps to load only chunk of audio in memory during `lhotse.audio.read_audio` function call.


lhotse/audio.py
```diff
@@ -15,6 +15,7 @@
 from typing import Any, Callable, Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence, Tuple, Union
 
 import numpy as np
+import soundfile as sf
```
pzelasko, line 18: Please stick to a local import, otherwise building the docs is going to fail (the import tries to load libsndfile.so into memory which is not available on read-the-docs servers).

lhotse/audio.py
```diff
@@ -231,8 +232,7 @@ def from_file(
         else:
             try:
                 # Try to parse the file using pysoundfile first.
-                import soundfile
-                info = soundfile.info(str(path))
+                info = sf.info(str(path))
```
pzelasko, line 235: is there a missing case for using `sph_info`?

lhotse/audio.py
```diff
@@ -1102,3 +1107,44 @@ def parse_channel_from_ffmpeg_output(ffmpeg_stderr: bytes) -> str:
...
+    # Construct the sph2pipe command depending on the arguments passed.
+    cmd = f'sph2pipe -f wav -p -t {offset}:'
+
+    if duration is not None:
+        cmd += f'{round(offset + duration, 5)}'
+    # Add the input specifier after offset and duration.
+    cmd += f' {sph_path}'
+
+    # Actual audio reading.
+    proc = BytesIO(run(cmd, shell=True, stdout=PIPE, stderr=PIPE).stdout)
```
pzelasko, line 1145: This makes sure the failures are propagated. 

```suggestion
    proc = BytesIO(run(cmd, shell=True, check=True, stdout=PIPE, stderr=PIPE).stdout)
```

pzelasko:
Thanks that looks good! Please address my comments and we'll merge it.

lhotse/audio.py
videodanchik: Right, good catch.

videodanchik:
Can we merge this as I want to add Fisher English that will depend on this PR?

pzelasko:
Hmm, the tests are failing because there is no sph2pipe in the CI. Let me try to add it, and then we can merge.

pzelasko:
@videodanchik I made a PR #370 that makes it easy to install sph2pipe in a way that enables Lhotse to auto-discover it. I'll wait for Dan to say what he thinks; as soon as it's merged, we can try re-run the CI for your PR again with no code changes -- it should help.

videodanchik:
@pzelasko Oh, sorry that it caused this additional effort, I haven't even realized that you can't install `sph2pipe` via `sudo apt install`.

pzelasko:
no worries, I think that it’s better for Lhotse users not to have to worry about how to find these esoteric programs…

pzelasko:
Seems to work now, merging.

PR #370, opened 2021-08-13
Utilities to install sph2pipe

pzelasko:


pzelasko:
@danpovey I'd like to get your feedback if you think this way of handling external programs is acceptable. This code downloads + compiles sph2pipe and appends it to $PATH at the time when `import lhotse` is called. The spawned subprocesses inherit $PATH so they can discover sph2pipe. By default it is installed in `~/.lhotse/tools` (if it's installed somewhere else it's up to the user to set the $PATH properly).

danpovey (commented):
LGTM!

PR #377, opened 2021-08-16
fix dev requires in setup.py

songmeixu:
To solve the issue:
The dev_requires wasn't used. And won't be install through `pip install -e '.[dev]'`.


pzelasko:
Thanks for spotting this!

Issue #378, opened 2021-08-19
Adding alignment constraints to supervision


danpovey:
Did you end up uploading a version of alignments for Librispeech training data, and can you recommend
how one might go about preparing these somehow for use in training?
I'm thinking that for Icefall we can start training with some alignment constraints for Librispeech.

For @csukuangfj and @pkufool: I think I figured out how we can apply alignment constraints without decoder
changes, using mechanisms that already exist in k2.  (Of course we can add a more efficient mechanism later
if we want).  Firstly: intersect_dense already supports the option `frame_idx_name`, where you can specify the 
name of an attribute corresponding to the frame-index, e.g. `frame_idx_name='frame_idx'`

We could also attach an attribute to the original FSA containing word sequences, the attribute being the time
of the word within the segment.  Let's suppose that we convert this to frames after subsampling, and let the corresponding
`filler` be -1.  (Can do this by setting, say, fsa.times = [tensor of times], and fsa.times_filler = -1, suppposing
fsa.times contained int32).  This will mean that when we construct the graph, it will have -1 for arcs that
don't have a word on.  (Might require expand_ragged_attributes()).  Once we attach the frame_idx to the
lattice, we can modify the lattice's scores via a torch expression that compares the frame_idx to the time of the
word as recorded in the attribute, possibly with some kind of collar to avoid constraining it too tightly.
We can compare as signed for one of the directions of comparison, and as unsigned for the other direction, 
which will give us the correct behavior for arcs that don't have a word label (i.e., like the ace in cards, the -1 can
be high or low).

We could add, say, -10 to arcs where the time is out of bounds.   That should be more than enough to get training
started.
[Or we can do 2 separate comparisons with 2 separate times, one for the beginning and one for the end of the word,
again possibly with a small collar.]

csukuangfj:
This repo
https://github.com/CorentinJ/librispeech-alignments

contains word alignment information. It contains the timestamp about the beginning and ending of each word\
within an utterance. Silences are also included.

Shall we make it available in `K2SpeechRecognitionDataset`?

csukuangfj:
> We could add, say, -10 to arcs where the time is out of bounds

Does -10 mean some`log_prob` here and is the purpose to make sure frames corresponding to silences have a lower probability?

danpovey:
Yes, that's what I mean about the -10.
Sure, I think using that repo is a good idea, if it seems easy. 
[...]

PR #379, opened 2021-08-20
Adding LibriSpeech word alignments in supervisions

pzelasko:


pzelasko:
Note: silences are represented as an empty string

pzelasko:
Adding the alignments in K2 dataset seems fairly easy; the "supervisions" dict has three more keys, "word", "word_start", "word_end"; it's list of lists of str/float.

@danpovey before I merge let me know if this format works for you guys (don't mind my transparent terminal with Spotify in the background).

image
image
image

csukuangfj:
> it's list of lists of str/float.

Would it be easier for later use if it returns frames for word_start and word_end, i.e., use int32_t ?

pzelasko:
Up to you guys. I don’t have a good idea of how you want to use it atm. Do you prefer that to be in frames?

csukuangfj:
> Up to you guys. I don’t have a good idea of how you want to use it atm. Do you prefer that to be in frames?

From
https://github.com/lhotse-speech/lhotse/issues/378#issue-974462075

>  Let's suppose that we convert this to frames after subsampling, and let the corresponding
filler be -1. (Can do this by setting, say, fsa.times = [tensor of times], and fsa.times_filler = -1, suppposing
fsa.times contained int32)

It says we need start/end frames.

But from https://github.com/lhotse-speech/lhotse/issues/378#issuecomment-901981386

> It would be easiest to use, I think, if the words had a 'begin_frame' and
'end_frame'  (or just a single frame index) and these were prepared with
the same shape as the words themselves-- not sure if it becomes a list of
list of int at some point?

> I assume that they'd be floating point times in seconds at the point we get
them from lhotse, as we need to set the frame rate.

It suggests using times in seconds.   

---

I am not sure which one is better.

danpovey:
Seconds is OK, it's best if the calling code converts that to frames because the calling code knows the frame rate.
I think this should be OK.

danpovey:
... BTW, part of the reason I want this to be in integers is that k2 basically assumes that floating-point attributes are "score-like", so for instance they will be added together when integer attributes would be converted to ragged, such as when removing epsilons; and the default value can only be 0, never -1.  Later we can change this behavior if it becomes a problem.

pzelasko:
I think the calling code doesn’t know the frame shift anymore (unless you are using precomputed features and use dataset with return_cuts=True so you can query the cuts, but then it will fail with on the fly features). Also we are already returning start frame and num frames for each supervision from the dataset, so this is inconsistent. I’d suggest using frames here after all, unless you’re sure about seconds.

danpovey:
ok.. frames is Ok.

pzelasko:
Let's see if this is better, if it's OK I'm going to merge (can't thoroughly test it right now but seems fine on isolated examples -- I plan to clean it up and add some tests later)

danpovey:
Thanks!!

csukuangfj:
+2

PR #382, opened 2021-08-22
Feature/volume perturbation

videodanchik:
Add volume perturbation

### DIFF
--- a/lhotse/dataset/cut_transforms/perturb_vol.py
+++ b/lhotse/dataset/cut_transforms/perturb_vol.py
@@ -1,0 +1,34 @@
+import random
+from typing import List, Sequence, Union
+
+from lhotse import CutSet
+
+
+class PerturbVol:
+    """
+    A transform on batch of cuts (``CutSet``) that perturbs the volume of the recordings
+    with a given probability :attr:`p`.
+
+    If the effect is applied, then one of the perturbation factors from the constructor's
+    :attr:`factors` parameter is sampled with uniform probability.
+    """
+
+    def __init__(
+            self,
+            factors: Union[float, List[float]],
+            p: float,
+            randgen: random.Random = None
+    ) -> None:
+        self.factors = factors if isinstance(factors, Sequence) else [factors]
+        self.p = p
+        self.random = randgen
+
+    def __call__(self, cuts: CutSet) -> CutSet:
+        if self.random is None:
+            self.random = random
+        return CutSet.from_cuts(
+            cut.perturb_vol(factor=self.random.choice(self.factors))
+            if self.random.random() >= self.p
+            else cut
+            for cut in cuts
+        )
--- a/lhotse/dataset/cut_transforms/perturb_vol.py
+++ b/lhotse/dataset/cut_transforms/perturb_vol.py
@@ -15,7 +15,7 @@ class PerturbVol:
 
     def __init__(
             self,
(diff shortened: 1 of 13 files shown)


lhotse/dataset/cut_transforms/perturb_vol.py
```diff
@@ -0,0 +1,34 @@
...
+    A transform on batch of cuts (``CutSet``) that perturbs the volume of the recordings
+    with a given probability :attr:`p`.
+
+    If the effect is applied, then one of the perturbation factors from the constructor's
+    :attr:`factors` parameter is sampled with uniform probability.
+    """
+
+    def __init__(
+            self,
+            factors: Union[float, List[float]],
```
pzelasko, line 18:
```suggestion
            factors: Union[float, Sequence[float]],
```
This allows tuples without type checker conflicts

test/augmentation/test_torchaudio.py
```diff
@@ -36,33 +43,66 @@ def test_speed_does_not_change_num_samples(audio):
...
+    audio_perturbed = vol(audio_vol, SAMPLING_RATE)
+    
+    assert audio_perturbed.shape == audio_vol.shape
+    assert_array_almost_equal(audio_perturbed, scale * audio_vol)
 
 
 def test_deserialize_transform(audio):
     speed = AudioTransform.from_dict({'name': 'Speed', 'kwargs': {'factor': 1.1}})
-    perturbed = speed(audio, SAMPLING_RATE)
-    assert perturbed.shape == (1, 14545)
```
pzelasko, line 50: These changes would ideally be introduced in a copy of this unit test (my principle is one logical assertion == one unit test, to avoid tests that check independent functionalities because they are confusing to readers and harder to debug)

test/augmentation/test_torchaudio.py
```diff
@@ -36,33 +43,66 @@ def test_speed_does_not_change_num_samples(audio):
...
+    assert_array_almost_equal(perturbed_vol, audio / 2.)
 
 
 def test_serialize_deserialize_transform(audio):
     speed_orig = Speed(factor=1.1)
-    data = speed_orig.to_dict()
-    speed = AudioTransform.from_dict(data)
-    perturbed = speed(audio, SAMPLING_RATE)
-    assert perturbed.shape == (1, 14545)
+    vol_orig = Vol(factor=0.5)
```
pzelasko, line 86: Same comment as above

pzelasko (commented):
Wow great work and nice test coverage! I left a couple of comments. I’m also thinking that naming it perturb_volume rather than perturb_vol would be more consistent with other Lhotse names (we have usually avoided abbreviations in APIs so far).

videodanchik:
I split tests and replaced `vol` with `volume` everywhere. Also did replacement you suggested: `List` -> `Sequence` in `lhotse/dataset/cut_transforms/perturb_volume.py` and `lhotse/dataset/cut_transforms/perturb_speed.py` (where I copied it from). And finally replaced `recording_id=f'{self.recording_id}_vp{factor}' if affix_id else self.id` with `recording_id=f'{self.recording_id}_vp{factor}' if affix_id else self.recording_id` in perturb_speed/tempo/volume as I believe it was a typos (I made it with a separate commit, so this can be easily reverted).

pzelasko:
Oh yeah that's a good catch. Thanks a lot!

PR #383, opened 2021-08-23
Change how lhotse entry point is registered in setuptools

pzelasko:
@csukuangfj please check if this fixes your issues with shebangs. Before it's merged into master, it can be either installed via git clone + branch checkout, or `pip install git+https://github.com/lhotse-speech/lhotse.git@feature/setuptools-tweaks`

I reproduced the error and this seems to fix it in colab for me.


csukuangfj:
Yes, I confirm that it works both for colab and for my local virtual environment.

PR #385, opened 2021-08-25
Versioning updates + release v0.8

pzelasko:


lhotse/__init__.py
```diff
@@ -17,6 +17,13 @@
 from .tools.env import add_tools_to_path as _add_tools_to_path
 from .qa import validate, validate_recordings_and_supervisions, fix_manifests
 
+try:
+    # Try to get Lhotse's version (should be created during running pip install / python setup.py ...)
+    from .version import __version__
+except:
+    # Use a default placeholder when the version is unavailable...
+    __version__ = '0.0.0.dev'
```
csukuangfj, line 25: Can we change `0.0.0.dev` to the values used in `setup.py`, i.e., change
it to `0.9.0.dev` since the current values are
```python
MAJOR_VERSION=0
MINOR_VERSION=9
IS_DEV_VERSION=True
```

I have been using `lhotse` without running `python setup.py install` or `pip install lhotse`.
I just modify my `PYTHONPATH` environment variable to use `lhotse`.

pzelasko: How about you run:

python setup.py sdist bdist_wheel

Which won’t install it but will create the version file? You can re-build every time you update.

danpovey: Surely it wouldn't be too much effort to update the setup.py every time the version is updated, for slightly more convenience?

csukuangfj: > How about you run:
> 
> python setup.py sdist bdist_wheel
> 
> Which won’t install it but will create the version file? You can re-build every time you update.

Yes, I think `python setup.py sdist` will work. Would be great if you can change `__init__.py`
when you do a new release. You need to change `setup.py` at each new release. I think it does not take
too much effort to edit `__int__.py` at the same time.

pzelasko: OK, I'll do it -- I have trouble remembering things like this though so please ping me if I forget to update either place. I might need to write some release instructions for myself or add some code to check the version consistency in `tools/make_release.sh`...

csukuangfj: I would place a note at
https://github.com/lhotse-speech/lhotse/blob/6c5fc211847e166a75727c725d5644225e6ca5b8/setup.py#L11
saying

```
Please remember to also update the version in `lhotse/__init__.py`
```

PR #405, opened 2021-09-27
Options to preserve cut IDs in cut operations and transforms

pzelasko:
All relevant cut operations (`.pad()`, `.truncate()`, `.mix()`, `.append()` and transforms (`PerturbSpeed`, `CutMix`, etc.) now have a `preserve_id` option that is set to ``False`` by default. Setting it to ``True`` allows to augment cuts and still match them by IDs.

CC: @csukuangfj

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -211,19 +211,33 @@ class Cut:
         """
         return [s.trim(self.duration) for s in self.supervisions]
 
-    def mix(self, other: 'Cut', offset_other_by: Seconds = 0.0, snr: Optional[Decibels] = None) -> 'MixedCut':
+    def mix(
+        self,
+        other: "Cut",
+        offset_other_by: Seconds = 0.0,
+        snr: Optional[Decibels] = None,
+        preserve_id: Optional[str] = None,
+    ) -> "MixedCut":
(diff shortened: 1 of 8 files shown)


lhotse/cut.py
```diff
@@ -211,19 +211,33 @@ def trimmed_supervisions(self) -> List[SupervisionSegment]:
...
         """Refer to :function:`~lhotse.cut.mix` documentation."""
-        return mix(self, other, offset=offset_other_by, snr=snr)
+        return mix(self, other, offset=offset_other_by, snr=snr, preserve_id=preserve_id)
 
-    def append(self, other: 'Cut', snr: Optional[Decibels] = None) -> 'MixedCut':
+    def append(
+        self,
+        other: "Cut",
+        snr: Optional[Decibels] = None,
+        preserve_id: Optional[str] = None,
```
csukuangfj, line 228: Should the type be `bool`, not `Optional[str]`?

csukuangfj:
Thanks!

lhotse/cut.py
pzelasko: Yeah for append and mix it allows you to choose which cuts ID to preserve (useful in some applications eg padding from both sides)

PR #414, opened 2021-10-05
Add torchaudio.load()/info() as audio backends + reduce the memory usage of CommonVoice data prep.

pzelasko:
The MP3 audio reads and info gathering is much faster with torchaudio than with audioread. CommonVoice manifests are written in a streaming way (and can be resumed) which significantly reduces memory usage. Might add parallel processing later.


pzelasko:
@mthrok I wonder why the result of `torchaudio.info()` seems to always return `num_frames` that is 1152 larger than the actual number of samples obtained via `torchaudio.load()` for mp3 files? I observe this behavior consistently in CommonVoice, e.g.:

```python
In [23]: torchaudio.info('/data/cv-corpus-7.0-2021-07-21/en/clips/common_voice_en_651325.mp3').num_frames
Out[23]: 114048

In [24]: torchaudio.load('/data/cv-corpus-7.0-2021-07-21/en/clips/common_voice_en_651325.mp3')[0].shape
Out[24]: torch.Size([1, 112896])
```

mthrok:
> @mthrok I wonder why the result of `torchaudio.info()` seems to always return `num_frames` that is 1152 larger than the actual number of samples obtained via `torchaudio.load()` for mp3 files? I observe this behavior consistently in CommonVoice, e.g.:
> 
> ```python
> In [23]: torchaudio.info('/data/cv-corpus-7.0-2021-07-21/en/clips/common_voice_en_651325.mp3').num_frames
> Out[23]: 114048
> 
> In [24]: torchaudio.load('/data/cv-corpus-7.0-2021-07-21/en/clips/common_voice_en_651325.mp3')[0].shape
> Out[24]: torch.Size([1, 112896])
> ```

It's probably the padding that `libmad` performs. The issue has been flagged [here](https://github.com/pytorch/audio/issues/1500) but I have not looked into it yet.

mthrok:
Sure. Sorry for the inconvenience. I will try to look into this after the upcoming release.

mthrok:
@pzelasko In torchaudio v0.12, we are switching MP3 decoding to FFmpeg-based implementation and as a result, the the tensor is no longer padded. The caveat is that now we ask users to install FFmpeg.

pzelasko:
Very cool, looking forward to trying it out.

mthrok:
> I think Lhotse is already kinda depending on ffmpeg anyway, so that should be fine. Do you call ffmpeg as a process or do you use some python api? y.
> […](#)

We use C API.

Code can be found here. https://github.com/pytorch/audio/tree/main/torchaudio/csrc/ffmpeg

The change about MP3 is done behind the scene, so you do not need to do anything, but now we have new APIs, which are explained 
- https://pytorch.org/audio/main/tutorials/streaming_api_tutorial.html
- https://pytorch.org/audio/main/tutorials/streaming_api2_tutorial.html
- https://pytorch.org/audio/main/hw_acceleration_tutorial.html
- https://pytorch.org/audio/main/tutorials/online_asr_tutorial.html
- https://pytorch.org/audio/main/tutorials/device_asr.html

PR #417, opened 2021-10-11
Saving and restoring the sampler's state to continue training where it left off

pzelasko:
I'm still testing it with the dataloaders to make sure it works all right. I'll write some usage docs later.

### DIFF
--- a/lhotse/dataset/sampling/base.py
+++ b/lhotse/dataset/sampling/base.py
@@ -125,7 +125,7 @@ class CutSampler(Sampler):
         """
         Return the current state of the sampler in a state_dict.
         Together with ``load_state_dict()``, this can be used to restore the
-        training loop's state to start where it currently is.
+        training loop's state to the one stored in the state_dict.
         """
         return {
             'epoch': self.epoch,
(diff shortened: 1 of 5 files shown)


pzelasko:
The intended use for this PR's features is when we stopped a training run, and want to resume exactly where it left off rather than at the start of the epoch. This is especially useful for **very large training sets**, where checkpointing on the epoch end is "dangerous" (e.g. because our job gets killed at 60% progress and we lose half a day of computations).

Suggested usage:

```python
     dataset = ...  # Some task-specific dataset initialization
     sampler = BucketingSampler(cuts, max_duration=200, shuffle=True, num_buckets=30)

     # Restore the sampler's state.
     state = torch.load('sampler-ckpt-ep5-step75000.pt')
     sampler.load_state_dict(state)

     dloader = DataLoader(dataset, batch_size=None, sampler=sampler, num_workers=4)

     global_step = sampler.diagnostics.total_cuts  # <-- Restore the global step idx.
     for epoch in range(sampler.epoch, 30):  # <-- Skip previous epochs that are already processed.

         dloader.sampler.set_epoch(epoch)
         for batch in dloader:
             # Note: the first batch is going to be from step 75009.
             # With DataLoader num_workers==0, it would have been 75001, but we get
             # +8 because of num_workers==4 * prefetching_factor==2

             # forward, backward, etc.

             global_step += 1

             if global_step % 5000 == 0:
                 state = dloader.sampler.state_dict()
                 torch.save(state, f'sampler-ckpt-ep{epoch}-step{global_step}.pt')
```

In the example I'm highlighting a caveat: when using DataLoader with num_workers > 1, then it is going to prefetch some batches, meaning that sampler's state will be for `num_workers * prefetching_factor` steps ahead than the training loop. This shouldn't be an issue in practice though.

PR #423, opened 2021-10-15
Add and update timit recipe

luomingshuang:
I add some scripts for timit recipe to adapt the new version lhotse. And I have checked and run the added scripts well.


pzelasko:
Thanks!!

PR #438, opened 2021-10-29
WIP support exporting multichannel audio into kaldi

jtrmal:
I was thinking about this -- let me know what you think.

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -108,7 +108,7 @@ def load_kaldi_data_dir(
     segments = path / "segments"
     if segments.is_file():
         with segments.open() as f:
-            supervision_segments = [l.strip().split() for l in f]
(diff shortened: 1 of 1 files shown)


lhotse/kaldi.py
```diff
@@ -292,8 +296,27 @@ def make_wavscp_string(source: AudioSource, sampling_rate: int) -> str:
         return f"{source.source} |"
     elif source.type == "file":
         if Path(source.source).suffix == ".wav":
-            return source.source
+            audios = dict()
```
pzelasko, line 299: Yeah that looks good to me! Can you change the return type of `make_wavscp_string` to `Dict[int, str]` (and maybe rename to sth like `make_channel_to_wavscp_string_map` or less verbose if you have a better idea).

jtrmal:
Yeah both good points. Thanks
Y.

PR #441, opened 2021-11-02
add export multichannel to kaldi

jtrmal:
similar to the POC code but now in the correct branch and resolved some minuscule changes

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -35,7 +35,8 @@ def get_duration(
             )
         from kaldiio import load_mat
 
-        # Note: kaldiio.load_mat returns (sampling_rate: int, samples: 1-D np.array[int])
+        # Note: kaldiio.load_mat returns
+        # (sampling_rate: int, samples: 1-D np.array[int])
         sampling_rate, samples = load_mat(path)
         assert len(samples.shape) == 1
         duration = samples.shape[0] / sampling_rate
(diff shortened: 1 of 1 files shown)


lhotse/kaldi.py
```diff
@@ -207,26 +209,30 @@ def export_to_kaldi(
...
+    # the recording <-> supervision information.
     cuts = CutSet.from_manifests(
         recordings=recordings, supervisions=supervisions
     ).trim_to_supervisions()
 
     # wav.scp
     save_kaldi_text_mapping(
         data={
-            recording.id: make_wavscp_string(
+            f"{recording.id}_{channel}": make_wavscp_channel_string_map(
```
pzelasko, line 221: I didn’t notice that before but this is possibly breaking existing workflows where data is exported to Kaldi for processing and then imported back to Lhotse (recording ids are changing). I am aware of such use cases. Could you avoid modifying recording ID for single channel recordings?

lhotse/kaldi.py
```diff
@@ -207,26 +209,30 @@ def export_to_kaldi(
...
         },
         path=output_dir / "wav.scp",
     )
     # segments
     save_kaldi_text_mapping(
         data={
-            cut.supervisions[0].id: f"{cut.recording_id} {cut.start} {cut.end}"
+            cut.supervisions[
+                0
+            ].id: f"{cut.recording_id}_{cut.channel} {cut.start} {cut.end}"
```
pzelasko, line 235: Same as above

lhotse/kaldi.py
```diff
@@ -248,7 +254,11 @@ def export_to_kaldi(
     )
     # reco2dur
     save_kaldi_text_mapping(
-        data={recording.id: recording.duration for recording in recordings},
+        data={
+            f"{recording.id}_{channel}": recording.duration
```
pzelasko, line 258: And here

lhotse/kaldi.py
```diff
@@ -248,7 +254,11 @@ def export_to_kaldi(
...
+            for recording in recordings
+            for channel in recording.sources[0].channels
```
pzelasko, line 260: Should probably iterate over sources here too?

jtrmal: I think we have an assert in the beginning to allow only single-source recordings, no?

lhotse/kaldi.py
jtrmal: will do. Will you take "if" instead of making the list comprehension even more complex? :)

pzelasko: Yeah. I think at some point this module might need unit tests, it’s definitely growing in complexity. But don’t worry about it for now.

lhotse/kaldi.py
pzelasko: Oh I see. I think your changes are very close to supporting that but it’s fine to leave as-is.

lhotse/kaldi.py
```diff
@@ -194,9 +196,9 @@ def export_to_kaldi(
         "Kaldi export of Recordings with multiple audio sources "
         "is currently not supported."
     )
-    assert all(r.num_channels == 1 for r in recordings), (
-        "Kaldi export of multi-channel Recordings is currently " "not supported."
-    )
+    # assert all(r.num_channels == 1 for r in recordings), (
+    #     "Kaldi export of multi-channel Recordings is currently " "not supported."
+    # )
```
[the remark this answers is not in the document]
jtrmal: removed. Sorry

jtrmal:
Actually, I just realized there has to be change to the way how the kaldi wav.scp strings are generated if we want to support full translation identity

jtrmal:
Well, let me know what you think -- it actually relates only to the .sph files

pzelasko:
I'm OK with merging as-is, if we need extra Kaldi compatibility we can follow it up later. Is it OK to merge or did you want to add anything else?

pzelasko:
Thank you, sir!

PR #442, opened 2021-11-02
make the callhome asr utterances following specific format for QOL

jtrmal:
make the callhome asr utterances following specific format for ease of export to kaldi (quality of life change)

specifically, quality of my life :)
but shouldn't be a problem to anyone, callhome english ASR is very new in lhotse
y.


pzelasko:
Thanks!

PR #447, opened 2021-11-03
Fixes to babel recipe

jtrmal:
Can you please check if this is the correct fix or do you suggest something else?
The main reason was this didn't work:
```
lhotse prepare babel  /export/corpora5/babel_CLSP/104-pashto/release-babel104b-v04.aY  ./babel/
```
so I wonder if those two variables should be converted somewhere in the command line interfacing code (like `./lhotse/bin/modes/recipes/babel.py` e.g.)

### DIFF
--- a/lhotse/recipes/babel.py
+++ b/lhotse/recipes/babel.py
@@ -9,6 +9,7 @@ import logging
 import re
 from collections import defaultdict
 from typing import Dict, Iterable, List, Optional, Union
+from pathlib import Path
 
 from cytoolz import sliding_window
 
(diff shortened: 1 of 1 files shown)


pzelasko:
I think it's my omission that I didn't add this conversion before (for user convenience, I generally assume in Lhotse that path inputs work with regular strings, and always convert them manually to `pathlib.Path`)

pzelasko:
Does it work now?

jtrmal:
yeah, all files are generated after these fixes.
y.

pzelasko:
Thanks!

PR #449, opened 2021-11-04
Fixed bug in preparing supervisions

desh2608:
@HuangZiliAndy thanks for pointing out the issue

**NOTE:** 5 supervision segments exceed the respective recording bounds and are filtered in the preparation.

### DIFF
--- a/lhotse/recipes/ami.py
+++ b/lhotse/recipes/ami.py
@@ -477,7 +477,9 @@ def prepare_supervision_ihm(
 def prepare_supervision_other(
     audio: RecordingSet, annotations: Dict[str, List[AmiSegmentAnnotation]]
 ) -> SupervisionSet:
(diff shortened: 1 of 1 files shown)


lhotse/recipes/ami.py
```diff
@@ -588,6 +590,16 @@ def prepare_ami(
         else prepare_supervision_other(audio, annotations)
     )
 
+    # Some supervision segments exceed the recording bounds, so we remove them here
```
pzelasko, line 593: Instead of doing this manually, you can run `recordings, supervisions = fix_manifests(recordings, supervisions)` at the end of processing

https://github.com/lhotse-speech/lhotse/blob/c42b181e728982f69ba1721959fe6f03078e821d/lhotse/qa.py#L40-L69

desh2608: Awesome!

pzelasko:
Thanks!

Issue #453, opened 2021-11-05
Storing Embeddings in Lhotse


danpovey:
This is kind of a major feature request, IDK whether you have time for this, Piotr, but it's part of a big direction that I want to go in, in Icefall, and it is a good way to leverage the strengths of Lhotse.

It's related to things like BERT, HUBERT, wav2vec2, and so on... basically the idea is that we will be predicting some kind of code generated from an embedding that was obtained from the activations of a previously trained network.  This could be used either as the primary loss or as an auxiliary loss.  In general, I want to have a flexible setup where you have a place on disk where you store a bunch of embeddings produced by previous models, and you train your model to predict a bunch of them simultaneously.  The advantage of doing this via codes dumped to disk, and not computing them as needed, is that we don't have to worry about loading a bunch of different models into memory and having all their associated code still working.  Also it's more efficient compute-wise as we only have to dump the data once.

Right now, for ease of implementation, I am assuming that each embedding, on each frame, is going to be converted into a short sequence of labels, with the labels spanning 0..255 (1 byte), where each label likely represents some subset of dimensions of the embedding.  In practice I anticipate that we might have anywhere between 1 and 16 labels per frame.  So the natural way these would be stored for a single cut is as a byte array, of shape (num_frames, seq_len), where e.g. seq_len=4.  [For cases where seq_len is very small, we might want to consider just storing this inline in the jsonl file, rather than using hdf5, but this is a side issue.]

There are two types of workflow that we'd have, when dealing with these labels.

 (i) Dumping the labels.  The scenario is we have trained some kind of model, and we want to extract embeddings from it.  The part of this that interfaces with Lhotse, is that we want to dump these embeddings to disk, and the relevant metadata.   The modeling code will give us the int8 array, e.g. of shape (B, T, S) i.e. (batch-size, num-frames, bytes-per-frame).  [I'd like to give these codes a special name, "encoding"].   I suppose we could dump the encodings for each supervision segment, but in general I'd like to think of this as corresponding to cuts rather than supervision segments.   We can if needed assume here that there is no Musan or SpecAug augmentation and no cut concatenation (but there may be speed augmentation).  It might be kind of slow to have a single process dump the encodings for large datasets (think Gigaspeech XL, which is 20K hours or something like that).  It would be nice if there were a way to split and re-combine data, that would work with this.

[...]

Issue #393, opened 2021-09-01
Supporting custom data (array/tensor) fields in cuts


pzelasko:
The main idea is that we want to be able to attach things like fbank, pitch, posteriors, alignments, etc. to the cuts in the same way we currently support `.features` and `.load_features`. This means a user could have a single cut object with multiple types of data. But we don't want to define all the possible attributes ahead of time.

The proposed syntax is inspired by pandas and k2 and would look like this:

```python
# attach feature manifest for fbank
cut.fbank = Features(...)
feats = cut.load_fbank()

# attach feature manifest for pitch
cut.pitch = Features(...)
pitch_feats = cut.load_pitch()

# attach a more generic array manifest
cut.posteriors = Array(...)
posts = cut.load_posteriors()

# and so on...
cut.alignment = Array(...)
ali = cut.load_alignment()

cut.ivector = Array(...)
ivec = cut.load_ivector()
```

This can be achieved with a little metaprogramming by overriding `setattr` like this (it's just pseudo-code):

```python
def __setattr__(self, key, value):
  self.key = value
  if isinstance(value, Array):
    method_key = f'load_{key}'
    self.method_key = ...  # assings a method that does the right thing
```

It'd also support serialization + deserialization to restore these attributes correctly. 

To support truncation, the `Array` manifest type would have a `shape` and `temporal_dim` attribute; if `temporal_dim` is None it means it's sth like an i-vector, otherwise we would use the `temporal_dim` to truncate that dimension.

For padding, it'd accept a user defined value. For mixing, `Features` support it already, and `Array` would raise an exception or a warning and strip out the attribute from the cut.

danpovey:
Hm, interesting.
So if there is a `temporal_dim` I assume that means it has a frame rate (is this worked out implicitly from the
number of features?)

pzelasko:
Yeah there needs to be a frame rate / frame shift

csukuangfj:
```python3
def __setattr__(self, key, value):
  self.key = value
  if isinstance(value, Array):
    method_key = f'load_{key}'
    self.method_key = ...  # assings a method that does the right thing
```

Could you explain how the following statement works?
> elf.method_key = ...  # assings a method that does the right thing

---

> It'd also support serialization + deserialization to restore these attributes correctly.

Also, how would it support serialization and deserialization?

pzelasko:
[...]

PR #459, opened 2021-11-09
Support custom MonoCut attrs (with special support for loading Arrays)

pzelasko:
This seems to work in trivial scenarios, but needs more testing. 

See the unit test for an example of usage: storing some data in HDF5, assigning its manfiest to `ivector`, and then calling `cut.load_ivector()` without Cut having that method defined anywhere.

This is fairly painless, but needs more testing especially with `TemporalArray` to see if padding and truncation are behaving correctly. `MixedCut` is not going to have this functionality (it's not well defined what it means to e.g. mix embeddings...)

For discussion/design/motivation see #453 and #393

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -779,7 +779,7 @@ class MonoCut(Cut):
         If ``item`` is found in ``custom``, we'll return ``custom[item]``.
         If ``item`` starts with "load_", we'll assume the name of the relevant
         attribute comes after that, and that value of that field is of type
-        Array or TemporalArray. We'll return it's ``load`` method to call by the user.
+        Array or TemporalArray. We'll return its ``load`` method to call by the user.
         """
         from lhotse.array import Array, TemporalArray
 
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -19,6 +19,7 @@ from typing import (
     Optional,
     Sequence,
     Set,
+    Tuple,
     Type,
     TypeVar,
     Union,
@@ -766,6 +767,11 @@ class MonoCut(Cut):
     custom: Optional[Dict[str, Any]] = None
 
     def __setattr__(self, key: str, value: Any):
+        """
+        This magic function is called when the user tries to set an attribute.
+        We use it as syntactic sugar to store custom attributes in ``self.custom``
+        field, so that they can be (de)serialized later.
+        """
         if key in self.__dataclass_fields__:
             super().__setattr__(key, value)
         else:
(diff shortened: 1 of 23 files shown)


pzelasko:
It also needs testing for serialization (but I think it should work out of the box)

lhotse/cut.py
```diff
@@ -760,6 +760,50 @@ class MonoCut(Cut):
...
+        If ``item`` is found in ``custom``, we'll return ``custom[item]``.
+        If ``item`` starts with "load_", we'll assume the name of the relevant
+        attribute comes after that, and that value of that field is of type
+        Array or TemporalArray. We'll return it's ``load`` method to call by the user.
+        """
+        from lhotse.array import Array, TemporalArray
+
+        custom = self.custom
+        if custom is None:
+            raise AttributeError()
```
csukuangfj, line 788: Shall we also display the name of the accessed attribute in the exception?
Just for better diagnostics.

test/cut/test_custom_attrs.py
```diff
@@ -0,0 +1,19 @@
...
+import numpy as np
+
+from lhotse import LilcomHdf5Writer, MonoCut
+from lhotse.array import Array
+
+
+def test_cut_load_array():
+    array = np.arange(20).astype(np.float32)
+    with NamedTemporaryFile(suffix=".h5") as f, LilcomHdf5Writer(f.name) as writer:
+        manifest = Array.store(key="utt1", value=array, writer=writer)
```
danpovey, line 12: I was a little confused here by the name "manifest" as I had assumed "manifest" always refers to a collection of metadata, but I suppose it refers to an item of metadata here?  I wonder if those should be given different names...

pzelasko: This is correct — calling Array.store returns a manifest for the stored data. Not sure how to balance informativeness and verbosity — Array.store_and_create_manifest is a mouthful. Do you have any suggestions?

pzelasko: I changed it to `manifest = writer.store_array(...)` in the other PR, maybe this is more helpful?

test/cut/test_custom_attrs.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        cut_trunc = cut.truncate(duration=5.0)
+
+        alignment_piece = cut_trunc.load_alignment()
+        assert alignment_piece.shape == (13,)  # 5.0 / 0.4 == 12.5 ~= 13
+        np.testing.assert_equal(alignment[:13], alignment_piece)
+
+
+@pytest.mark.parametrize("pad_value", [-1, 0])
+def test_cut_load_temporal_array_pad(pad_value):
```
pzelasko, line 86: This test shows how arrays read via padded cuts are auto-padded.

pzelasko:
The auto-padding code seems to work fine including edge-cases; I will start working on the next PR that shows how to use this in a PyTorch Dataset. I might do some refactoring too to make the Cut/MixedCut `__setattr__` code easier to comprehend/maintain.

test/cut/test_custom_attrs_randomized.py
```diff
@@ -0,0 +1,61 @@
...
+            ),
+            label=f"Padded cut duration",
+        )
+        padded = cut.pad(
+            duration=duration,
+            direction=pad_direction,
+            pad_value_dict={"codebook_indices": -1},
+        )
+        # Test the invariants
+        array = padded.load_codebook_indices()
```
pzelasko, line 55: This randomized test runs 500 times with cuts of different sampling_rates, durations, etc. It attaches a 1D TemporalArray called "codebook_indices" with random ints, pads the cut, and reads out the padded array from the padded cut.

lhotse/testing/fixtures.py
```diff
@@ -124,3 +130,13 @@ def _with_alignment(
...
+
+    def _with_custom_temporal_array(self, cut: MonoCut, frame_shift: Seconds) -> None:
+        d = TemporaryDirectory()
+        self.dirs.append(d)
+        num_frames = seconds_to_frames(cut.duration, frame_shift=frame_shift)
+        array = np.random.randint(256, size=(num_frames,))
+        with NumpyHdf5Writer(d.name) as storage:
+            cut.codebook_indices = storage.store_array(
+                key="ali1", value=array, frame_shift=frame_shift, temporal_dim=0
+            )
```
pzelasko, line 142: This snippet shows how "codebook_indices" are stored and attached to a Cut manifest.

danpovey:
Thanks!  @glynpu is working on using this framework to dump encodings of embeddings for Librispeech, and train on them as an auxiliary loss.  Then we'll go on to try to handle the multi-dataset case, with Gigaspeech.

pzelasko:
Great. Keep me updated on the progress and if you need anything else. I’ll write some docs and examples for this mechanism soon.

danpovey:
Thanks!

PR #460, opened 2021-11-10
Fix misleading manifest format in the comments

stachu86:
This is fix for outdated comments, about format of manifest files, in the examples scripts. Although it is very small issue, it made me to search for non-existing YAML files and left me a little confused.


pzelasko:
Thanks for this! I admit that the `examples` directory might be outdated, because at some point we started to focus more on Lhotse usage in `snowfall` and then `icefall`.

pzelasko:
I set up issue #464 to track this

PR #466, opened 2021-11-11
Ensure the users who installed PyTorch also install torchaudio themselves

pzelasko:


pzelasko:
Trying to make torchaudio version detection right, I reached the conclusion: if you really need a specific PyTorch version, just install torchaudio along with it. If you don't care, we'll fetch whatever is latest on PyPI for you.

There are cases where it's impossible to provide the right torchaudio version for the users: e.g., PyTorch LTS requires `-f` option with the right link as it's not on PyPI; since 1.10.0, torchaudio on PyPI seems to require CUDA; etc.

pzelasko:
CC: @csukuangfj @danpovey

setup.py
```diff
@@ -140,34 +140,18 @@ def mark_lhotse_version(version: str) -> None:
...
+    try:
+        import torchaudio
+    except ImportError:
         raise ValueError(
-            f"Lhotse requires torch>=1.7.1 and torchaudio>=0.7.2 -- "
-            f"please update your torch (detected version: {torch_ver})."
+            "We detected that you have already installed PyTorch, but haven't installed torchaudio. "
+            "Unfortunately we can't detect the compatible torchaudio version for you; "
+            "you will have to install it manually. "
+            "For instructions, please refer to https://pytorch.org/get-started/locally/"
```
csukuangfj, line 154: I would recommend putting the link about how to install torchaudio here, i.e.,
https://github.com/pytorch/audio#dependencies

PR #476, opened 2021-11-17
Collation of custom cut fields into PyTorch tensors

pzelasko:
CC @danpovey 

This should be a minimal version that works for your use-cases, I think I'm only missing an optional tag for array manifests that can be used e.g. to indicate what model created it.


lhotse/dataset/collation.py
```diff
@@ -168,6 +168,66 @@ def collate_audio(
...
+
+        - ``(batch, )`` for anything else, such as int or float: we will simply stack them into
+            a list and tensorize it.
+
+    :param cuts: a :class:`CutSet` used to load the features.
+    :param field: name of the custom field to be retrieved.
+    :param pad_value: value to be used for padding the temporal arrays.
+        Ignored for non-temporal array and non-array attributes.
+    :param pad_direction: where to apply the padding (``right``, ``left``, or ``both``).
+    :return: a tuple of tensors ``(features, features_lens)``.
```
csukuangfj, line 198: It can also return a single tensor.

test/dataset/test_collation.py
```diff
@@ -58,3 +70,145 @@ def test_collate_feature_padding():
...
+        for cut in cuts:
+            expected_xvectors.append(np.random.randn(EMBEDDING_SIZE).astype(np.float32))
+            cut.xvector = writer.store_array(cut.id, expected_xvectors[-1])
+
+        xvectors = collate_custom_field(cuts, "xvector")
+        assert isinstance(xvectors, torch.Tensor)
+        assert xvectors.dtype == torch.float32
+        assert xvectors.shape == (len(cuts), EMBEDDING_SIZE)
+        for idx, xvec in enumerate(expected_xvectors):
+            torch.testing.assert_allclose(xvectors[idx], xvec)
```
danpovey, line 90: So is the idea that we would then write out the CutSet again with the new metadata included, after writing out the data?

csukuangfj: Yes, I believe so.

I have used it in
https://github.com/k2-fsa/icefall/blob/62ada37d4ec0205512264f8c173b33691309ec29/egs/librispeech/ASR/conformer_ctc/ali.py#L230

at https://github.com/k2-fsa/icefall/pull/125

lhotse/dataset/collation.py
```diff
@@ -168,6 +168,66 @@ def collate_audio(
...
+            "Cannot collate manifests of type Array with different shapes, "
+            "because we don't know which dimension must be padded. "
+            "Use TemporalArray manifests and try again."
+        )
+        return torch.stack([torch.from_numpy(c.load_custom(field)) for c in cuts])
+    elif isinstance(first_manifest, TemporalArray):
+        # Expected data type: variable-sized tensors (along only one dimension).
+        # Pad across that dimension, then stack at dimension 0.
+        assert (
+            pad_value is not None
```
csukuangfj, line 216: How about reading the padding value from cut's attribute?

For instance, if there is a custom field called `xxx` and if
there exists another field called `xxx_pad_value`, can we use `xxx_pad_value`
as the padding value for `xxx`?

 If there is no `xxx_pad_value`, can we use a default value, say 0, for the padding value?

We are doing similar things in k2, see
https://github.com/k2-fsa/k2/blob/bdcaaf828212913366110a0c45aab4ea40b1cbf9/k2/python/k2/fsa.py#L308
```
        ans = getattr(self, attribute_name + '_filler', 0)
```

test/dataset/test_collation.py
pzelasko: That's right.

Issue #169, opened 2020-12-19
Trasformation - reverberation + noise


cyrta:
Hi, 
I would like to help with the project and post pr with some code to simulate room, as a data augmentation. 
RIR generated on fly created with some predefined distribution and source-sink position.
I've seen [lazy loading audio transformation code](https://github.com/lhotse-speech/lhotse/pull/167). 
However, I would like to ask you Piotr and Dan how you see this type of augmentation inside lhotse.

danpovey:
Sounds interesting but well have.to.decide.whether it makes more sense to
dump them to disk from a.separate.repo,to avoid excess dependencies.
My feeling is the fine detail in the RIR may not matter so much.  could
maybe start with code to apply RIRs?

pzelasko:
Sounds good. Keep in mind that we'll need to be able to estimate how many samples are added due to reverberation to update the num_samples and duration in Recording/Cut manifests (without actually running it). I don't think we need to update the supervisions start/duration.

We could start with [torchaudio's libsox bindings for reverb](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/augmentation/torchaudio.py#L158) unless you had a more refined method in mind?

pzelasko:
BTW there are 3 ways to perform data augmentation in Lhotse right now (WavAugment,  torchaudio's SoxEffectTransform via `augment_fn` API, and the newly introduced lazy mechanism). I will deprecate the first two in favor of the lazy mechanism, so going forward we'll probably be adding new augmentations as methods on `Recording`/`Cut`/`CutSet`. It's more work, but it seems the API is much more user friendly and bug-free (e.g. [see the diff here](https://github.com/k2-fsa/snowfall/pull/54/files)).

desh2608:
@cyrta is there any update on this?

@pzelasko I see that there is a recipe available for MUSAN. Would it make sense to add a recipe for preparation of the [RIR corpus](https://www.openslr.org/28/) and then add a Cut method to convolve it with a selected RIR as lazy augmentation? For isotropic noises I suppose the `cut.mix()` method should work?

pzelasko:
Thanks @desh2608, that makes sense to me. Note that we might want to eventually support multiple types of reverb methods: RIR-based, geometric-based (like pyroomacoustic), effect-based (like SOX), etc. so it makes sense to name it explicitly like `reverb_rir()`.

cyrta:
[...]

PR #477, opened 2021-11-19
Reverberation using room impulse response

desh2608:
This PR adds capabilties for augmentation by adding RIR based reverberation. The core implementation is based on Kaldi's [wav-reverberate](https://github.com/kaldi-asr/kaldi/blob/master/src/featbin/wav-reverberate.cc) utility, and the FFT based block convolution is borrowed from the implementation in [torch-audiomentations](https://github.com/asteroid-team/torch-audiomentations/blob/master/torch_audiomentations/utils/convolution.py).

This partly addresses https://github.com/lhotse-speech/lhotse/issues/169.

Here are examples of how to use:

```python
# Apply reverb to single cut
rir = Recording.from_file('/path/to/rir.wav')
cut_rvb = cut.reverb_rir(rir)

# Apply reverb to cut set
rirs = RecordingSet.from_dir('/path/to/rirs/')
cuts_rvb = cuts.reverb_rir(rirs)
```


lhotse/augmentation/torchaudio.py
```diff
@@ -366,6 +368,92 @@ def reverse_timestamps(
...
+        """
+        :param samples: The audio samples to reverberate (must be single-channel).
+        :param sampling_rate: The sampling rate of the audio samples.
+        """
+        assert samples.shape[0] == 1, "The input audio must be single-channel."
+        sampling_rate = int(sampling_rate)  # paranoia mode
+
+        from lhotse import Recording
+
+        rir_ = Recording.from_dict(copy.deepcopy(self.rir)).load_audio()
```
desh2608, line 402: @pzelasko Here I am passing a copy of the rir dict so that the original doesn't get modified in the `from_dict()` method. Is this reasonable or is there a better way to do this?

lhotse/cut.py
```diff
@@ -2135,6 +2208,51 @@ def perturb_volume(self, factor: float, affix_id: bool = True) -> "MixedCut":
             ],
         )
 
+    def reverb_rir(
+        self,
+        rir_cut: "MonoCut",
```
pzelasko, line 2213: Is there a reason to pass RIR as a cut instead of as a recording? Do you expect to truncate / pad them somehow?

lhotse/cut.py
```diff
@@ -1260,6 +1262,56 @@ def perturb_volume(self, factor: float, affix_id: bool = True) -> "MonoCut":
             supervisions=supervisions_vp,
         )
 
+    def reverb_rir(
+        self,
+        rir_cut: "MonoCut",
+        shift_output: bool = True,
```
pzelasko, line 1268: Note: if you set `shift_output` to False, it might affect the cuts duration (and maybe the offset). Not sure if this is worth supporting, maybe it's better to remove this option from cuts and let the users reverberate recordings first if they don't want to shift output? If you decide to go this way, please add a note in the docs.

lhotse/cut.py
```diff
@@ -2135,6 +2208,51 @@ def perturb_volume(self, factor: float, affix_id: bool = True) -> "MixedCut":
...
+        assert (
+            self.has_recording
+        ), "Cannot apply reverberation on a MixedCut without Recording."
+        if self.has_features:
+            logging.warning(
+                "Attempting to reverberate a MixedCut that references pre-computed features. "
+                "The feature manifest will be detached, as we do not support feature-domain "
+                "reverberation."
+            )
+            self.features = None
```
pzelasko, line 2238: I don't think this code will work, IIRC `features` is not defined for MixedCut (but `has_features` and `load_features` are). Instead, rely on the implementation inside MonoCut.

pzelasko: I see the whole implementation is copied from MonoCut -- this is not correct. MixedCut should just apply the transform to every underlying track and return a new MixedCut.

However, if you're transforming the time information somehow, you might need to adjust the offsets in MixTracks. Probably it's not worth transforming time information to avoid the complexity.

lhotse/cut.py
```diff
@@ -3306,6 +3426,35 @@ def perturb_volume(self, factor: float, affix_id: bool = True) -> "CutSet":
             lambda cut: cut.perturb_volume(factor=factor, affix_id=affix_id)
         )
 
+    def reverb_rir(
+        self,
+        rir_cuts: "CutSet",
```
pzelasko, line 3431: Same as above: why not `RecordingSet`?

lhotse/dataset/cut_transforms/reverberate.py
```diff
@@ -0,0 +1,42 @@
...
+class ReverbWithImpulseResponse:
+    """
+    A transform on batch of cuts (``CutSet``) that convolves each cut with an impulse
+    response with some probability :attr:`p`. The impulse response is chosen randomly from
+    a specified CutSet of RIRs :attr:`rir_cuts`.
+    """
+
+    def __init__(
+        self,
+        rir_cuts: CutSet,
```
pzelasko, line 15: Same q as above

lhotse/cut.py
desh2608: Yeah, good point. I passed it as a Cut so it may be possible to support transformations on the RIR in the future. But you're right, perhaps it's best to use it as a Recording for now.

lhotse/cut.py
desh2608: Okay I'll remove the option and only have it so the duration remains consistent.

lhotse/cut.py
desh2608: I think reverb should be applied on the mixed audio instead of the underlying tracks?

lhotse/augmentation/torchaudio.py
pzelasko: I see -- but I think that shallow copy should be sufficient here (from_dict is only popping keys, the values are untouched).

lhotse/cut.py
pzelasko: Convolutions are linear so it shouldn't matter too much, unless I'm missing something here?

lhotse/augmentation/torchaudio.py
desh2608: That makes sense, thanks!

lhotse/cut.py
desh2608: It won't matter when you are mixing 2 speech sources, but when you are mixing a speech source and a noise source, you wouldn't want to reverberate the noise source.

pzelasko: But in that case you wouldn't call `cut.mix(noise_cut).reverb_rir()` (which is the case being implemented in MixedCut's `reverb_rir`); you'd be calling `cut.reverb_rir().mix(noise_cut)`, and that's the case for MonoCut's `reverb_rir`.

desh2608: Hmm, yeah that's true. I'll make the change accordingly.

pzelasko: Also if you want reverb to be applied on the mixed audio, it might be best to run `compute_and_store_recording()` on a `MixedCut` first; or just use `ReverbWithImpulseResponse` transform directly on `mixed_cut.load_audio()`

test/augmentation/test_torchaudio.py
```diff
@@ -64,6 +76,24 @@ def test_volume_does_not_change_num_samples(audio):
...
+
+
+@pytest.mark.parametrize("normalize_output", [True, False])
+def test_reverb_normalize_output(audio, rir, normalize_output):
+    augment_fn = ReverbWithImpulseResponse(rir=rir, normalize_output=normalize_output)
+    orig_energy = np.sum(np.abs(audio) ** 2) / audio.shape[1]
+    for _ in range(10):
+        augmented_audio = augment_fn(audio, sampling_rate=SAMPLING_RATE)
+        rvb_energy = np.sum(np.abs(augmented_audio) ** 2) / augmented_audio.shape[1]
+        if normalize_output:
```
pzelasko, line 93: should we assert something when `normalize_output` is `False`? Do you expect the reverb-ed signal energy to be higher in that case?

test/augmentation/test_torchaudio.py
```diff
@@ -113,6 +143,17 @@ def test_serialize_deserialize_transform_volume(audio):
     assert_array_almost_equal(perturbed_volume, audio * 0.5)
 
 
+@pytest.mark.parametrize("recording_to_dict", [True, False])
+def test_serialize_deserialize_transform_reverb(audio, rir, recording_to_dict):
+    rir = rir.to_dict() if recording_to_dict else rir
+    reverb_orig = ReverbWithImpulseResponse(rir=rir)
+    data_reverb = reverb_orig.to_dict()
+    reverb = ReverbWithImpulseResponse.from_dict(data_reverb)
+    rvb_audio = reverb(audio, SAMPLING_RATE)
```
pzelasko, line 152: I think you don't need to apply the effect here, just compare equality of transform objects before and after serialization (since they're dataclasses I think __eq__ is auto-defined).

pzelasko (commented):
Thanks, it looks great! Two minor comments and LGTM

pzelasko:
BTW just as a sanity check -- did you listen to some audio transformed with this reverb transform?

desh2608:
> BTW just as a sanity check -- did you listen to some audio transformed with this reverb transform?

Yeah, I reverberated a LibriSpeech utterance and the output sounds reasonable (it is also in the new fixture).

test/augmentation/test_torchaudio.py
desh2608: Hmm, TBH I am not sure what to expect. I borrowed this flag from the Kaldi implementation, where it is set to True by default. I haven't seen any uses where it is called with this flag set to false, but I would think the energy would be higher in that case.

pzelasko:
I'll just merge, thanks! Should be very useful.

PR #479, opened 2021-11-19
A change for tedlium.py

luomingshuang:
I just change `{split}_supervisions.json` to `supervisions_{split}.json`,  `{split}_recordings.json` to `recordings_{split}.json`. After changing, the function `read_manifests_if_cached` in lhotse.recipes.utils can load the json files in `manifests.`
https://github.com/lhotse-speech/lhotse/blob/7c1e3bd33f3b1ef23b67ee3c2e70443bf302e655/lhotse/recipes/utils.py#L39-L41


pzelasko:
Thanks!

PR #481, opened 2021-11-19
Force single-threaded execution for ffmpeg

pzelasko:


glynpu:
Option -thread 1 will also create lots of threads. 
I didn't run the python code. But with a simple shell command:
```
ffmpeg -threads 1  -i /ceph-hw/ly/speech_data/giga/release/GigaSpeech/audio/youtube/P0144/YOU0000014377.opus -ar 16000 -f wav - > haha.wav
```

Lots of threads(I didn't count, but I think the number is os.cpu_cunts) are created as following figure:
image

jtrmal:
maybe it doesn't matter as long as they don't accrue more cpu time?
y.

pzelasko:
I think it matters because once you have a lot of parallelism in the dataloader, you are spawning an absurd number of threads and it may slow down your OS. I think it happened to me on a few occasions but I didn’t know where it’s coming from back then. We should try to find a solution for this but I’m not sure what are the next steps.

jtrmal:
Maybe this would be related
https://stackoverflow.com/questions/41006716/ffmpeg-is-using-more-threads-than-i-expect-when-using-threads
Seems like it's a positional argument (input and output and who knows what
else).
No idea, really.
But spawning threads/processes should not cause problems, AFAIK, unless you
do thousands and more  -- but I'm always opened to be surprised :shrug:

jtrmal:
but I definitely agree it's better to have it under the control... I don't
mean to sound argumentative :)
y

pzelasko:
Oh I see so you have to provide the -threads option twice, once for input and once for output. @glynpu can you check it out?

glynpu:
> provide the -threads option twice

Yes, It  works.
The command I am using is:
```
ffmpeg -threads 1  -i audio/youtube/P0144/YOU0000014377.opus -threads 1 -ar 16000 -f wav - > foo.wav
```

Issue #478, opened 2021-11-19
size mismatch when using `collate_custom_field`.


csukuangfj:
In `icefall`, when we say `subsampling_factor` is 4, we mean
```
num_output_frames = ((num_input_frames - 1 ) // 2) - 1) // 2
```

But in `lhotse`, I think it assumes
```
num_output_frames = round(num_input_frames / 4)
```

I just encountered the following problem. Suppose a `CusSet` has 2 cuts:
- cut 1:
  - duration: 4.9 seconds, num_frames: 490, after subsampling by a factor of 4 in `icefall`, it has 121 frames
  - A custom field is attached to it, named `label_alignment`, which has 121 frames
  
- cut 2:
  - It has two cuts, a MonoCut and a PaddingCut
  - The MonoCut has duration: 4.895 seconds, num_frames 490, after subsampling by a factor of 4 in `icefall`, it has 121 frames
  - A custom field is attached to it, named `label_alignment`, which has 121 frames
  - There is a PaddingCut with a duration 0.005 seconds

While loading `cut 1`, it returns an array of shape `(121,)`, see the code below
https://github.com/lhotse-speech/lhotse/blob/7c1e3bd33f3b1ef23b67ee3c2e70443bf302e655/lhotse/cut.py#L835-L837

While loading `cut 2`, it will pad 2 extra frames, which returns an array of shape `(123,)`, see the code below
https://github.com/lhotse-speech/lhotse/blob/7c1e3bd33f3b1ef23b67ee3c2e70443bf302e655/lhotse/cut.py#L1849-L1861
https://github.com/lhotse-speech/lhotse/blob/7c1e3bd33f3b1ef23b67ee3c2e70443bf302e655/lhotse/array.py#L285-L287

The consequence is that it throws in the following code while trying to stack the above arrays.
https://github.com/lhotse-speech/lhotse/blob/7c1e3bd33f3b1ef23b67ee3c2e70443bf302e655/lhotse/dataset/collation.py#L227-L228

```
    tensors = torch.stack([torch.from_numpy(c.load_custom(field)) for c in cuts])
RuntimeError: stack expects each tensor to be equal size, but got [121] at entry 0 and [123] at entry 1
```

csukuangfj:
Can we just return a list of arrays and let the user do the padding?

pzelasko:
[...]

PR #482, opened 2021-11-20
`custom_collate_field` pads to the longest read array instead of using `CutSet.pad()`

pzelasko:
This test is for the scenario described in #478 @csukuangfj 

I gave it some thought and I reached the conclusion that... the alignments for the two cuts you describe actually can be represented as having a different frame shift. This is because their number of frames is equal, but durations are not; so each frame in the longer cut represents a tiny bit more amount of duration than in the shorter cut. If you set a per-cut `frame_shift=cut.duration / ali.shape[0]`, then it seems you don't run into the issue anymore, even though the micro-padding-cut is still created. See the test I added for details.

I can test this more to make sure that there are no broken cases in this approach; I think this might be the cleanest solution because it preserves Lhotse's "magic" ability to pad every time-related value attached to a cut (duration, audio, features, and now generic temporal arrays) with one function call. But it might make sense to modify the API for `writer.store_array` -- allow specifying `duration` instead of `frame_shift`, and infer `frame_shift` inside the function from `duration`, `array.shape` and `temporal_dim`.

WDYT?


csukuangfj:
👍

pzelasko:
@csukuangfj that idea didn't work out; it'd have failed if we padded to some large duration e.g. 300s, then the shapes are different again because of different frame shifts. 

I figured it'll be best to modify `collate_custom_field` to just read out the arrays and collate them to match the longest found in a cut set, along `temporal_dim`. This way it's independent of how you define `frame_shift`. If you want to use `cuts.pad()`, then the old problem persists though, and I'm not sure what is the right solution yet.

pzelasko:
@csukuangfj what do you think about the following:

```python
cut.label_alignment = ...
diff = cut.duration - cut.label_alignment.duration
if diff > 0:
    cut = cut.truncate(duration=cut.label_alignment.duration, offset=diff / 2)
    # this is centering the cut, if left/right shift is more adequate make offset=0 or offset=diff
```

Having this in the user code ensures you don't run into issues related to `frame_shift`, and seems the correct way to go for me.  I could add some kind of utility that can be called by the user like `cut.trim(to=cut.label_alignment, align='left|right|center')` to make it easier to use arrays which don't align perfectly with cuts.

pzelasko:
BTW I added some code that can be run to ensure you don't run into duration-related surprises later, see the test:

https://github.com/lhotse-speech/lhotse/pull/482/files#diff-829adc154c465e8174b0d2311a1537958fd35259f520ea863c65a367d9ab0c51R145-R166

csukuangfj:
Instead of comparing `duration`
```
diff = cut.duration - cut.label_alignment.duration
```

how about comparing the number of frames between `cut.features.num_frames / subsampling_factor` and `cut.label_alignment.num_frames`.
If the diff is within `subsampling_factor - 1`, then everything is assumed to be OK; otherwise, we just throw.
During neural network computation, we just truncate the number of output frames to match `cut.label_alignment`.

I think this is the approach we are currently using in k2.
Please see
- https://github.com/k2-fsa/snowfall/pull/181#discussion_r623684445
- https://github.com/k2-fsa/k2/pull/732

lhotse/array.py
```diff
@@ -121,9 +121,13 @@ def shape(self) -> List[int]:
...
         return len(self.shape)
 
+    @property
+    def num_frames(self) -> int:
+        return self.shape[self.temporal_dim]
+
     @property
     def duration(self) -> Seconds:
-        return self.shape[self.temporal_dim] * self.frame_shift
+        return self.num_frames * self.frame_shift
```
csukuangfj, line 130: Is `frame_shift` supposed to be different from the one contained in `cut.frame_shift`?

pzelasko: Originally, `cut` was designed to have only features and recording, so there was only one `frame_shift`. But now, because the custom fields are generic, you can put as many arrays with different frame shifts as you want. 

Lhotse doesn't expect any relationship between `cut.frame_shift` and a custom field's `frame_shift`. It is completely up to the user.

pzelasko: Just to give a specific example:

`cut.features` -> `frame_shift=0.01`
`cut.phone_alignment` -> `frame_shift=0.04`
`cut.online_ivectors` -> `frame_shift=0.1`

pzelasko:
> Instead of comparing `duration`
> 
> ```
> diff = cut.duration - cut.label_alignment.duration
> ```
> 
> how about comparing the number of frames between `cut.features.num_frames / subsampling_factor` and `cut.label_alignment.num_frames`. If the diff is within `subsampling_factor - 1`, then everything is assumed to be OK; otherwise, we just throw. During neural network computation, we just truncate the number of output frames to match `cut.label_alignment`.
> 
> I think this is the approach we are currently using in k2. Please see
> 
> * [Vgg frontend k2-fsa/snowfall#181 (comment)](https://github.com/k2-fsa/snowfall/pull/181#discussion_r623684445)
> * [Add `allow_truncate` to DenseFsaVec.__init__ k2-fsa/k2#732](https://github.com/k2-fsa/k2/pull/732)

That makes sense to me, but you will also have to adjust the duration of the cut, so that Lhotse "knows" these things are consistent. Anyway, I think this is something you have to do in the user code, as it is specific to your use of `custom` fields. The best I can do from Lhotse side is to give you the tools to validate the consistency between Cut and TemporalArray, and maybe some tools to help you auto-fix cases like these.

pzelasko:
I will merge this one since it fixes a number of issues, but let's continue the discussion and I'll make other PRs as needed.

csukuangfj:
Thanks! The current fix works fine.

Issue #485, opened 2021-11-22
Save audio function


desh2608:
I wonder if it would be useful to have a `save_audio()` method for the cuts and recordings? Currently I just use torchaudio, but perhaps we can just have a wrapper around the torchaudio function.

jtrmal:
sounds useful to me
y.

pzelasko:
It’s called “compute_and_store_recording” but maybe it’s not the best name. It’s because it triggers all the computation from lazy transforms.. do you think we should rename it?

jtrmal:
how about save_computed_audio or save_transformed_audio?
I do not have strong opinion, just slacking around, but "store" wouldn't
exactly make me believe that is the function I want to save something to
disk :)
y.

pzelasko:
It's not necessarily computed or transformed, could be the same as the original one. `save_audio` makes sense to me, but I think the "old" name should not be removed but just deprecated with a warning, as some people are using it. If any of you would like to go for it, please make a PR.

desh2608:
Closing since this is now added in https://github.com/lhotse-speech/lhotse/pull/486.

PR #486, opened 2021-11-22
Deprecate compute_and_store_recording, add save_audio

desh2608:
This addresses https://github.com/lhotse-speech/lhotse/issues/485

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -490,7 +490,9 @@ class Cut:
                     )
         return indexed
 
-    @deprecated("Please use save_audio() instead.")
+    @deprecated(
+        "Cut.compute_and_store_recording will be removed in a future release. Please use save_audio() instead."
+    )
     def compute_and_store_recording(
         self,
         storage_path: Pathlike,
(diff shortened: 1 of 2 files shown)


lhotse/cut.py
```diff
@@ -489,6 +490,7 @@ def index_supervisions(
                     )
         return indexed
 
+    @deprecated("Please use save_audio() instead.")
```
pzelasko, line 493: how about sth like: 

```suggestion
    @deprecated("Cut.compute_and_store_recording will be removed in a future release. Please use save_audio() instead.")
```

pzelasko: also to avoid unnecessary code duplication this function could just call `save_audio` under the hood.

lhotse/cut.py
```diff
@@ -3867,6 +3921,7 @@ def compute_and_store_features_batch(
 
         return CutSet.from_cuts(cuts_with_feats)
 
+    @deprecated("Please use save_audios() instead.")
```
pzelasko, line 3924: Same comments as for the single cut case.

desh2608:
@pzelasko Done :)

pzelasko (approved):

pzelasko:
Thanks!

PR #487, opened 2021-11-23
Add WenetSpeech recipe

pkufool:


lhotse/recipes/wenet_speech.py
```diff
@@ -0,0 +1,141 @@
...
+            recordings=RecordingSet.from_recordings(manifests[sub]["recordings"]),
+            supervisions=SupervisionSet.from_segments(manifests[sub]["supervisions"]),
+        )
+        validate_recordings_and_supervisions(
+            recordings=recordings, supervisions=supervisions
+        )
+
+        if output_dir is not None:
+            supervisions.to_json(output_dir / f"supervisions_{sub}.json")
+            recordings.to_json(output_dir / f"recordings_{sub}.json")
```
pzelasko, line 97: I suggest writing these out to JSONL format (one JSON per line) as it enables to read the Lhotse manifest lazily. I expect these are also possibly quite large, so you could also gzip them. It'd look like:

```suggestion
            recordings.to_file(output_dir / f"recordings_{sub}.jsonl.gz")
```

(the same for supervision manifest)

lhotse/recipes/wenet_speech.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    segments = defaultdict(dict)
+    for sub in subsets:
+        segments[sub] = []
+    for seg in audio["segments"]:
+        segment = SupervisionSegment(
+            id=seg["sid"],
+            recording_id=audio["aid"],
+            start=seg["begin_time"],
+            duration=seg["end_time"] - seg["begin_time"],
```
pzelasko, line 134: I suggest to use `from lhotse.utils import add_durations` and then: 

```suggestion
            duration=add_durations(seg["end_time"], -seg["begin_time"], sampling_rate),
```

It will ensure that duration doesn't suffer from floating point precision errors.

pzelasko (commented):
Thanks, looks great! I left a couple of comments, let me know when it's good to merge.

pkufool:
Thanks for your comments, will let you know when I think it is ready.
BTW, what should I do to fix codecov CI, sorry, I can not open the details page.

pkufool:
I think this is ready for merging.

pkufool:
Wait, it raises an exception when extracting features, there are some errors in manifests.
```
  File "/ceph-hw/kangwei/dev_tools/anaconda3/envs/k4/lib/python3.8/site-packages/lhotse-1.0.0.dev0+git.5bf44d4.dirty-py3.8.egg/lhotse/utils.py", line 587, in
wrapper
    return fn(*args, **kwargs)
  File "/ceph-hw/kangwei/dev_tools/anaconda3/envs/k4/lib/python3.8/site-packages/lhotse-1.0.0.dev0+git.5bf44d4.dirty-py3.8.egg/lhotse/audio.py", line 354, in
load_audio
    samples = source.load_audio(
  File "/ceph-hw/kangwei/dev_tools/anaconda3/envs/k4/lib/python3.8/site-packages/lhotse-1.0.0.dev0+git.5bf44d4.dirty-py3.8.egg/lhotse/audio.py", line 152, in
load_audio
    raise ValueError(
ValueError: Requested more audio (1716.88s) than available (1706.219s)

During handling of the above exception, another exception occurred:

  File "/ceph-hw/kangwei/dev_tools/anaconda3/envs/k4/lib/python3.8/site-packages/lhotse-1.0.0.dev0+git.5bf44d4.dirty-py3.8.egg/lhotse/utils.py", line 587, in
wrapper
    return fn(*args, **kwargs)
  File "/ceph-hw/kangwei/dev_tools/anaconda3/envs/k4/lib/python3.8/site-packages/lhotse-1.0.0.dev0+git.5bf44d4.dirty-py3.8.egg/lhotse/cut.py", line 952, in lo
ad_audio
    return self.recording.load_audio(
  File "/ceph-hw/kangwei/dev_tools/anaconda3/envs/k4/lib/python3.8/site-packages/lhotse-1.0.0.dev0+git.5bf44d4.dirty-py3.8.egg/lhotse/utils.py", line 589, in
wrapper
    raise type(e)(
ValueError: Requested more audio (1716.88s) than available (1706.219s)
[extra info] When calling: Recording.load_audio(args=(Recording(id='Y0000004754_BfYkD2JDxo0', sources=[AudioSource(type='file', channels=[0], source='/ceph-hw
/kangwei/code/icefall_wenet/egs/wenetspeech/ASR/download/wenetspeech/audio/train/youtube/B00018/Y0000004754_BfYkD2JDxo0.opus')], sampling_rate=16000, num_samp
les=27470080, duration=1716.88, transforms=None),) kwargs={'channels': 0, 'offset': 0, 'duration': 1716.88})
```

pzelasko:
It might be an issue if the original JSON manifest from WenetSpeech has incorrect duration values... In that case you might want to contact the authors to fix it (we did that with GigaSpeech) or create recording manifests using `Recording.from_file(path_to_opus, recording_id=aid, force_opus_sampling_rate=16000)`.

Another option is if you have an old ffmpeg, in which case its OPUS decoder returns different number of samples (but I think that's unlikely as your difference is 10+ seconds).

danpovey:
Is there some option we can use to catch these kinds of errors?  Or maybe
even just make it continue by default?

pkufool:
> It might be an issue if the original JSON manifest from WenetSpeech has incorrect duration values... 

I checked the real duration of that audio, it is 1716.88s (the same as the value in the original JSON manifest). 
This mismatching issue did not occur a second time. Now I have another issue, the feature extraction process always be killed by the system, Dan thought that would be the reason of using too much virtual memory.  I saw you have a lot of discussion about gigaspeech extraction, I will try to find answer there.
Let's see this mismatching issue when I can extract feature normally.

danpovey:
Piotr, for solving OOM issues in feature extraction, what do you think is the best approach?

danpovey:
I think what is happening is that the virtualization system we are under is a little strict about total memory usage.
Liyong was running with 8 processes; each was taking about 7 or 8 G of physical memory and a similar amount of virtual.  The system limit was 54G, so this was getting close to the limit or even going over anyway.
I did not notice that the processes were using gradually more memory.
I did notice also that the python processes were launching ffmpeg via the clone() system call.  That may have used even more virtual memory if they all called clone() in the same instant (i.e. before doing exec()).
Is there any silver bullet that we can use that will dramatically decrease memory usage while still enabling enough ffmpeg processes?

pzelasko:
> Is there some option we can use to catch these kinds of errors? Or maybe even just make it continue by default?
> […](#)

I added the option for the user to configure that now in #491

pzelasko:
> Piotr, for solving OOM issues in feature extraction, what do you think is the best approach?

I was offline for the last few days, and just noticed this now. I assume this is now solved by @csukuangfj, and the issue turned out to be the chunkwise extraction bug in kaldifeat that is now solved? If there's anything else just let me know.

csukuangfj:
> the issue turned out to be the chunkwise extraction bug in kaldifeat that is now solved?

That bug only caused CUDA OOM for utterances that are several hours long, not CPU OOM.  I think the key is to split the manifests into smaller pieces.

pzelasko:
Got it. Yes, splitting the manifest into pieces is one option that already works today. Another option is to open the input manifest lazily with `CutSet.from_jsonl_lazy()`, then you don’t have to hold all the input cut manifests 
in memory.

I’m going to add another option that stores the manifest iteratively so the memory usage doesn’t grow with time. Combined with a lazy input manifest, it will enable extracting the features on GPU with greatly reduced memory usage.

csukuangfj:
Another benefit of splitting the manifests into pieces is that each piece has its features stored in a separate file.
If you don't do it this way, the resulting feature file can be as large as 900 GB or more than 1 TB, e.g., for datasets like GigaSpeech.

pzelasko:
@pkufool is this good to merge?

pkufool:
> @pkufool is this good to merge?

Yes, I think the manifests are OK.

pzelasko:
Thanks again!

PR #493, opened 2021-11-29
Option to create a CutSet lazily when everything is sorted on recording IDs

pzelasko:
This is the first of a bunch of PRs I'm going to add to make Lhotse less memory-hungry when working on ~10k hours sized corpora like GigaSpeech. Most of the changes in these PRs will boil down to using these two methods (`CutSet` can be replaced with any other set class):
- for reads `CutSet.from_jsonl_lazy()`
- for writes `with CutSet.open_writer() as writer:`

I want the PRs to be separate for better changes visibility.

This one shows how to create a large CutSet practically not touching the memory. The users have to open recordings, supervisions, etc. lazily and the cuts are saved and freed one-by-one.

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import random
 import warnings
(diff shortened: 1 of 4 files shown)


lhotse/cut.py
```diff
@@ -4525,3 +4490,235 @@ def compute_supervisions_frame_mask(
             )
             mask[st:et] = 1.0
     return mask
+
+
+def create_cut_set_eager(
```
pzelasko, line 4495: Note: this code is practically unchanged, just moved to a separate function.

PR #498, opened 2021-12-01
Let kaldifeat handle the device of input waves.

csukuangfj:
In GigaSpeech, some utterances are more than 20 hours long, which cause CUDA OOM errors if they are moved to GPU for feature extraction.

I have updated kaldifeat (v1.12) to move frames to GPU in a chunk-wise manner. Also, to reduce the number of transfers between CPU and GPU, the default chunk size is changed from 10 seconds to 20 minutes.

See also https://github.com/k2-fsa/icefall/pull/120#issuecomment-983466377


pzelasko:
Thanks!

PR #503, opened 2021-12-03
``merge_supervisions`` method for cuts and cut sets

pzelasko:


lhotse/cut.py
```diff
@@ -1365,6 +1369,82 @@ def filter_supervisions(
...
+        msup = SupervisionSegment(
+            id=merge(s.id for s in sups),
+            # For MonoCut, the recording_id is always the same.
+            recording_id=sups[0].recording_id,
+            start=mstart,
+            duration=mduration,
+            # For MonoCut, the channel is always the same.
+            channel=sups[0].channel,
+            text=" ".join(s.text for s in sups if s.text) or None,
+            speaker=merge(s.speaker for s in sups if s.speaker) or None,
```
danpovey, line 1420: I don't understands how this becomes None if none of the speaker attributes are defined; it looks to me like it becomes "cat", because length=0 is not treated specially in merge()

lhotse/cut.py
```diff
@@ -1365,6 +1369,82 @@ def filter_supervisions(
...
+            language=merge(s.language for s in sups if s.language) or None,
+            gender=merge(s.gender for s in sups if s.gender) or None,
+            custom={
+                # Merge the string representations of custom fields.
```
danpovey, line 1424: What kinds of things do the custom fields of Supervision contain?
(I assume the encodings we discussed are in the custom fields of the Cut?)

lhotse/cut.py
```diff
@@ -1365,6 +1369,82 @@ def filter_supervisions(
...
+                    for s in sups
+                    if s.custom is not None and k in s.custom
+                )
+                for k in custom_keys
+            },
+            alignment={
+                # Concatenate the lists of alignment units.
+                k: reduce(
+                    add,
+                    (
```
danpovey, line 1436: Hm, it seems to me like this might be a bit of a simplification?  E.g. valid only if all the supervisions directly touch?
Might be more appropriate just to throw if alignments are present if it's not easy to implement?

lhotse/cut.py
pzelasko: It is not super clean; the trick is that if merge() returns an empty string, it trigger `or None` which returns the right hand side arg (and the whole expression evaluates to `None`). I’ll modify it so that `merge` returns Optional[str] instead which should read better.

lhotse/cut.py
pzelasko: It can be anything really; extra pieces of metadata (e.g. accent, text without normalization), numeric values (snr, some scores), array manifests, really anything. I don’t know of a generic way to support all of that here.. Maybe it should be set to `None` instead?

lhotse/cut.py
pzelasko: Yeah there are unit-level timestamps, it should be fine.

lhotse/cut.py
danpovey: My point was that I think when given the empty list, merge() may return "cat", not None.

lhotse/cut.py
danpovey: If we end up using this with custom fields we may need a way to do custom types of merge, but I'd say for now this is probably OK.  We'd probably notice if it was wrong.

lhotse/cut.py
pzelasko: Good point, fixed.

lhotse/cut.py
pzelasko: I added a parameter `custom_merge_fn` that can be used to customize this merging behavior, and retained the current code as the default implementation.

PR #504, opened 2021-12-03
Opensmile wrapper

marcinwitkowski:
Hi!

I have implemented a feature extractor which wraps [python-opensmile](https://audeering.github.io/opensmile-python/) that might be useful for people working on emotion recognition. At the moment it provides essentially using [the predefined OpenSmile configs](https://audeering.github.io/opensmile-python/api-smile.html#featureset).

Few notes:
1. OpenSmile allows for extraction of so called `Functional` features and `LowLevelDescriptors`, which are referred as feature levels. Functionals are statistics computed using an entire signal, and LLDs are computed frame-wise with `frame_shift` equal to 10ms for all predefined settings. As such, I decided to hardcode this value in `frame_shift()` property for now. Because Functionals might be considered as an embedding from entire utterance, I'm looking forward to any ideas how to set `frame_shift` in this case. 
2. Adding custom OpenSmile config would require its parsing and due to the fact that one config may include multiple "framers", it is not trivial to get single `frame_shit` property without "understanding" the entire config. Hence, for now it can be assumed that implementation does not support custom configs.
3. I have added some tests in test/features/test_opensmile.py, so please take a look at use examples there.
4. This extractor requires [opensmile-python](https://github.com/audeering/opensmile-python/tree/v2.2.0). However, I did some efforts to make this package optional for all users which do not want to use this  feature extractor.

Since this is my first contribution to Lhotse, I will highly appreciate any comments on that.

In close future we plan to run some emotion recognition experiments with Lhotse, so soon we will probably provide an initial implementation of EmotionRecognitionDataset and or recipies for publicly available corpora. 

Kind regards and stay healthy, 
MW


pzelasko:
Thanks, this is very cool! I'll review the code in detail shortly. Responding to your questions:

> 1. OpenSmile allows for extraction of so called `Functional` features and `LowLevelDescriptors`, which are referred as feature levels. Functionals are statistics computed using an entire signal, and LLDs are computed frame-wise with `frame_shift` equal to 10ms for all predefined settings. As such, I decided to hardcode this value in `frame_shift()` property for now. Because Functionals might be considered as an embedding from entire utterance, I'm looking forward to any ideas how to set `frame_shift` in this case.

I think it's fine to hardcode, and if somebody needs a non-default configuration, they can always take it from here. 

For "Functionals" you might want to use recently added custom attribute support in Lhotse. It allows you to create manifests for as many arrays as you'd like and attach them to supervisions or cuts to be read later. I am going to add a description of it in docs soon, for now you can see the examples of usage e.g. here: https://github.com/lhotse-speech/lhotse/blob/551a18f00cc43ff192daf474b0347f09096642fb/test/cut/test_custom_attrs.py#L18-L29 

This allows you to attach more than one type of feature to a cut. We also support auto-padding/batching of custom fields with `collate_custom_field` https://github.com/lhotse-speech/lhotse/blob/551a18f00cc43ff192daf474b0347f09096642fb/test/dataset/test_collation.py#L118-L133 (the best place to call it is inside your `Dataset` object).

> 2. Adding custom OpenSmile config would require its parsing and due to the fact that one config may include multiple "framers", it is not trivial to get single `frame_shit` property without "understanding" the entire config. Hence, for now it can be assumed that implementation does not support custom configs.

👍🏻 

> In close future we plan to run some emotion recognition experiments with Lhotse, so soon we will probably provide an initial implementation of EmotionRecognitionDataset and or recipies for publicly available corpora.

Wonderful!

lhotse/features/opensmile.py
```diff
@@ -0,0 +1,149 @@
...
+        Returns list of strings with names of pretrained FeatureSets available in opensmile.
+        """
+        assert is_module_available("opensmile"), \
+         'To use opensmile extractors, please "pip install opensmile" first.'
+        import opensmile
+        return list(opensmile.FeatureSet.__members__)
+
+
+@register_extractor
+class OpenSmileWrapper(FeatureExtractor):
```
pzelasko, line 55: How about `OpenSmileExtractor`? Most of Lhotse extractors are actually wrappers :) same for the `name` attribute

lhotse/features/opensmile.py
```diff
@@ -0,0 +1,149 @@
...
+    """Wrapper for extraction of features implemented in OpenSmile."""
+
+    name = "opensmile-wrapper"
+    config_type = OpenSmileConfig
+    smileExtractor = None
```
pzelasko, line 60: I think this line can be removed because you're always setting this field in the constructor.

lhotse/features/opensmile.py
```diff
@@ -0,0 +1,149 @@
...
+                Defined featureset: {}'.format(self.config.feature_set))                
+                
+    def feature_dim(self, sampling_rate: int) -> int:
+        return len(self.feature_names)
+        
+    def feature_names(self) -> List[str]:
+        return self.smileExtractor.feature_names()
+
+    def extract(self, samples: np.ndarray, sampling_rate: int) -> np.ndarray:    
+        assert sampling_rate == self.config.sampling_rate
```
pzelasko, line 116: the config comment said `self.config.sampling_rate` can also be None, but it would crash here.

lhotse/features/opensmile.py
```diff
@@ -0,0 +1,149 @@
...
+        import opensmile
+        
+        feats = self.smileExtractor.process_signal(samples, sampling_rate=sampling_rate).to_numpy()
+        # Add last diff frames to the end of feats matrix to fit lhotse.utils.compute_num_frames
+        if self.is_lld_or_lld_de():
+            duration = np.shape(samples)[1]/sampling_rate
+            diff = compute_num_frames(duration, self.frame_shift, sampling_rate) - np.shape(feats)[0]
```
pzelasko, line 123: Is it worth checking if that value is too large (e.g. have a tolerance of +/- 5 frames)? It could indicate a more serious error.

test/features/test_opensmile.py
```diff
@@ -0,0 +1,108 @@
+from tempfile import NamedTemporaryFile
+import numpy as np
+import pytest
+from lhotse import Recording, OpenSmileWrapper, OpenSmileConfig
+from lhotse.utils import is_module_available, compute_num_frames, compute_num_samples
+
```
pzelasko, line 6: you can add the following line and remove all skipif marks and local imports:

```python
opensmile = pytest.importorskip("opensmile", reason=...)
```

pzelasko (commented):
Thanks again, looks good to me! Left some comments.

lhotse/features/opensmile.py
marcinwitkowski: Agree. Done.

lhotse/features/opensmile.py
marcinwitkowski: Agree. Done.

test/features/test_opensmile.py
marcinwitkowski: Agree. Done.

lhotse/features/opensmile.py
marcinwitkowski: Actually in the test there were examples with difference equal to 5. This is the result of difference in treatment of the last frame_len between Lhotse and OpenSmile. Now I print warning if abs(diff)>5.

lhotse/features/opensmile.py
marcinwitkowski: You're right. I changed the default value to None and added raising of a `ValueError` instead of assertion.

marcinwitkowski:
Thank you for immediate review. I addressed all issues in the last commit.
Once I processed files with black, I realized it will be probably hard to analyze diffs. Sorry for that. 

I need a while to check `Array`. When I figure out how to use it in elegant way to extract "Functionals", I will definietely let you know.

lhotse/features/opensmile.py
```diff
@@ -0,0 +1,180 @@
...
+    For now custom configs are not supported in this extractor.
+    """
+
+    feature_set: Union[str, Any] = "ComParE_2016"  # default feature set or
+    # string with set name
+    feature_level: Union[str, Any] = "lld"  # default feature level or level name
+    options: dict = None  # dictionary with optional script parameters
+    loglevel: int = 2  # log level (0-5), the higher the number the more log
+    # messages are given
+    logfile: str = None  # if not ``None`` log messages will be stored to this file
```
csukuangfj, line 32:
```suggestion
    logfile: Optional[str] = None  # if not ``None`` log messages will be stored to this file
```
And also for other fields with default value `None`.

lhotse/features/opensmile.py
```diff
@@ -0,0 +1,180 @@
...
+            warnings.warn(f"Unusual difference in number of frames: {diff}")
+        if diff > 0:
+            feats = np.append(feats, feats[-diff:, :], axis=0)
+        elif diff < 0:
+            feats = feats[:-diff, :]
+        return feats
+
+
+#    @staticmethod
+#    def mix(
```
csukuangfj, line 165: Shall we remove such blocks?
If we need it later, we can always find it in the git history.

pzelasko:
Can you also edit the unit tests workflow file in `.github` so that opensmile is installed as a part of CI? It will make your tests run continuously.

pzelasko:
See here: https://github.com/lhotse-speech/lhotse/blob/master/.github/workflows/unit_tests.yml#L41

marcinwitkowski:
> Can you also edit the unit tests workflow file in `.github` so that opensmile is installed as a part of CI? It will make your tests run continuously.

Done.

pzelasko:
Thanks! Great first contribution!

PR #508, opened 2021-12-07
Resumable batch feature extraction with reduced memory usage

pzelasko:
I haven't tested it yet. This should greatly reduce RAM usage for extracting features from large manifests. It also allows to resume a previously interrupted run from where it left off. Most of the functionality was in place in Lhotse, just needed to be glued together.

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -4158,7 +4158,7 @@ class CutSet(Serializable, Sequence[Cut]):
                             features=feat_manifest,
                             recording=None,
                         )
-                    cuts_writer.write(cut)
+                    cuts_writer.write(cut, flush=True)
 
                 progress.update(len(cuts))
 
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -4010,6 +4010,7 @@ class CutSet(Serializable, Sequence[Cut]):
         num_workers: int = 4,
         augment_fn: Optional[AugmentFn] = None,
         storage_type: Type[FW] = LilcomHdf5Writer,
+        overwrite: bool = False,
     ) -> "CutSet":
         """
         Extract features for all cuts in batches.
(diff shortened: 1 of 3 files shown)


lhotse/cut.py
```diff
@@ -4060,16 +4067,29 @@ def compute_and_store_features_batch(
         from lhotse.qa import validate_features
 
         frame_shift = extractor.frame_shift
-        dataset = UnsupervisedWaveformDataset(collate=False)
+
+        # We're opening a sequential cuts writer that can resume previously interrupted
+        # operation. It scans the input JSONL file for cut IDs that should be ignored.
+        # Note: this only works when ``manifest_path`` argument was specified, otherwise
+        # we hold everything in memory and start from scratch.
+        cuts_writer = CutSet.open_writer(manifest_path, overwrite=overwrite)
```
csukuangfj, line 4075: Where is `open_writer` defined? I cannot find it.

csukuangfj: I see. It is from its parent.

csukuangfj:
>  It should further help with memory issues for long-recording feature extraction.

Will it still be helpful to reduce the RAM usage if the large manifest, like the one in GigaSpeech, is split into 2000 pieces?
In GigaSpeech, each piece occupies ~20 MB disk space.

---

Will the manifest file be flushed to disk if the process is killed by the system because of OOM? Before this PR, we created the manifest after all the cuts were processed, so there was little chance that the process is killed during the writing.

Also, it becomes a little complicated to detect whether we have finished extracting features for a manifest with this PR.
We used to check the presence of the manifest to achieve that.

pzelasko:
> Will it still be helpful to reduce the RAM usage if the large manifest, like the one in GigaSpeech, is split into 2000 pieces? In GigaSpeech, each piece occupies ~20 MB disk space.

I think it won’t make a difference in this case.

> Will the manifest file be flushed to disk if the process is killed by the system because of OOM? Before this PR, we created the manifest after all the cuts were processed, so there was little chance that the process is killed during the writing.

Good point, I will add the flushing option. 

> Also, it becomes a little complicated to detect whether we have finished extracting features for a manifest with this PR. We used to check the presence of the manifest to achieve that.

With this PR you could run the feature extraction again and it would exit without doing anything I guess. 

Based on your description I think you won’t actually benefit from this as you found a workaround, but at least these changes shouldn’t break your existing workflow in any way.

csukuangfj:
> With this PR you could run the feature extraction again and it would exit without doing anything I guess.

I think it needs to open the file and read all of its contents to check whether all cuts have finished extracting their features.
Before this PR, we only need to check the existence of the file and don't need to read it from disk.

pzelasko:
Fair point. I’ll merge it later anyway but as I wrote earlier, existing code will behave identically as before, so you don’t have to adjust anything.

PR #510, opened 2021-12-09
adding previous utterance for libritts supervisions

oplatek:
one can reconstruct continuous chains of utterances using the supervision

### DIFF
--- a/lhotse/recipes/libritts.py
+++ b/lhotse/recipes/libritts.py
@@ -148,8 +148,8 @@ def prepare_libritts(
             #   84_121123_000008_000000 Villefort rose, half ashamed of being surprised in such a paroxysm of grief.    Villefort rose, half ashamed of being surprised in such a paroxysm of grief.
 
             # book.tsv contains additional metadata
-            utt2snr = {
-                rec_id: float(snr)
+            utt2snr = [
+                (rec_id, float(snr))
                 for rec_id, *_, snr in map(
                     str.split,
                     (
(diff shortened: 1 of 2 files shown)


lhotse/recipes/libritts.py
```diff
@@ -148,8 +148,8 @@ def prepare_libritts(
             #   84_121123_000008_000000 Villefort rose, half ashamed of being surprised in such a paroxysm of grief.    Villefort rose, half ashamed of being surprised in such a paroxysm of grief.
 
             # book.tsv contains additional metadata
-            utt2snr = {
-                rec_id: float(snr)
+            utt2snr = [
```
pzelasko, line 151: I don't understand why this change is needed (you can access utterance ids later with utt2snr.keys())

pzelasko (commented):
I understand it's useful for providing past text context for an utterance (otherwise unavailable from pure supervisions because many utterances have been filtered out in LibriTTS). I think this description should be included somewhere as a comment.

Could we add a flag to make the adding of this new field optional in the manifests? You can use the description above to help document it.

lhotse/recipes/libritts.py
oplatek: I need them sorted as they appear in the file. The order is not kept during initialisation via kwargs arguments what this syntax actually boils down to. Plus Python 3.6 does not support OrderedDict behaviours by default on standard dict anyways only Python 3.7+ does.

pzelasko: Python 3.6 does support that actually. I think they made it a guaranteed behavior in 3.7 but the change was made one release earlier. Anyway we can keep these changes.

pzelasko:
Thanks!

PR #513, opened 2021-12-13
Skipping problematic audios

pkufool:


lhotse/dataset/collation.py
```diff
@@ -163,7 +163,18 @@ def collate_audio(
...
+                    f"AudioLoadingError for {cut} \nError messages: {e} \nSkipping this cut."
+                )
+                continue
+            except DurationMismatchError as e:
+                warnings.warn(
+                    f"DurationMismatchError for {cut} \nError messages: {e} \nSkipping this cut."
+                )
+                continue
     else:
         for idx, example_audio in enumerate(executor.map(_read_audio, cuts)):
```
pkufool, line 179: @pzelasko Any suggestions to handle exceptions in this multithread executor.

pzelasko: hmm, one option is to do the following: ` _read_audio` could return `Union[torch.Tensor, Exception]` and the code in this loop could check `if isinstance(audio_or_error, Exception): raise audio_or_error`

lhotse/dataset/collation.py
```diff
@@ -163,7 +163,18 @@ def collate_audio(
     audio = torch.empty(len(cuts), first_cut.num_samples)
     if executor is None:
         for idx, cut in enumerate(cuts):
-            audio[idx] = _read_audio(cut)
+            try:
+                audio[idx] = _read_audio(cut)
+            except AudioLoadingError as e:
```
pzelasko, line 168: These excepts are getting quite verbose; could you add the following code in `utils.py` and use it here and in other places?


Code: 

```python
class suppress_and_warn:
    """Context manager to suppress specified exceptions that logs the error message.

    After the exception is suppressed, execution proceeds with the next
    statement following the with statement.

         with warn_and_suppress(FileNotFoundError):
             os.remove(somefile)
         # Execution still resumes here if the file was already removed
    """

    def __init__(self, *exceptions):
        self._exceptions = exceptions

    def __enter__(self):
        pass

    def __exit__(self, exctype, excinst, exctb):
        # Returning True from __exit__ in a context manager tells Python
        # to suppress an exception.
        should_suppress = exctype is not None and issubclass(exctype, self._exceptions)
        if should_suppress:
            logging.warning(f"[Suppressed {exctype.__qualname__}] Error message: {excinst}")
        return should_suppress
```

Usage:

```python
with suppress_and_warn(AudioLoadingError, DurationMismatchError):
    audio[idx] = _read_audio(cut)
```

pzelasko: Also note that the current code introduces an error, because the failed utterances are still in the batch just with zeroes. Can you convert `audio` to a list, append to it instead of assigning to specific positions, and later stack it? It's probably less efficient but I doubt it would result in a noticeable slowdown.

pzelasko (commented):
I left some comments. This PR made me think that maybe, as a more general solution, Lhotse needs a "strict" mode switch, that would turn off some assertions and let functions like these continue despite some errors when loading data. I would keep the "strict" mode on by default though to help people not shoot themselves in the foot.

lhotse/dataset/collation.py
```diff
@@ -404,7 +410,14 @@ def read_audio_from_cuts(
     cuts: Iterable[Cut], executor: Optional[Executor] = None
 ) -> List[torch.Tensor]:
     map_fn = map if executor is None else executor.map
-    return list(map_fn(_read_audio, cuts))
+    audios = []
+    for idx, audio_or_none in enumerate(map_fn(_read_audio, cuts)):
+        if audio_or_none is None:
+            del cuts[cuts[idx].id]  # delete the problematic cut
```
pzelasko, line 416: I think this code is technically not correct, as `cuts` may be an `Iterable` (i.e. it may be not indexable). Instead of trying to delete the cuts, I would recommend changing the return type to ``List[Optional[torch.Tensor]]`` and have the caller of this function check each tensor for `None` to see if it was read successfully.

pzelasko: Actually, that won't solve the issue, disregard that remark. It's better to change the `cuts` type annotation to `CutSet`. BTW does `CutSet` support using `del`? I never checked..

pkufool: >  BTW does `CutSet` support using `del`? I never checked..

No, I thought it is a dict.  It doesn't support `del`.

I saw in the docs, said "Operations on cut sets are not mutating...", so should I add a delete operator to it ?

pzelasko: I don't like that option, there has to be a cleaner way to do it. Give me some time to think about it.

pzelasko:
I am closing this one; please see the continuation in #533

PR #516, opened 2021-12-16
bugfix for prepare_librispeech command

rosrad:
bugfix for prepare_librispeech command when the default dataset_parts=auto is used.

bug:
command below will not correctly prepare the manifest of librispeech
lhotse prepare librispeech data/LibriSpeech data/manifests

### DIFF
--- a/lhotse/bin/modes/recipes/librispeech.py
+++ b/lhotse/bin/modes/recipes/librispeech.py
@@ -35,8 +35,8 @@ def librispeech(
     num_jobs: int,
 ):
     """(Mini) Librispeech ASR data preparation."""
-    if len(dataset_parts)==1:
-        dataset_parts=dataset_parts[0]
+    if len(dataset_parts) == 1:
+        dataset_parts = dataset_parts[0]
     prepare_librispeech(
         corpus_dir,
         output_dir=output_dir,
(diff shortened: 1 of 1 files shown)


pzelasko:
Thanks! Can you fix the linter errors?

rosrad:
Tried to fix the errors from black. By the way, do we have some configure for the black and flack check here ?

csukuangfj:
> Tried to fix the errors from black. By the way, do we have some configure for the black and flack check here ?

I think it exists only in the CI, see
- https://github.com/lhotse-speech/lhotse/blob/master/.github/workflows/flake8.yml
- https://github.com/lhotse-speech/lhotse/blob/master/.github/workflows/black.yml

I would recommend adding something like what we're doing in icefall, e.g., https://github.com/k2-fsa/icefall/blob/master/.pre-commit-config.yaml
See also https://icefall.readthedocs.io/en/latest/contributing/code-style.html

rosrad:
@csukuangfj thanks for the inform. it is good to have this pre-commit check. 
I found that in this repo there are different line-length check from black(80) and flake8(127).  @pzelasko  do you think we should align this line-length to be 80 in both black and flake8. And I think we should add pre-commit check as well so that beginner will get the lint error earlier than the PR merge.

pzelasko:
> @csukuangfj thanks for the inform. it is good to have this pre-commit check.
> I found that in this repo there are different line-length check from black(80) and flake8(127). @pzelasko do you think we should align this line-length to be 80 in both black and flake8. And I think we should add pre-commit check as well so that beginner will get the lint error earlier than the PR merge.

Thanks, I think it's a good idea, but I might not have the time to add it right now. Would you be interested in contributing this?

PR #517, opened 2021-12-20
`DynamicBucketingSampler`: on-the-fly bucketing with restricted memory usage

pzelasko:
This is to start getting early feedback on the design of "streaming" or "dynamic" bucketing that doesn't require reading the whole cut set into memory.

The basic idea is to sample N (e.g. ~10k) cuts and estimate the boundary durations for buckets. Then, we maintain a buffer of M cuts (stored separately in K buckets) and every time we sample a batch, we consume the input cut iterable for the same amount of cuts. The memory consumption is limited by M at all times.

Next steps:
- [x] possibly wrap it as a Sampler/CutSampler
- [x]  make it support everything/most things that other samplers are supporting
- [x] make it work with cut pairs

### DIFF
--- a/lhotse/dataset/sampling/__init__.py
+++ b/lhotse/dataset/sampling/__init__.py
@@ -1,6 +1,7 @@
 from .bucketing import BucketingSampler
 from .cut_pairs import CutPairsSampler
 from .data_source import streaming_shuffle
+from .dynamic_bucketing import DynamicBucketingSampler
 from .single_cut import SingleCutSampler
 from .utils import find_pessimistic_batches
 from .zip import ZipSampler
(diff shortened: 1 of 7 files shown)


danpovey:
The basic approach sounds good!

pzelasko:
I think it should be good enough to give it a test run with some real data. I think the only feature not supported yet is saving and restoring sampler state; it can be added later by saving the index of the last read line of the input files, and all the cuts that are held in the reservoir sampling buffer and bucketing buffers (since these do not necessarily correspond to the last M read lines minus buffer_size if we used shuffling and filters).

pzelasko:
Tested on manifests for 500h iterated lazily from S3, seems to work just fine. Merging!

pzelasko:
Fun fact: on that machine, importing Lhotse took 260MB of memory. Iterating through the whole sampler raised mem usage to 440MB (so the lazy reading, buffering, and iteration took just 180MB of memory). Reading the manifest eagerly and iterating with the old `BucketingSampler` required 1GB of memory.

PR #521, opened 2021-12-22
GlobalMVN and Wav2LogFilterBank torchscriptable

janvainer:


lhotse/features/kaldi/layers.py
```diff
@@ -506,25 +510,12 @@ def __init__(
...
-            warnings.warn(
-                "Could not import torch.cuda.amp.autocast -- "
-                "when using mixed precision with another package such as apex, "
-                "you might experience numerical stability issues."
-            )
-            pow_spec = torch.matmul(pow_spec.float(), self._fb.float())
-        else:
-            with autocast(enabled=False):
-                pow_spec = torch.matmul(pow_spec.float(), self._fb.float())
+        pow_spec = torch.matmul(pow_spec.float(), self._fb.float())
```
pzelasko, line 518: @jesus-villalba do you think this change could be dangerous with fp16?

lhotse/features/kaldi/layers.py
```diff
@@ -160,6 +160,9 @@ def forward(
                 x, self._length, self._shift, self.snip_edges
             )
             log_energy = _get_log_energy(x_strided, self.energy_floor)  # size (m)
+        else:
+            # placeholder for torchscript
+            log_energy = torch.zeros_like(x)
```
pzelasko, line 165: You can use `None` if you explicitly annotate the types;

function output -> `Tuple[torch.Tensor, Optional[torch.Tensor]]`
variable -> `log_energy: Optional[torch.Tensor] = None`

lhotse/dataset/signal_transforms.py
```diff
@@ -40,7 +40,7 @@ def from_file(cls, stats_file: Pathlike) -> "GlobalMVN":
     def to_file(self, stats_file: Pathlike):
         torch.save(self.state_dict(), stats_file)
 
-    def forward(self, features: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+    def forward(self, features: torch.Tensor) -> torch.Tensor:
```
pzelasko, line 43: can you add a `supervision_segments: Optional[torch.IntTensor] = None,` arg here to make it compatible with existing setups?

pzelasko (commented):
Looks good, can you add a test that runs `torch.jit.script` on each of these modules and forwards some dummy tensor through the jit modules?

lhotse/features/kaldi/layers.py
jesus-villalba: That multiplication only gave me trouble when the signal is in (-2^15, 2^15). If the signal is in (-1,1) I think fp16 worked fine

lhotse/dataset/signal_transforms.py
janvainer: adding

lhotse/features/kaldi/layers.py
janvainer: Hm so can I leave it as is @pzelasko ?

lhotse/features/kaldi/layers.py
janvainer: adding

lhotse/features/kaldi/layers.py
pzelasko: Yeah

pzelasko (approved):
Thanks!!

Issue #518, opened 2021-12-21
RAM keep rising during training


tz301:
I have more than 10000h audio and store features on disk and using lazy format "cuts.jsonl.gz".

During training, I use K2SpeechRecognitionDataset and SingleCutSampler, and set --world-size=4 and --num-workers=1, memory can be seen below. OOM happened and can not train for one epoch.

<img width="477" alt="1" src="https://user-images.githubusercontent.com/35760938/146870980-28b143ff-a0a1-4f8d-8f71-d5dbeecad4e0.png">

If I set --world-size=1 (not using DistributedDataParallel) and --num-workers=5, main process seems to have stable memory, but all dataloader workers also seem to show rising memory. 

<img width="1154" alt="2" src="https://user-images.githubusercontent.com/35760938/146871119-7d142668-76f9-46e5-ad56-f99dcd92a8a8.png">

<img width="378" alt="3" src="https://user-images.githubusercontent.com/35760938/146871123-84fe49d3-ad19-43d4-b628-47bef6075e9b.png">

csukuangfj:
How about disabling the caching mechanism in Lhotse?
That is, use
```python3
from lhotse import set_caching_enabled
set_caching_enabled(False)
```

There is some kind of caching while reading features, see
https://github.com/lhotse-speech/lhotse/blob/5e99fb2fc250ac9f3c335dd75cfde6ccd22cd8a9/lhotse/features/io.py#L243-L244

https://github.com/lhotse-speech/lhotse/blob/5e99fb2fc250ac9f3c335dd75cfde6ccd22cd8a9/lhotse/caching.py#L37-L44

If the size of your feature file is large, your RAM usage will go up as the training proceeds.

tz301:
> How about disabling the caching mechanism in Lhotse? That is, use
> 

That occurs in lhotse==0.11.0, seems no cache is used at that version.

https://github.com/lhotse-speech/lhotse/blob/6d60f333da89abc3c2836f395f95ae9d53ca5cd0/lhotse/features/io.py#L184-L192

I also update to lhotse==0.12.0 and disable the caching mechanism at the top of train.py. Memory still rising now.

![12](https://user-images.githubusercontent.com/35760938/146885798-bfaa4030-401b-4714-8899-adf7e82a3ff2.png)

csukuangfj:
Then how about deleting the following line
[...]

PR #522, opened 2021-12-22
Custom binary feature storage format

pzelasko:
Since HDF5 might not be the best choice for this project (we'll see how the discussion proceeds in #518), here is an alternative that doesn't use it. 

It's a copy of `ChunkedLilcomHdf5Writer` that uses a plain binary file and stores the compressed array chunks next to each other. The `storage_key` keeps the global file offset for each array followed by a list of chunk relative offsets. 

Seems to work on par with HDF5 storage we had so far: practically identical file sizes, seems to have roughly the same iteration speed when reading shuffled feats. 

@danpovey you have experience with building formats like these in Kaldi (`ark`), any thoughts here?

### DIFF
--- a/lhotse/bin/modes/manipulation.py
+++ b/lhotse/bin/modes/manipulation.py
@@ -1,5 +1,7 @@
 import json
 import os
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from itertools import groupby
 from pathlib import Path
 from typing import Optional
 
@@ -43,11 +45,22 @@ def copy(input_manifest, output_manifest):
     default="lilcom_chunky",
     help="Which storage backend should we use for writing the copied features.",
 )
+@click.option(
+    "-j",
+    "--max-jobs",
+    default=-1,
+    type=int,
+    help="Maximum number of parallel copying processes. "
+    "By default, one process is spawned for every existing feature file in the "
+    "INPUT_MANIFEST (e.g., if the features were extracted with 20 jobs, "
+    "there will typically be 20 files).",
+)
 def copy_feats(
     input_manifest: Pathlike,
     output_manifest: Pathlike,
     storage_path: str,
     storage_type: str,
+    max_jobs: int,
 ) -> None:
     """
     Load INPUT_MANIFEST of type :class:`lhotse.FeatureSet` or `lhotse.CutSet`,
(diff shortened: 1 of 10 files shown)


danpovey:
Oh- this seems pretty cool.
So it's just about removing the hdf5 dependency? 
Do we use hdf5 much otherwise?

pzelasko:
It’s about providing the same functionality without incurring a potentially large memory cost due to some opaque caching schemes of HDF5. So far at least two groups noticed constantly growing CPU RAM during each epoch and I pinpointed that to be HDF5 related. 

I’m going to keep the h5py dependency for now to provide backward compatibility — the users will be able to use the features they have already precomputed. I might deprecate the HDF5 writers as soon as I’ve convinced myself that the new storage type is stable.

danpovey:
Great!

danpovey:
Piotr, do you think it's possible that as a quick fix for the constantly growing memory in hdf5, we could periodically just close all filehandles, e.g. every 100 batches?  I'm not sure how easy this might be to do.
It might fix the memory-growing issue without requiring to re-dump data.

pzelasko:
> Piotr, do you think it's possible that as a quick fix for the constantly growing memory in hdf5, we could periodically just close all filehandles, e.g. every 100 batches? I'm not sure how easy this might be to do.
> It might fix the memory-growing issue without requiring to re-dump data.

@danpovey I wrote an example showing how to do it; I'm not sure if it makes sense to merge this? For now I marked it as not for merge.
https://github.com/lhotse-speech/lhotse/pull/527

danpovey:
There is a conflict.

I suppose we could test this (but I don't know if we keep track of memory usage, so the only clear signal would be a training process not being killed, which is unpredictable anyway).

IMO there wouldn't be much downside in periodically closing filehandles if it's over a long enough period.  I doubt re-opening files once every 100 batches will really impact performance much.  Anyway the OS can do its caching in memory.

pzelasko:
I fixed the conflict on this branch, but that fix for HDF5 memory is in another PR, here: https://github.com/lhotse-speech/lhotse/pull/527

When I was debugging these issues, I tested this solution previously with a small script that just iterates the dataloader and does nothing with the data. The more often the files are closed, the less the memory seems to be used. The worst case I/O slowdown was 50% when closing the files after every mini-batch, otherwise it didn't seem to drastically degrade.

danpovey:
Ah.  So it seems to me that #527 could be merged with essentially no downside??
I know it's an ugly thing, but it seems to me that it resolves the problem and doesn't have negative effects.

pzelasko:
When you put it like that it makes sense 😉 I’ll add a comment why this code is here later and merge

pzelasko:
This thing seems to work with no memory/size/speed issues, which is now validated by someone else than me; merging.

PR #523, opened 2022-01-05
Streaming Kaldi feature extractors

pzelasko:
This updates the Kaldi-like feature extractor layers with a new function `online_inference`. It accepts an extra argument which is the remainder of the previous waveform (or `None`) and it returns the remainder from the currently processed audio chunk. It works basically the same like Kaldi's online feature extraction, except it has a stateless API which plays better with TorchScript and other runtimes. The callers should manage the state themselves.

### DIFF
--- a/lhotse/features/kaldi/layers.py
+++ b/lhotse/features/kaldi/layers.py
@@ -680,7 +680,10 @@ def _get_strided_batch(waveform, window_length, window_shift, snip_edges):
         npad_right = npad - npad_left
         # waveform = nn.functional.pad(waveform, (npad_left, npad_right), mode='reflect')
         pad_left = torch.flip(waveform[:, 1 : npad_left + 1], (1,))
-        pad_right = torch.flip(waveform[:, -npad_right - 1 : -1], (1,))
+        if npad_right >= 0:
+            pad_right = torch.flip(waveform[:, -npad_right - 1 : -1], (1,))
+        else:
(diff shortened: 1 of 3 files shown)


pzelasko:
I was able to replicate the WER of a pretrained ASR model after these modifications.

lhotse/features/kaldi/layers.py
```diff
@@ -13,10 +13,17 @@
 
 This implementation works well with autograd and batching, and can be used neural network
 layers.
+
+Update January 2022:
+These modules now expose a new API function called "online_inference" that
+may be used to compute the features when the audio is streaming.
+The implementation is stateless, and passes the waveform remainders
+back to the user to feed them to the modules once new data becomes available.
+The implementation is compatible with JIT scripting via TorchScript.
```
csukuangfj, line 22: Are there tests to ensure this?

pzelasko: Yeah, this PR #521 added torchscript compatibility and tests for it, and in this PR the new methods are annotated with `torch.jit.export`, so they're automatically tested by the previously existing tests (I also manually torchscripted it and ran some forward passes to make sure).

danpovey:
A great feature!

pzelasko:
I refined it to give the same results as the torchaudio extractors with the default settings used in Lhotse. Merging.

PR #524, opened 2022-01-06
Add early reverb option for RIR-based augmentation

desh2608:
When this option is set to `true`, the audio is convolved with only the first 50ms of the impulse response.


pzelasko:
Cool! Can you add a test for either a recording or a cut, and explicitly mention 50ms in the docs of the `early_only` param? Thanks!

desh2608:
> Cool! Can you add a test for either a recording or a cut, and explicitly mention 50ms in the docs of the `early_only` param? Thanks!

Done!

pzelasko:
Thanks!

PR #526, opened 2022-01-07
ICSI meeting corpus

desh2608:
This PR adds a recipe for the ICSI meeting corpus. The recipe is similar to that for AMI, and provides the option to prepare `ihm`, `ihm-mix`, `sdm`, and `mdm` mics.

NOTE: ICSI does not have an official train/dev/test split, so we follow previous work (from @pswietojanski) and assign {Bmr021, Bns00} to `dev` and {Bmr013, Bmr018, Bro021} to `test`.

### DIFF
--- a/docs/corpus.rst
+++ b/docs/corpus.rst
@@ -85,6 +85,8 @@ a CLI tool that create the manifests given a corpus directory.
     - :func:`lhotse.recipes.prepare_heroico`
   * - HiFiTTS
     - :func:`lhotse.recipes.prepare_hifitts`
+  * - ICSI
+    - :func:`lhotse.recipes.prepare_icsi`
   * - L2 Arctic
     - :func:`lhotse.recipes.prepare_l2_arctic`
   * - LibriCSS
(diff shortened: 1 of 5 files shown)


pzelasko:
Thanks!! 

I wonder what is the difference between the LDC ICSI distribution and the one available here: https://groups.inf.ed.ac.uk/ami/icsi/

The latter one has the benefit that everyone can download it for free (I don't think that's true with LDC?)

desh2608:
> Thanks!!
> 
> I wonder what is the difference between the LDC ICSI distribution and the one available here: https://groups.inf.ed.ac.uk/ami/icsi/
> 
> The latter one has the benefit that everyone can download it for free (I don't think that's true with LDC?)

Oh that's a good point. I will check if all the data is available there and add a `download` function.

desh2608:
@pzelasko I have updated the recipe to use the publicly available version, which is, in fact, better than the LDC version since it also contains a Mix-Headset audio and more consistent session naming.

pzelasko:
Great! Thanks!

PR #527, opened 2022-01-09
Possible workaround for HDF5 growing memory issue

pzelasko:
Intended as a workaround for growing memory issues when using HDF5 with larger datasets.

### DIFF
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -3,7 +3,7 @@ from typing import Callable, Dict, List, Union
 import torch
 from torch.utils.data.dataloader import DataLoader, default_collate
 
-from lhotse import validate
+from lhotse import close_cached_file_handles, validate
 from lhotse.cut import CutSet
 from lhotse.dataset.input_strategies import BatchIO, PrecomputedFeatures
 from lhotse.utils import compute_num_frames, ifnone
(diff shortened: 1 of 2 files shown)


lhotse/dataset/speech_recognition.py
```diff
@@ -92,6 +93,10 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List[str]]]
         """
         validate_for_asr(cuts)
 
+        if self.batch_counter > 0 and self.batch_counter % 100 == 0:
+            close_cached_file_handles()
```
csukuangfj, line 97: Shall we reset `self.batch_counter` in case it overflows if it runs for enough time?

danpovey: Python uses special integers that can never overflow, I believe.

pzelasko:
I cleaned up the code and added some docs. I'd appreciate if somebody can test it out on a large scale training and confirm that it does help; I was only able to test on small-to-medium sized problems so far.

pzelasko:
In case anybody reads this: this workaround doesn't completely solve the issue, and the memory keeps growing, just slower. There also seem to be random spikes in memory usage around the time the HDF5 files are open/closed. 

I generally don't advise using LilcomHdf5Writer for data larger than ~1k hours as it will likely blow up the memory at some point.

desh2608:
@pzelasko which is the recommended storage type for large data (5-10k hours)?

pzelasko:
The default one (`LilcomChunkyWriter`) should work as well as HDF5 but without memory leaks.

cyrta:
And what about cloud storage ?

pzelasko:
The recent PRs with WebDataset integration work really well with that; please refer to #582, #599, and #602. You'll need to read the documentation in the code because we don't have a high-level tutorial for these workflows yet.

pzelasko:
... alternatively, there is also `LilcomUrlWriter` for features and `AudioSource(type="url", ...)` for audio which you can use with any cloud storage supported by `smart_open` library. It will be much less I/O efficient because it will spawn a new connection for each item, but for some types of use-cases might be viable. But WebDataset-based flow would be roughly 50-100x faster.

PR #529, opened 2022-01-10
The People's Speech recipe

pzelasko:


lhotse/audio.py
```diff
@@ -317,10 +318,16 @@ def from_file(
             instead of the one we read from the manifest. This is useful for OPUS files that always
             have 48kHz rate and need to be resampled to the real one -- we will perform that operation
             "under-the-hood". For non-OPUS files this input is undefined.
+        :param force_read_audio: Set it to ``True`` for audio files that do not have any metadata
+            in their headers (e.g., "The People's Speech" FLAC files).
```
pzelasko, line 322: To clarify: we're not forcing to read audio for People's Speech in this recipe (we got away with reading the raw distributions manifests that have the info we need); just creating an option to handle audio files with incomplete metadata (I've never before encountered FLAC files that were missing the number of samples/duration in their headers).

PR #531, opened 2022-01-12
Fix bug in export to kaldi function

HuangZiliAndy:
Hi @desh2608 @jtrmal @pzelasko , I meet some problems when I am converting a meeting directory to Kaldi. The function is giving an incorrect number of segments. Thanks!

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -206,12 +206,6 @@ def export_to_kaldi(
             )
         )
 
-    ## Create a simple CutSet that ties together
-    ## the recording <-> supervision information.
-    #cuts = CutSet.from_manifests(
-    #    recordings=recordings, supervisions=supervisions
-    #).trim_to_supervisions()
-
     if all(r.num_channels == 1 for r in recordings):
         # if all the recordings are single channel, we won't add
         # the channel id affix to retain back compatibility
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -206,11 +206,11 @@ def export_to_kaldi(
             )
         )
 
-    # Create a simple CutSet that ties together
-    # the recording <-> supervision information.
-    cuts = CutSet.from_manifests(
-        recordings=recordings, supervisions=supervisions
-    ).trim_to_supervisions()
+    ## Create a simple CutSet that ties together
+    ## the recording <-> supervision information.
+    #cuts = CutSet.from_manifests(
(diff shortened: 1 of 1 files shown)


pzelasko:
Why are you losing segments if you convert recording + supervisions to cuts? It would indicate that those segments either belong to non-existent recordings or are past the recording's end time..

HuangZiliAndy:
Hi Piotr, I don't think so. I think some are overlapping segments. I tried with another dataset and lost almost half of the segments. Let me try to find an example.

HuangZiliAndy:
I updated in the doc above. Thanks

pzelasko:
Ah, of course, now I see what is the issue. In segments like the following:

```
|-------seg1--------|
       |---seg2---|
```

`seg2` would be never written. Possibly the same thing happens with partially overlapping segments.

 I think your changes make sense now. Can you resolve the conflicts?

lhotse/kaldi.py
```diff
@@ -206,11 +206,11 @@ def export_to_kaldi(
...
-    # Create a simple CutSet that ties together
-    # the recording <-> supervision information.
-    cuts = CutSet.from_manifests(
-        recordings=recordings, supervisions=supervisions
-    ).trim_to_supervisions()
+    ## Create a simple CutSet that ties together
+    ## the recording <-> supervision information.
+    #cuts = CutSet.from_manifests(
+    #    recordings=recordings, supervisions=supervisions
+    #).trim_to_supervisions()
```
pzelasko, line 213: Can you remove the commented code?

pzelasko:
Once you resolve the conflicts with master, we can merge it.

lhotse/kaldi.py
```diff
@@ -231,8 +225,8 @@ def export_to_kaldi(
         # segments
         save_kaldi_text_mapping(
             data={
-                cut.supervisions[0].id: f"{cut.recording_id} {cut.start} {cut.end}"
-                for cut in cuts
+                sup.id: f"{sup.recording_id} {sup.start} {sup.start + sup.duration}"
```
desh2608, line 228: You can also just use `sup.end` (`SupervisionSegment` has an `end()` [property](https://github.com/lhotse-speech/lhotse/blob/b41e4f8c57954b94790f2d05de310a11071512bc/lhotse/supervision.py#L187).

pzelasko:
Thanks!

PR #532, opened 2022-01-12
Feature/unsup multichan waveform dataset

kkarrancsu:
Added Dataset which supports multichannel audio-samples.  Updated collator to drop pad-tracks.

### DIFF
--- a/lhotse/dataset/collation.py
+++ b/lhotse/dataset/collation.py
@@ -7,7 +7,7 @@ import torch
 from torch.nn import CrossEntropyLoss
 
 from lhotse import CutSet
-from lhotse.cut import Cut, MixedCut
+from lhotse.cut import Cut, MixedCut, PaddingCut
 from lhotse.utils import DEFAULT_PADDING_VALUE
 
 
@@ -312,6 +312,44 @@ def collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
     audio = torch.empty(len(cuts), len(first_cut.tracks), first_cut.num_samples)
     for idx, cut in enumerate(cuts):
         audio[idx] = torch.from_numpy(cut.load_audio())
+    return audio
+
+
+def remove_pad_tracks(cuts):
+    for cut in cuts:
+        tracks = cut.tracks
+        tracks_nopad = [t for t in tracks if not isinstance(t.cut, PaddingCut)]
+        cut.tracks = tracks_nopad
+    return cuts
+
+
+def collate_multi_channel_audio_rmpad(cuts: CutSet) -> torch.Tensor:
+    """
+    Load audio samples for all the cuts and return them as a batch in a torch tensor.
+    The cuts have to be of type ``MixedCut`` and their tracks will be interpreted as individual channels.
+    The output shape is ``(batch, channel, time)``.
+    The cuts will be padded with silence if necessary.
+    """
+    assert all(cut.has_recording for cut in cuts)
+    assert all(isinstance(cut, MixedCut) for cut in cuts)
+    
+    # TODO: how to ensure that each track is synced across batches?  i.e. dim=1 is the track index
+    #  and should correspond to the same mic across batches 
+
+    cuts = maybe_pad(cuts)
+    cuts = remove_pad_tracks(cuts)
+    
+    # NOTE: what to do when the # of tracks is not the same across cuts, right now
+    #  this is zero-padding but that seems bad ...
+    max_ntrack = max(len(cut.tracks) for cut in cuts)
+    max_nsamp = max(cut.num_samples for cut in cuts)
+
+    # NOTE: this ends up zero-padding! is that appropriate?
+    audio = torch.zeros(len(cuts), max_ntrack, max_nsamp) 
+    for idx, cut in enumerate(cuts):
+        ntrack = len(cut.tracks)
+        nsamp = cut.num_samples
+        audio[idx, 0:ntrack, 0:nsamp] = torch.from_numpy(cut.load_audio(mixed=False))
     return audio
 
 
--- a/lhotse/dataset/speech_recognition.py
+++ b/lhotse/dataset/speech_recognition.py
@@ -90,7 +90,7 @@ class K2SpeechRecognitionDataset(torch.utils.data.Dataset):
         Return a new batch, with the batch size automatically determined using the constraints
         of max_frames and max_cuts.
         """
-        validate_for_asr(cuts)
+        #validate_for_asr(cuts)
 
         # Sort the cuts by duration so that the first one determines the batch time dimensions.
         cuts = cuts.sort_by_duration(ascending=False)
--- a/lhotse/dataset/unsupervised.py
+++ b/lhotse/dataset/unsupervised.py
@@ -5,7 +5,7 @@ import torch
 from lhotse import validate
 from lhotse.augmentation import AugmentFn
 from lhotse.cut import CutSet
-from lhotse.dataset.collation import collate_audio, collate_features, collate_matrices
+from lhotse.dataset.collation import collate_audio, collate_features, collate_matrices, collate_multi_channel_audio_rmpad
 from lhotse.features import FeatureExtractor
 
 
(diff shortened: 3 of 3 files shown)


lhotse/dataset/collation.py
```diff
@@ -315,6 +315,44 @@ def collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
...
+    The output shape is ``(batch, channel, time)``.
+    The cuts will be padded with silence if necessary.
+    """
+    assert all(cut.has_recording for cut in cuts)
+    assert all(isinstance(cut, MixedCut) for cut in cuts)
+    
+    # TODO: how to ensure that each track is synced across batches?  i.e. dim=1 is the track index
+    #  and should correspond to the same mic across batches 
+
+    cuts = maybe_pad(cuts)
```
pzelasko, line 339: This should not be needed as you're manually zero-padding later.

lhotse/dataset/collation.py
```diff
@@ -315,6 +315,44 @@ def collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
...
+    cuts = remove_pad_tracks(cuts)
```
pzelasko, line 340: I think there is a pitfall here, what if a MixedCut looks like:

```
|-------cut1-------||---padding---||----cut2----|
```

or any variation of the situation where padding is in between of two cuts. I don't think Lhotse would handle these situations well with your current code. Maybe you should try only removing the padding at the end (and beginning, but for that one you have to be careful about modifying the offsets on the remaining tracks). Rather than manually removing PaddingCuts, I suggest using `.truncate()` with carefully computed offset and duration arguments; that method will handle a lot of pitfalls and edge-cases.

lhotse/dataset/collation.py
```diff
@@ -315,6 +315,44 @@ def collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
...
+    #  this is zero-padding but that seems bad ...
+    max_ntrack = max(len(cut.tracks) for cut in cuts)
+    max_nsamp = max(cut.num_samples for cut in cuts)
+
+    # NOTE: this ends up zero-padding! is that appropriate?
+    audio = torch.zeros(len(cuts), max_ntrack, max_nsamp) 
+    for idx, cut in enumerate(cuts):
+        ntrack = len(cut.tracks)
+        nsamp = cut.num_samples
+        audio[idx, 0:ntrack, 0:nsamp] = torch.from_numpy(cut.load_audio(mixed=False))
```
pzelasko, line 352: Note that if you did `cut.mix(musan_cut)` here, it will also add an extra track; as is, the code would not work with additive noise data augmentation

lhotse/dataset/collation.py
```diff
@@ -315,6 +315,44 @@ def collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
...
+    assert all(isinstance(cut, MixedCut) for cut in cuts)
+    
+    # TODO: how to ensure that each track is synced across batches?  i.e. dim=1 is the track index
+    #  and should correspond to the same mic across batches 
+
+    cuts = maybe_pad(cuts)
+    cuts = remove_pad_tracks(cuts)
+    
+    # NOTE: what to do when the # of tracks is not the same across cuts, right now
+    #  this is zero-padding but that seems bad ...
```
pzelasko, line 343: I think you won't escape zero-padding of examples with less channels if you need to collate the data. However, I suggest you modify this function to return a 3-tuple: `(audio, audio_lens, channel_indexes)` where `audio` is the collated data with shape `(B, C, T)`, `audio_lens` has the length of each multi-channel example of shape `(B,)`, and `channel_indexes` is a list of lists of which `C` dim indexes have meaningful channels for examples (it could also be `channel_lens` tensor of shape `(B,)` assuming first `c` channels are always meaningful, if it's possible to guarantee).

pzelasko: But in the end your models will have to somehow deal with the non-meaningful channels anyway. As long as you're working on same-number-of-channels data, no need to overthink this.

lhotse/dataset/collation.py
```diff
@@ -315,6 +315,44 @@ def collate_multi_channel_audio(cuts: CutSet) -> torch.Tensor:
...
+    Load audio samples for all the cuts and return them as a batch in a torch tensor.
+    The cuts have to be of type ``MixedCut`` and their tracks will be interpreted as individual channels.
+    The output shape is ``(batch, channel, time)``.
+    The cuts will be padded with silence if necessary.
+    """
+    assert all(cut.has_recording for cut in cuts)
+    assert all(isinstance(cut, MixedCut) for cut in cuts)
+    
+    # TODO: how to ensure that each track is synced across batches?  i.e. dim=1 is the track index
+    #  and should correspond to the same mic across batches
```
pzelasko, line 337: You can ensure the tracks are sorted by some property; I imagine this is something very corpus specific and should be done by the user, not by the library.

lhotse/dataset/unsupervised.py
```diff
@@ -74,6 +74,41 @@ def _validate(self, cuts: CutSet) -> None:
         assert all(cut.has_recording for cut in cuts)
 
 
+class UnsupervisedMultiChanWaveformDataset(UnsupervisedDataset):
```
pzelasko, line 77:
```suggestion
class MultiChannelWaveformDataset(UnsupervisedDataset):
```

somehow reads better to me

lhotse/dataset/unsupervised.py
```diff
@@ -74,6 +74,41 @@ def _validate(self, cuts: CutSet) -> None:
...
+        if self.collate:
+            audio = collate_multi_channel_audio_rmpad(cuts)
+            audio_lens = 0 # TODO
+            return {
+                "cuts": cuts,
+                "audio": audio,
+                "audio_lens": audio_lens,
+            }
+        else:
+            return {"cuts": cuts, "audio": [c.load_audio(mixed=False) for c in cuts]}
```
pzelasko, line 104: This line would again have the extra padding channels problem. This suggests that maybe the solution should not be (entirely) in the `collate` function, but inside `load_audio`, e.g. controlled by an extra argument?

lhotse/dataset/speech_recognition.py
```diff
@@ -96,7 +96,7 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List[str]]]
         Return a new batch, with the batch size automatically determined using the constraints
         of max_frames and max_cuts.
         """
-        validate_for_asr(cuts)
+        #validate_for_asr(cuts)
```
pzelasko: This should be uncommented

pzelasko (commented):
Thanks! I think it's a good start, will take a bit more effort to handle multi-channel properly here, but we'll get there.

pzelasko:
For anybody interested in this, here's some context of our earlier discussion with @kkarrancsu 

I expect it you to run into issues related to padding and MUSAN data augmentation with it. Basically, padding and augmentation creates extra tracks in MixedCut, and neither MixedCut nor collate_multi_channel_audio know which tracks are the data, and which tracks are the padding / noise. So, for 4-channel audio, you might end up with 6-channel output from `collate_multi_channel_audio`; you might want to modify it somehow to avoid that (interpret padding as padding all channels and noise augmentation as… adding the noise to each channel? or just one of them? I don’t know)

pzelasko:
@kkarrancsu I have a different idea -- we could add a new attribute to `MixTrack` that's called `separate_channel: bool = True` -- it would indicate if a given track is a "data" channel or an "augmentation" channel. We would add a parameter with the same name to `mix` (default=True) and `pad` (default=False) operations on all cuts. 

We would need to extend `FeatureMixer` and `AudioMixer` to handle these cases properly. The idea is to again add `separate_channel` parameter to `add_to_mix` method, and modify the  property `unmixed_audio`:

https://github.com/lhotse-speech/lhotse/blob/b41e4f8c57954b94790f2d05de310a11071512bc/lhotse/audio.py#L984-L990

so that instead of simply vstacking the right channels, it vstacks only the "separate" channels, downmixes the remaining channels to mono, and adds them to each of the "separate" channels. The analogous operation is needed for `FeatureMixer`.

Then, `collate_multi_channel_audio` could remain almost unmodified w.r.t. to what's there in the codebase, and it would not need any special logic to figure out what's a real channel and what is just padding/augmentation.

Of course we'd need to add more unit tests to make sure this doesn't break anything and works as expected.

danpovey:
It seems to me that the more "correct" way to do this would be, when adding noise to multi-channel audio, to add multiple channels of noise.  I assume this would require some nontrivial simulation, possibly with multiple sources.
I would have thought that for any application that was going to process multiple-channel audio in a non-trivial way, just adding a single type of noise to all channels would not really be sufficient.

danpovey:
OK, sure, it was just a thought.

PR #533, opened 2022-01-13
Skipping problematic audios - continued

pzelasko:
Hey @pkufool,

it took me a while to get back to this, but I finally found a way to fully support what you started building in #513. I set up a new PR based on your branch + some modifications from me. 

The core idea is to allow `read_audio_from_cuts`, `collate_audio`, `AudioSamples`, and `OnTheFlyFeatures` to return a reduced CutSet in addition to the collated tensors. Since this change is possibly breaking to other users, I added a flag `fault_tolerant` that controls whether this additional CutSet is returned or not. It's `False` by default, so existing users will not observe any change, but if you flip it to `True`, `K2SpeechRecognitionDataset` will do the right thing. It is a bit messy but I did not see a more elegant way to integrate that change.

LMK if this helps.

### DIFF
--- a/lhotse/dataset/input_strategies.py
+++ b/lhotse/dataset/input_strategies.py
@@ -1,7 +1,7 @@
 import logging
 from concurrent.futures import Executor, ThreadPoolExecutor
 from functools import lru_cache
-from typing import Callable, Dict, List, Optional, Tuple, Type, TypeVar
+from typing import Callable, Dict, List, Optional, Tuple, Type, TypeVar, Union
 
 import torch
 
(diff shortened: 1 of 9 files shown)


lhotse/dataset/input_strategies.py
```diff
@@ -180,7 +180,34 @@ class AudioSamples(BatchIO):
     .. automethod:: __call__
     """
 
-    def __call__(self, cuts: CutSet) -> Tuple[torch.Tensor, torch.IntTensor]:
+    def __init__(
```
danpovey, line 183: Above where it says "It pads the recordings if needed" (line 178), would be nice to say why that might be needed.

pzelasko: I extended the docs in that place and added some tests. Merging!

pkufool:
Thanks! I will continue to do the wenetspeech experiment with this fix.

PR #537, opened 2022-01-14
[kaldi] Fixed wav source for command type

desh2608:
The function expects to return a dictionary indexed by channel.


pzelasko:
Good catch, thanks! I think we could use some unit tests for kaldi import / export stuff..

PR #545, opened 2022-01-23
Minor updates for AMI and ICSI

desh2608:
This PR contains some minor updates for the AMI and ICSI recipes that were made during mini SCALE.
* New Kaldi-style text normalization option
* Handling some AMI sessions where headset-mics are stereo
* Minor bug fix for ICSI channel names


lhotse/recipes/ami.py
```diff
@@ -265,7 +265,7 @@ class AmiSegmentAnnotation(NamedTuple):
 
 
 def parse_ami_annotations(
-    annotations_dir: Pathlike, normalize_text: bool = True
+    annotations_dir: Pathlike, normalize: str = "upper"
```
pzelasko, line 268: why not make "kaldi" default here as well for consistency?

lhotse/recipes/ami.py
```diff
@@ -343,19 +341,44 @@ def _parse_href(href, wid_to_word):
...
+                            start_time=start_time,
+                            end_time=end_time,
+                        )
                     )
-                )
 
     return annotations
 
 
+def normalize_text(text: str, normalize: str = "upper") -> str:
```
pzelasko, line 359: ... and here

lhotse/recipes/icsi.py
```diff
@@ -126,9 +127,9 @@
 }
 
 MIC_TO_CHANNELS = {
-    "ihm": [1, 2, 3, 4, 5, 6, 8, 9], # we include 6 since it is used as back-off from some speakers for which no headset-mic exists
-    "sdm": [6],
-    "mdm": ["E", "F", 6, 7],
+    "ihm": ["1", "2", "3", "4", "5", "6", "8", "9"], # we include 6 since it is used as back-off from some speakers for which no headset-mic exists
```
pzelasko, line 130: what's the reason these are represented as strings now?

pzelasko (commented):
Thanks! I left a bunch of comments

lhotse/recipes/icsi.py
LasseWolter: If not converted to string this line won't work: 

`audio_paths = audio_dir.rglob(f"chan[{channels}].sph")`
https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/icsi.py#L505

LasseWolter:
@desh2608 because you worked with the ICSI recipe, you might know more about  #551

pzelasko:
LGTM

lhotse/recipes/ami.py
desh2608: Makes sense, I'll do it next time I raise a PR.

lhotse/recipes/icsi.py
desh2608: @LasseWolter thanks, that's exactly the reason. Sorry I was traveling last couple of days so couldn't reply to these comments.

Issue #559, opened 2022-02-02
Silent fail with certain Fbank Configs and num_jobs > 1


LasseWolter:
I'm working on the ICSI dataset. Thus, I'm using the latest/unreleased lhotse version. 

I tried to use certain configs with the Fbank extractor but some of them don't work when I'm tyring to use multiple jobs.
 

- See the CPU utilisation in the video below since the progress bar alone isn't as expressive for multiple jobs
- Note that once the problematic field is removed all CPUs go to 100% utilisation

I'm a bit confused about how this could be the case. Please check the code + video below to see my issue. 

---
**EDIT:** *not working* means that the Fbank extractor doesn't finish and doesn't utilise the CPU properly. It seems to just be stuck without progress - that's why I called it 'silent fail'
**with** `num_mel_bins=128`
  - using 1 job the `compute_and_store_features` function finishes in around 70s
  - using 8 jobs I can't see any progress after 5 minutes  
  
**without** `num_mel_bins=128`
   - using 1 job the `compute_and_store_features` function finishes in around 70s
  - using 8 jobs it finishes after around 24s 
---

*python version*:  Python 3.9.7
*python env used*: [bug-req.txt](https://github.com/lhotse-speech/lhotse/files/7982707/bug-req.txt)
*lhotse version*: latest version at commit dd2342abad571a9a06dfa25df38a823cbcd21d86
**Code I used in the video below**

```
from lhotse import CutSet, Fbank, FbankConfig
import sys

num_jobs = int(sys.argv[1])
f2 = Fbank(
    FbankConfig(
        frame_shift=0.02275,
        num_mel_bins=128
    )
)
cuts = CutSet.from_jsonl('cutset.jsonl')   # a cutset consisting of 86 cuts

cuts.compute_and_store_features(
    storage_path='test_feats',
    extractor=f2, 
    num_jobs=num_jobs
)
```

The same happened when I used `num_filters=128`.

https://user-images.githubusercontent.com/29123172/152073109-bd581f0f-87d2-4dba-bbee-050ee8b56698.mp4

pzelasko:
Interesting. What if you set `torch.set_num_threads(1)` and `torch.set_num_interop_threads(1)`? PyTorch uses multithreaded BLAS by default which doesn't play really well with multiprocessing.

danpovey:
For the rest of us, could someone explain what the issue is?  Just the CPU utilization is not as expected, or there's somethign wrong with the output?

[...]

PR #561, opened 2022-02-02
Add warning if num_jobs >1 and torch.num_threads >1

LasseWolter:
+ in some cases this can result in an endless computation
  This warning makes the user aware of this
  
addresses #559

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -4059,7 +4059,7 @@ class CutSet(Serializable, Sequence[Cut]):
         if num_jobs > 1 and torch.get_num_threads() > 1:
             logging.warning(
                 "num_jobs is > 1 and torch's number of threads is > 1 as well: "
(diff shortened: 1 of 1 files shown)


lhotse/cut.py
```diff
@@ -4055,6 +4056,13 @@ def compute_and_store_features(
             )
             executor = None
 
+        if num_jobs > 1 and torch.get_num_threads() > 1:
+            logging.warning(
+                "num_jobs is > 1 and torch's number of threads is > 1 as well: "
+                "For certain configs this can result in a never ending computation. "
+                "If this happens, use torch.set_num_threads(1) to circumvent this."
```
pzelasko, line 4063: Should we also mention `torch.set_num_interop_threads(1)`?

pzelasko:
Thanks!

PR #565, opened 2022-02-03
CutSet multiplexing

pzelasko:
CC @danpovey I think you might find it interesting for dataset combination, the code would look roughly like this:

```python
libri_cuts = load_manifest_lazy(...)
giga_cuts = load_manifest_lazy(...)
tedlium_cuts = load_manifest_lazy(...)

muxed_cuts = CutSet.mux(libri_cuts, giga_cuts, tedlium_cuts, weights=[...], seed=10)

sampler = DynamicBucketingSampler(muxed_cuts, shuffle=True, num_buckets=..., ...)
```

It completely circumvents the issue of mismatched duration buckets with ZipSampler, as all datasets participate in estimation of duration buckets in this approach (might want to increase `num_cuts_for_bucket_estimation` param in sampler for a good estimate).

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -4633,7 +4633,7 @@ class CutSet(Serializable, Sequence[Cut]):
         cls,
         *cut_sets: "CutSet",
         weights: Optional[List[Union[int, float]]] = None,
-        rng: Optional[random.Random] = None,
+        seed: int = 0,
     ) -> "CutSet":
         """
         Merges multiple CutSets into a new CutSet by lazily multiplexing them during iteration time.
@@ -4644,11 +4644,11 @@ class CutSet(Serializable, Sequence[Cut]):
             They can be either lazy or eager, but the resulting manifest will always be lazy.
         :param weights: an optional weight for each CutSet, affects the probability of it being sampled.
             The weights are uniform by default.
-        :param rng: an optional ``random.Random`` instance for reproducible iteration order.
+        :param seed: the random seed, ensures deterministic order across multiple iterations.
         """
         from lhotse.serialization import LazyIteratorMultiplexer
 
-        return cls(cuts=LazyIteratorMultiplexer(*cut_sets, weights=weights, rng=rng))
+        return cls(cuts=LazyIteratorMultiplexer(*cut_sets, weights=weights, seed=seed))
 
 
 def make_windowed_cuts_from_features(
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -4628,6 +4628,28 @@ class CutSet(Serializable, Sequence[Cut]):
         )
         return CutSet(cuts=merged)
 
+    @classmethod
+    def mux(
+        cls,
+        *cut_sets: "CutSet",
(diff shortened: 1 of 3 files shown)


lhotse/cut.py
```diff
@@ -4628,6 +4628,28 @@ def __add__(self, other: "CutSet") -> "CutSet":
...
+        seed: int = 0,
+    ) -> "CutSet":
+        """
+        Merges multiple CutSets into a new CutSet by lazily multiplexing them during iteration time.
+        If one of the CutSets is exhausted before the others, we will keep iterating until all CutSets
+        are exhausted.
+
+        :param cut_sets: cut sets to be multiplexed.
+            They can be either lazy or eager, but the resulting manifest will always be lazy.
+        :param weights: an optional weight for each CutSet, affects the probability of it being sampled.
```
janvainer, line 4645: @pzelasko why not set the weights proportional to cutset sizes by default? This way you would deplete all cutsets at the same time on average. If we keep it uniform, we are risking that a small cutset A is depleted fast and for the rest of the epoch, there will be only larger cutset B.

pzelasko: Leaving it up to the user in case the cut sets are very large and opened lazily (I can also imagine len not being available for some types of lazy manifests in the future)

lhotse/serialization.py
```diff
@@ -527,6 +528,75 @@ def __add__(self, other) -> "LazyIteratorChain":
...
+                    (i, w)
+                    for i, (is_exhausted, w) in enumerate(zip(exhausted, self.weights))
+                    if not is_exhausted
+                ]
+            )
+            idx = rng.choices(active_indexes, weights=active_weights, k=1)[0]
+            selected = iters[idx]
+            try:
+                item = next(selected)
+                yield item
```
csukuangfj, line 579: Is it possible to return also the `idx`?

We may need to select a different network to process the returned `item` depending on the returned `idx`.

For instance, when combining gigaspeech and librispeech in transducer training, the encoder network is shared and there are two separate decoder+joiner networks for each dataset. If the returned item is from gigaspeech, we would run the decoder+joiner for gigaspeech; if the returned item is from librispeech, we would run the decoder+joiner for librispeech.

If this function returns only `item` without `idx`, it is difficult to tell which dataset the returned `item` is sampled from. Therefore, it's also difficult to select which decoder+joiner to run.

danpovey:
Nice!
Yes, this might be a good way to go.  We can see which is better (might affect both speed and convergence).

lhotse/serialization.py
```diff
@@ -527,6 +528,75 @@ def __add__(self, other) -> "LazyIteratorChain":
...
+        else:
+            self.weights = weights
+
+        assert len(self.iterators) == len(self.weights)
+
+    def __iter__(self):
+        rng = random.Random(self.seed)
+        iters = [iter(it) for it in self.iterators]
+        exhausted = [False for _ in range(len(iters))]
+        while not all(exhausted):
```
csukuangfj, line 567: Also, can we support to specify that as soon as a specified cutset is exhausted, it breaks the while loop.

PR #566, opened 2022-02-04
Add method to create SupervisionSet from RTTM files

desh2608:
I'm not sure how useful this method would be for others (I have been using this for some of my own work). If you think it can be added, I will include some tests.


lhotse/supervision.py
```diff
@@ -512,6 +512,50 @@ def from_dicts(data: Iterable[Dict]) -> "SupervisionSet":
             SupervisionSegment.from_dict(s) for s in data
         )
 
+    @staticmethod
+    def from_rttm(path: Pathlike) -> "SupervisionSet":
+        """
+        Read an RTTM file located at ``path`` and create a :class:`.SupervisionSet` manifest for them.
+        Can be used to create supervisions from custom RTTM files (see, for example, :class:`lhotse.dataset.DiarizationDataset`).
+
+        Use the :meth:`.from_rttms` method if you want to read multiple RTTM files at once.
```
pzelasko, line 521: How about modifying this function to accept both a single Pathlike and an iterable of them (you can check with `isinstance`)?

lhotse/supervision.py
```diff
@@ -512,6 +512,50 @@ def from_dicts(data: Iterable[Dict]) -> "SupervisionSet":
             SupervisionSegment.from_dict(s) for s in data
         )
 
+    @staticmethod
+    def from_rttm(path: Pathlike) -> "SupervisionSet":
```
pzelasko, line 516: It would be nice to add a short description of RTTM format in this docstring (which column stands for what)

pzelasko (commented):
Looks good!

desh2608:
@pzelasko done. Also added a simple test.

desh2608:
Do not merge this yet --- I am looking for a way to make the fixture as a tmpfile from the code itself instead of adding it in the fixtures directory.

lhotse/supervision.py
```diff
@@ -512,6 +513,89 @@ def from_dicts(data: Iterable[Dict]) -> "SupervisionSet":
...
+                        channel=int(parts[2]),
+                        start=float(parts[3]),
+                        duration=float(parts[4]),
+                        speaker=parts[7],
+                    )
+                )
+        return SupervisionSet.from_segments(segments)
+
+    @staticmethod
+    def from_rttms(paths: Union[Pathlike, Iterable[Pathlike]]) -> "SupervisionSet":
```
pzelasko, line 574: I meant if we can merge `from_rttms` and `from_rttm` into a single method that auto-detects the right thing to do with a single path vs an iterable of them. The names differ by just one character which makes them easy to confuse. WDYT?

lhotse/supervision.py
```diff
@@ -512,6 +513,89 @@ def from_dicts(data: Iterable[Dict]) -> "SupervisionSet":
...
+
+        .. hint::
+
+            For creating a SupervisionSet from a single RTTM file, use the :meth:`.from_rttm` method.
+
+        :param paths: Path to RTTM files (can be path to a directory or an iterable of paths to RTTM files).
+        :return: a new ``SupervisionSet`` instance containing segments from the RTTM files.
+        """
+        if isinstance(paths, Pathlike.__args__):
+            paths = check_and_rglob(paths, "*.rttm")
```
pzelasko, line 592: I think this rglob pattern wouldn't find files named `rttm` which are somewhat common in Kaldi recipes?

lhotse/supervision.py
desh2608: Yeah that makes sense. I'll make that change.

lhotse/supervision.py
desh2608: I suppose in that case I will remove rglob and just assume that the user has provided the path to a file or an iterable of paths.

desh2608:
Maybe ready to merge? @pzelasko

pzelasko:
Looks great, thanks!

PR #570, opened 2022-02-06
Quick fix for an edge case in CutSet.mix

pzelasko:


danpovey:
Thanks, will test.

danpovey:
OK it's gone past the point it failed at last time, so probably good to merge, thanks!

PR #571, opened 2022-02-06
WIP: add cut extend method

desh2608:
This method can be used to extend a cut by adding context on one or both sides. I'm still working on the implementation for `MixedCut` and need some feedback about what exactly should be done for this case?

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -161,6 +161,7 @@ class Cut:
 
         >>> cut_2_to_4s = cut.truncate(offset=2, duration=2)
         >>> cut_padded = cut.pad(duration=10.0)
+        >>> cut_extended = cut.extend(duration=5.0, direction='both')
         >>> cut_mixed = cut.mix(other_cut, offset_other_by=5.0, snr=20)
         >>> cut_append = cut.append(other_cut)
         >>> cut_24k = cut.resample(24000)
@@ -226,6 +227,7 @@ class Cut:
     drop_supervisions: Callable
     truncate: Callable
     pad: Callable
+    extend: Callable
     resample: Callable
     perturb_speed: Callable
     perturb_tempo: Callable
@@ -1079,6 +1081,10 @@ class MonoCut(Cut):
         it's only during the call to :meth:`MonoCut.load_features` / :meth:`MonoCut.load_audio`
         when the actual changes happen (a subset of features/audio is loaded).
 
+        .. hint::
+
+            To extend a cut by a fixed duration, use the :meth:`MonoCut.extend` method.
+
         :param offset: float (seconds), controls the start of the new cut relative to the current MonoCut's start.
             E.g., if the current MonoCut starts at 10.0, and offset is 2.0, the new start is 12.0.
         :param duration: optional float (seconds), controls the duration of the resulting MonoCut.
@@ -1090,9 +1096,6 @@ class MonoCut(Cut):
             large number of supervisions. Intended as an internal parameter.
         :return: a new MonoCut instance. If the current MonoCut is shorter than the duration, return None.
         """
-        # Note: technically, truncate's code can be used for "expanding" the cut as well:
-        #       In that case, we must ensure that the start of MonoCut is not before the start
-        #       of the actual Recording, hence max(..., 0).
         assert (
             offset >= 0
         ), f"Offset for truncate must be non-negative (provided {offset})."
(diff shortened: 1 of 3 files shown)


lhotse/cut.py
```diff
@@ -1152,6 +1155,80 @@ def truncate(
...
+        """
+        Returns a new MonoCut that is an extended region of the current MonoCut.
+
+        Note that no operation is done on the actual features or recording -
+        it's only during the call to :meth:`MonoCut.load_features` / :meth:`MonoCut.load_audio`
+        when the actual changes happen (an extended version of features/audio is loaded).
+
+        .. hint::
+
+            This method extends a cut by a given duration, either to the left or to the right (or both).
```
pzelasko, line 1174: This could mentioned using the "real" contents of the recording that the cut is a part of.

lhotse/cut.py
```diff
@@ -1152,6 +1155,80 @@ def truncate(
...
+            To "expand" a cut by padding, use :meth:`MonoCut.pad`. To "truncate" a cut, use :meth:`MonoCut.truncate`.
+
+        :param duration: float (seconds), controls the duration of the resulting MonoCut.
```
pzelasko, line 1177: this is unclear whether the cut is extended **by** duration or extended **to** duration

lhotse/cut.py
```diff
@@ -1152,6 +1155,80 @@ def truncate(
...
+        :param direction: string, 'left', 'right' or 'both'. Determines whether to extend on the left,
+            right, or both sides. If 'both', extend on both sides by the duration specified in `duration`.
```
pzelasko, line 1179: I think this is inconsistent with other methods like `pad` which, when direction="both", would extend in each direction by `duration/2`

lhotse/cut.py
```diff
@@ -1152,6 +1155,80 @@ def truncate(
...
+        if self.has_features:
+            if new_start < self.features.start or new_end > self.features.end:
+                logging.warning(
+                    "Attempting to extend a MonoCut that exceeds the range of pre-computed features. "
+                    "The feature manifest will be detached."
+                )
+                extra_kwargs["features"] = None
+
+        if self.custom is not None:
+            extra_kwargs["custom"] = self.custom
```
pzelasko, line 1213:
```suggestion
            extra_kwargs["custom"] = self.custom.copy()
```
otherwise you will modify the existing cut instance

lhotse/cut.py
```diff
@@ -1152,6 +1155,80 @@ def truncate(
...
+        new_duration = round(new_duration, ndigits=8)
+
+        new_supervisions = (
+            segment.with_offset(-(self.start - new_start))
+            for segment in self.supervisions
+        )
+
+        extra_kwargs = {}
+        if self.has_features:
+            if new_start < self.features.start or new_end > self.features.end:
```
pzelasko, line 1205: I'm pretty sure this conditional is begging for edge-cases where we'll be off by a single frame and accidentally drop features (or custom fields). I'd suggest to add a tolerance of at least `frame_shift`.

pzelasko: (same for custom fields below)

lhotse/cut.py
```diff
@@ -1594,6 +1671,38 @@ def truncate(
...
+        self,
+        *,
+        duration: Seconds,
+        direction: str = "both",
+        preserve_id: bool = False,
+        **kwargs,
+    ) -> "PaddingCut":
+        new_duration = self.duration + duration
+        if direction == "both":
+            new_duration += duration
```
pzelasko, line 1685: same comment as above re `both` and duration

lhotse/cut.py
```diff
@@ -2184,6 +2293,45 @@ def truncate(
 
         return new_cut
 
+    def extend(
```
pzelasko, line 2296: I don't believe this method is well defined on `MixedCut`, TBH I'd just `raise ValueError("It is undefined to extend() a MixedCut.")`

test/cut/test_cut_extend.py
```diff
@@ -0,0 +1,86 @@
...
+        "extend_duration",
+        "extend_direction",
+        "expected",
+    ],
+    [
+        (0.3, 0.8, 0.0, 1.0, 0.1, "both", True),
+        (0.3, 0.8, 0.3, 0.8, 0.1, "both", False),
+    ],
+)
+def test_extend_cut_with_features(
```
pzelasko, line 65: we'd need another test for cuts with attached custom temporal arrays

pzelasko (commented):
Looks good to me, the only misleading thing is that `duration` behaves differently than in `pad` and `truncate` (and maybe other methods): there, it's the **total** duration of cut after the operation, whereas here it's the **added** duration. But if we document it well maybe it is not an issue...

desh2608:
> Looks good to me, the only misleading thing is that `duration` behaves differently than in `pad` and `truncate` (and maybe other methods): there, it's the **total** duration of cut after the operation, whereas here it's the **added** duration. But if we document it well maybe it is not an issue...

@pzelasko I agree, but the meaning of `duration` is different because of the intended use case. In case of `truncate` or `pad`, we want to truncate/pad the cut "to" some length, whereas in `extend`, we want to extend it "by" some length (at least that seemed more straightforward in my use case). I will make this distinction more explicit in the docstring.

pzelasko:
An elegant solution might be to rename the method to `extend_by`

lhotse/cut.py
desh2608: Should I change it to be consistent with the `pad` usage?

pzelasko: After the discussion I think it makes sense to leave it the way you did, just make sure it's explicit in the documentation

lhotse/cut.py
desh2608: Good point. Are there any utility functions already available to make such comparisons? (I don't want to duplicate code)

pzelasko: I think you can use `lhotse.utils.compute_num_frames`, and maybe accessing `num_frames` on either Cut or Features will come in handy? Not sure, but for this part we'll definitely need a good test coverage :)

desh2608:
@pzelasko I have made the suggested changes and added some tests.

pzelasko:
Thanks Desh, great work!! I'll let the tests pass after I committed a small change and then merge.

lhotse/cut.py
```diff
@@ -1152,6 +1159,121 @@ def truncate(
...
+        Note that no operation is done on the actual features or recording -
+        it's only during the call to :meth:`MonoCut.load_features` / :meth:`MonoCut.load_audio`
+        when the actual changes happen (an extended version of features/audio is loaded).
+
+        .. hint::
+
+            This method extends a cut by a given duration, either to the left or to the right (or both), using
+            the "real" content of the recording that the cut is part of. For example, a MonoCut spanning
+            the region from 2s to 5s in a recording, when extended by 2s to the right, will now span
+            the region from 2s to 7s in the same recording (provided the recording length exceeds 7s).
```
csukuangfj, line 1182: What if the length of the recording is shorter than 7s?

lhotse/cut.py
```diff
@@ -1152,6 +1159,121 @@ def truncate(
...
+
+        new_supervisions = (
+            segment.with_offset(self.start - new_start) for segment in self.supervisions
+        )
+
+        feature_kwargs = {}
+        if self.has_features:
+            # We compare in terms of frames, not seconds, to avoid rounding errors.
+            # We also allow a tolerance of 1 frame on either side.
+            new_start_frames = compute_num_frames(
```
csukuangfj, line 1220: Can this block of code be refactored and be reused by the following block for handling custom items?

I find that the only difference is changing `self.features` to `array` and other lines are basically the same.

lhotse/cut.py
```diff
@@ -1594,6 +1716,38 @@ def truncate(
...
         )
 
+    # noinspection PyUnusedLocal
+    def extend_by(
+        self,
+        *,
+        duration: Seconds,
+        direction: str = "both",
+        preserve_id: bool = False,
+        **kwargs,
```
csukuangfj, line 1726: This method lacks documentation. Also, it is not clear what `kwargs` can contain.

lhotse/cut.py
```diff
@@ -2184,6 +2338,24 @@ def truncate(
...
+        direction: str = "both",
+        preserve_id: bool = False,
+    ) -> "MixedCut":
+        """
+        This raises a ValueError since extending a MixedCut is not defined.
+
+        :param duration: float (seconds), duration (in seconds) to extend the MixedCut.
+        :param direction: string, 'left', 'right' or 'both'. Determines whether to extend on the left,
+            right, or both sides. If 'both', extend on both sides by the duration specified in `duration`.
+        :param preserve_id: bool. Should the truncated cut keep the same ID or get a new, random one.
```
csukuangfj, line 2354: > Should the truncated cut

The cut is not truncated but extended, right?

lhotse/cut.py
desh2608: It is only extended up to the duration of the recording. I will add this in the comment.

lhotse/cut.py
desh2608: Good point. I will refactor to avoid duplication.

lhotse/cut.py
desh2608: I mostly just copied this from the `truncate()` method for PaddingCut. But you're right --- I'll add some documentation, and remove kwargs since I don't think they're required.

desh2608:
I have made the changes suggested by @csukuangfj, and also added tests to cover PaddingCut and MixedCut.

pzelasko:
Wonderful, thanks!

PR #572, opened 2022-02-06
Account for duration param in extract_from_recording_and_store in lhotse/features/base.py

armusc:
Looks fine now as the asser from qa.py is no more activated

    assert expected_num_frames == f.num_frames, (
        f"Features: manifest is inconsistent: declared num_frames is {f.num_frames}, "
        f"but duration ({f.duration}s) / frame_shift ({f.frame_shift}s) results in {expected_num_frames} frames. "
        f"If you're using a custom feature extractor, you might need to ensure that it preserves "
        f"this relationship between duration, frame_shift and num_frames (use rounding up if needed - "
        f"see lhotse.utils.compute_num_frames)."
    )

### DIFF
--- a/lhotse/features/base.py
+++ b/lhotse/features/base.py
@@ -286,7 +286,7 @@ class FeatureExtractor(metaclass=ABCMeta):
             channels=channels if channels is not None else recording.channel_ids,
             # The start is relative to the beginning of the recording.
             start=offset,
-            duration=recording.duration,
+            duration=recording.duration if duration is not None else recording.duration,
             type=self.name,
             num_frames=feats.shape[0],
             num_features=feats.shape[1],


pzelasko:
Could you add a unit test that would fail before this change and works after your fix? I want to make sure we won’t introduce regressions for fixed issues like these.

armusc:
> Could you add a unit test that would fail before this change and works after your fix? I want to make sure we won’t introduce regressions for fixed issues like these.

well, I would have to understand how to do it first because, honestly, I have not done it before in github.
Just to let you know the context, I have changed several things so to accept importing from Kaldi with stereo channels; and, among other things, I have added a 
compute_and_store_features_from_recording in cut.py that would call 
extract_from_recording_and_store
rather than
extract_from_samples_and_store

the difference, if I have understood, is that extract_from_samples_and_store does not really allow me to extract the channel I want (it seems to only store this info on the Feature manifest), while extract_from_recording_and_store would
but within that function the input duration (in my case the duration of a cut that represents an utterance) would not be accounted for, but alwayd the duration of the input recording (the entire audio file, of which the utterance cut is only a segment)
that would prompt qa.py to fail that assert since the feature length would correspond to the recording length and not the utterance one
I did not want necessarily to share those changes because I know you are working on a different direction to support multi-channels, but it looked to me straightforward that extract_from_recording_and_store should use the duration param input if specified

pzelasko:
I see. Yeah we're intending to introduce `MultiCut` to properly support all kinds of multi-channel processing scenarios. 

Right now a `MonoCut` is intended to point to a specific channel, if you want to model multi-channel data you should use pairs (or tuples) of `MonoCut`s. Of course if your own changes work well for you, feel free to stick with them.

As for the test, you'd have to add a test method to one of the following test files:
- https://github.com/lhotse-speech/lhotse/blob/master/test/features/test_kaldi_features.py or
- https://github.com/lhotse-speech/lhotse/blob/master/test/cut/test_feature_extraction.py

I skimmed through the tests and I'm not sure that we have any test for the method you modified directly; instead it's indirectly tested via other methods. Since the changes didn't break anything, I'll just merge, and if you'd like to do an exercise in unit testing, feel free to submit another PR with the test.

Thanks for the contribution!

armusc:
ok, thanks for the heads up
I'm sure I'll just pull your next features on MultiCut; I only wanted to have rapid experiments on my datasets that are mostly multi-channels and with data already prepared for kaldi.
hope to contribute more

PR #582, opened 2022-02-14
WebDataset integration for optimized sequential I/O

pzelasko:
I'm exploring an option of integrating Lhotse with WebDataset to speed up the I/O. There are some scenarios when random access reads with HDF5/Chunky/any other format may not be not fast enough. 

The main idea is to allow the Cuts and other Python objects to keep the binary data inside of them, rather than refer to some storage paths. Of course, that can't be stored in JSON[L], but it can be stored in binary formats such as WebDataset (it basically uses tarballs, but also supports sharding and cloud storage). My initial idea (implemented here) is to make it fairly seamless to use (i.e. no API changes in Lhotse) by creating a WebDataset iterator inside of a lazy CutSet. I'll be experimenting with this a bit to understand better if this is the way to go or not. The current draft is runnable code which seems to work, but is not tested thoroughly and not benchmarked yet.


danpovey:
Nice!

pzelasko:
This is mostly ready, I got nice 50-100x speed-ups in I/O intensive workloads. 

The only cave-at is that multi-GPU DDP training requires using the `join` or `Join` context managers that supports uneven number of batches in multiple nodes. See [this](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel.join) and [this](https://pytorch.org/tutorials/advanced/generic_join.html#distributed-training-with-uneven-inputs-using-the-join-context-manager) for details and examples.

The usage looks like this:

Exporting data to WebDataset:
image


Loading WebDataset tarballs as lazy CutSet with in-memory data:
image


Training with Lhotse + WebDataset:
image

danpovey:
Cool!

pzelasko:
Merging, probably gonna follow up with minor changes here and there.

PR #583, opened 2022-02-14
Updated ICSI download structure/args for clarity

LasseWolter:
+ introduced target_dir arg which is the root for downloaded data
+ by default an 'audio' and 'transcripts' dir in this target
  dir will hold the corpus data
+ the user can pass arguments to adjust the name of audio and transcripts dir

I found it a bit confusing that the `audio_dir` was also used for the transcripts by default and intuitively didn't expect that passing a `transcripts_dir` means that the final transcripts will be in a subfolder of that folder, namely `<your passed dir>/transcripts`. 

Using one `target_dir` which contains an `audio` dir for audio and a `transcripts` dir for transcripts seemed more intuitive.

Let me know what you think.

### DIFF
--- a/lhotse/recipes/icsi.py
+++ b/lhotse/recipes/icsi.py
@@ -193,17 +193,18 @@ def download_icsi(
 ) -> None:
     """
     Download ICSI audio and annotations for provided microphone setting.
-    :param target_dir: Pathlike, the path in which audio and transcripts dir are created by default. 
+    :param target_dir: Pathlike, the path in which audio and transcripts dir are created by default.
     :param audio_dir: Pathlike (default = None), the path to store the audio data.
-    :param transcripts_dir: Pathlike (default = None), path to store the transcripts data 
+    :param transcripts_dir: Pathlike (default = None), path to store the transcripts data
     :param force_download: bool (default = False), if True, download even if file is present.
     :param url: str (default = 'http://groups.inf.ed.ac.uk/ami'), download URL.
     :param mic: str {'ihm','ihm-mix','sdm','mdm'}, type of mic setting.
     """
     target_dir = Path(target_dir)
-    audio_dir = Path(audio_dir) if audio_dir else target_dir / 'audio'
-    transcripts_dir = Path(
-        transcripts_dir) if transcripts_dir else target_dir / 'transcripts'
+    audio_dir = Path(audio_dir) if audio_dir else target_dir / "audio"
+    transcripts_dir = (
+        Path(transcripts_dir) if transcripts_dir else target_dir / "transcripts"
+    )
 
     # Audio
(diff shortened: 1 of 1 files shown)


LasseWolter:
Renaming the `transcripts` dir doesn't seem that clean to me. If you have a better way of doing this, let me know.

pzelasko:
@desh2608 could you review?

desh2608:
@LasseWolter Thanks, LGTM in general. Minor comment: could you replace "audio" with "speech" in the directory name (to be consistent with previous usage) and mention in the docstring for `audio_dir` and `transcript_dir` what the default values would be if they are not provided?

desh2608:
> Renaming the `transcripts` dir doesn't seem that clean to me. If you have a better way of doing this, let me know.

I think it's okay for now. I'm assuming several people who may use these data prep recipes would have the data downloaded in some specific structure anyway, so specifying where their `transcript_dir` is makes sense to me.

LasseWolter:
>  I'm assuming several people who may use these data prep recipes would have the data downloaded in some specific structure anyway, so specifying where their transcript_dir is makes sense to me.

I agree that it's a good feature to have. Was just wondering if: 
```
Path("transcripts").rename(transcripts_dir)
```

in line 233 is a good way of doing that.

LasseWolter:
> could you replace "audio" with "speech" in the directory name

Sure, will do that. Should I still keep the argument name as `audio_dir` or change it to `speech_dir` as well? 
Or should we keep `audio_dir` to be consistent with prior usage?

desh2608:
I think you can keep it as audio_dir since that has also been used in other recipes.

LasseWolter:
will commit those changes later today

lhotse/recipes/icsi.py
```diff
@@ -194,14 +194,14 @@ def download_icsi(
     """
     Download ICSI audio and annotations for provided microphone setting.
     :param target_dir: Pathlike, the path in which audio and transcripts dir are created by default.
-    :param audio_dir: Pathlike (default = None), the path to store the audio data.
-    :param transcripts_dir: Pathlike (default = None), path to store the transcripts data
+    :param audio_dir: Pathlike (default = '<target_dir>/audio'), the path to store the audio data.
```
pzelasko, line 197:
```suggestion
    :param audio_dir: Pathlike (default = '<target_dir>/speech'), the path to store the audio data.
```

is that right?

LasseWolter: Yes, it is. These are set at the very top of the download_icsi function. (line 204 and 205)
The reason I left the argument default `= None` is that these dirs are supposed to be relative to the target dir if they are not specified explicitly. 
Does that make sense?

pzelasko:
Thanks!

Issue #556, opened 2022-01-31
Memory Hungry


ngoel17:
scripts such as compute_fbank_musan.py consume 3.6GB of memory. 
This limits how many jobs can be run in parallel, despite the number of available CPUs. 
On another note - lhotse kaldi import $data_dir/train 8000 data/manifests/train
kind of commands also run a little slow.

pzelasko:
It's possible to leverage lazy cuts in Lhotse to reduce the memory overhead. You can use the following:
- creating cuts from recordings and supervisions: [`lhotse.cut.create_cut_set_lazy`](https://github.com/lhotse-speech/lhotse/blob/f1b66b8a8db2ea93e87dcb9db3991f6dd473b89d/lhotse/cut.py#L5063) (make sure to read the docs as it assumes specific sorting)
- extracting features with lazy cuts and iterative writing: [`CutSet.compute_and_store_features_batch`](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/cut.py#L4101)

As to Kaldi data dir imports, I don't think there's too much we could do -- it depends on the data dir size.

danpovey:
Nagenrda told me on a call that he was expecting an import command from a Kaldi data dir to take about a minute and it took something like half an hour.  Piotr, what would likely be the limiting factor in such a command?  Does it access the individual data files somehow, e.g. to validate that they exist or to check their length?  (I imagine getting the num-samples in a file might potentially take some time, especially if the wav.scp was based on a pipe).

pzelasko:
That’d be the most likely explanation. There’s a num jobs argument that can help speed this up https://github.com/lhotse-speech/lhotse/blob/f1b66b8a8db2ea93e87dcb9db3991f6dd473b89d/lhotse/kaldi.py#L60

pzelasko:
> BTW - I don't think it would help to increase the jobs for that scenario, because the speed will be disk-io limited.

If Kaldi's `utils/data/get_reco2dur.sh` or `steps/make_mfcc.sh` gets faster with more jobs then this would be as well; I've seen solid speedups even on fairly slow grids.

> ok. If it's verifying then that would make sense. But it does demand existence of "segments" files for creating supervision.

That's correct.

armusc:
> > BTW - I don't think it would help to increase the jobs for that scenario, because the speed will be disk-io limited.
> 
[...]

PR #584, opened 2022-02-14
Replace kaldiio with kaldi_native_io.

csukuangfj:
Need to test `lhotse/kaldi.py`

### DIFF
--- a/lhotse/features/io.py
+++ b/lhotse/features/io.py
@@ -951,26 +951,28 @@ Kaldi-compatible feature reader
 @register_reader
 class KaldiReader(FeaturesReader):
     """
-    Reads Kaldi's "feats.scp" file using kaldiio.
+    Reads Kaldi's "feats.scp" file using kaldi_native_io.
     ``storage_path`` corresponds to the path to ``feats.scp``.
     ``storage_key`` corresponds to the utterance-id in Kaldi.
 
     .. caution::
-        Requires ``kaldiio`` to be installed (``pip install kaldiio``).
+        Requires ``kaldi_native_io`` to be installed (``pip install kaldi_native_io``).
     """
 
-    name = "kaldiio"
+    name = "kaldi_native_io"
 
     def __init__(self, storage_path: Pathlike, *args, **kwargs):
-        if not is_module_available("kaldiio"):
+        if not is_module_available("kaldi_native_io"):
             raise ValueError(
-                "To read Kaldi feats.scp, please 'pip install kaldiio' first."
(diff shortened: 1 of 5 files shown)


csukuangfj:
From https://github.com/lhotse-speech/lhotse/issues/556
> On another note - lhotse kaldi import $data_dir/train 8000 data/manifests/train
kind of commands also run a little slow.

@ngoel17

Could you try this PR?

lhotse/features/io.py
```diff
@@ -951,26 +951,28 @@ def write(self, key: str, value: np.ndarray) -> str:
 @register_reader
 class KaldiReader(FeaturesReader):
```
pzelasko, line 952: To avoid breaking compatibility for users who used these readers/writers before, it'd be better to copy and modify those classes (`KaldiNativeReader`/`Writer`) instead, and make them default everywhere instead. It also makes sense to add a deprecation warning to `KaldiReader` and `KaldiWriter` pointing to the new implementation.

csukuangfj: How about keeping the name fixed, i.e., still using the name kaldiio? As we only change the implementation detail, it should be backward compatible.

pzelasko: makes sense!

csukuangfj: Reverted to `kaldiio` for reader/writer `type`.

csukuangfj:
I have tested it on the `yesno` recipe and everything seems to be working fine.
```
kaldi/egs/yesno/s5$ lhotse kaldi import -f 0.01 ./data/train_yesno 8000 $PWD/ttt

started

$ ls ttt
features.jsonl.gz  recordings.jsonl.gz
```
It is ready to merge now.

Maybe someone can use it to import a larger dataset and compare the importing time 
between `kaldiio` and `kaldi_native_io`.

lhotse/kaldi.py
```diff
@@ -42,7 +44,7 @@ def get_duration(
         import soundfile
 
         info = soundfile.info(path)
-    except:
+    except ImportError:
```
pzelasko, line 47: I don't think this is right -- the idea of this logic is to go from most-adequate/fastest backend -> most-flexible/slowest backend. With audio it seems you can never predict whether a given backend will be able to decode a given file type (missing/non-typical headers, format options in WAV/FLAC/SPHERE/OPUS/etc.), so if they fail on anything, we just try to go to the next one. 

Can you revert this change?

csukuangfj: Reverted.

pzelasko:
Thanks!

Issue #588, opened 2022-02-16
Bug in `CutSet.sample()`


desh2608:
Currently, the random ids are generated by calling `randint()` for `n_cuts` times, and then these ids are selected from the cuts. There are 2 possible issues here:

1. When `n_cuts` is larger than `len(cuts)`, this will definitely produce repeated cut ids, which leads to an `AssertionError` in `index_by_id_and_check` at the time of `CutSet` creation.
2. Even when `n_cuts` is smaller than `len(cuts)` but sufficiently large, repeated ids are very likely to be produced.

I think this line 
https://github.com/lhotse-speech/lhotse/blob/1e06fffec2ed5a023c524347a2d5486a20e27df5/lhotse/cut.py#L3939
should be changed to
```python
cut_indices = random.sample(range(len(self)), min(n_cuts, len(self)))
```
to fix this.

If you agree, I will make a PR with this change. (I am not sure how this would affect the samplers etc.)

pzelasko:
Thanks, I think you're right. It won't affect the samplers, they do not use this method, instead they iterate over the whole cutset.

PR #589, opened 2022-02-16
Fix CutSet sample, and add mix padding

desh2608:
* Fix for https://github.com/lhotse-speech/lhotse/issues/588
* Allow `offset` to be larger than reference cut duration by padding the reference cut

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -266,12 +266,18 @@ class Cut:
         self,
         other: "Cut",
         offset_other_by: Seconds = 0.0,
+        allow_padding: bool = False,
         snr: Optional[Decibels] = None,
         preserve_id: Optional[str] = None,
     ) -> "MixedCut":
         """Refer to :function:`~lhotse.cut.mix` documentation."""
         return mix(
-            self, other, offset=offset_other_by, snr=snr, preserve_id=preserve_id
+            self,
+            other,
+            offset=offset_other_by,
+            allow_padding=allow_padding,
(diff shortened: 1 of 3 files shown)


pzelasko:
Thanks, looks good to me! When it comes to mixing cuts, I am always a bit in paranoid mode because it gave me a lot of edge case headaches in the past. Would you mind modifying one of my randomized tests to check that this feature really doesn't break anything? See an example here:

https://github.com/lhotse-speech/lhotse/blob/master/test/cut/test_invariants_randomized.py#L23

It uses a library called "hypothesis" that generates random test cases given a set of constraints. I think you'd need to specifically modify this line:

https://github.com/lhotse-speech/lhotse/blob/master/test/cut/test_invariants_randomized.py#L47

to allow for larger offset, so that it tests your new feature. I suggest running it locally with a larger `MAX_EXAMPLES` setting (like 5000 -- generates many more edge cases) and if it's OK, leave it at 500 once you commit (otherwise it will take very long to run the tests in the CI).

desh2608:
@pzelasko The test `TestMixedCutNumFramesNumSamplesRandomized::test_invariants_mix` passed successfully. I got some weird error in `test_invariants_append` with 
```
formats: can't open input file `/tmp/tmp9xw8tbal.wav': Too many open files in system
```
but I don't think that is related to the mixing?

pzelasko:
Probably too many tests were being run before the file handles were cleaned up. Not an issue. Thanks!

PR #595, opened 2022-02-22
WIP: fix fault tolerant load audio

oplatek:
raise NonPositiveEnergyError so MuxCut works with fault-tolerant audio loading

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1121,9 +1121,11 @@ class AudioMixer:
             self.reference_energy = audio_energy(base_audio)
         else:
             self.reference_energy = reference_energy
-        assert (
-            self.reference_energy > 0.0
-        ), f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy})"
+
+        if self.reference_energy <= 0.0:
+            raise NonPositiveEnergyError(
+                f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy})"
+            )
 
     @property
     def unmixed_audio(self) -> np.ndarray:
@@ -1224,7 +1226,7 @@ class AudioMixer:
 
 
 def audio_energy(audio: np.ndarray) -> float:
-    return float(np.average(audio ** 2))
+    return float(np.average(audio**2))
 
 
 FileObject = Any  # Alias for file-like objects
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -2774,11 +2774,18 @@ class MixedCut(Cut):
             reference_audio = reference_cut.load_audio()
             reference_energy = audio_energy(reference_audio)
 
-        mixer = AudioMixer(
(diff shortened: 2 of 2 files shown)


lhotse/audio.py
```diff
@@ -1121,9 +1121,11 @@ def __init__(
             self.reference_energy = audio_energy(base_audio)
         else:
             self.reference_energy = reference_energy
-        assert (
-            self.reference_energy > 0.0
-        ), f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy})"
+
+        if self.reference_energy <= 0.0:
+            raise NonPositiveEnergyError(
+                f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy})"
```
pzelasko, line 1127: Can you use the opportunity to display the cut ID?

pzelasko:
Please apply black and LGTM

lhotse/audio.py
oplatek: No, the caller needs to do it. The tracks do not contain any cut info.

lhotse/cut.py
```diff
@@ -2792,7 +2799,7 @@ def load_audio(self, mixed: bool = True) -> Optional[np.ndarray]:
                 )
             except NonPositiveEnergyError as e:
                 logging.warning(
-                    str(e) + f' MonoCut with id "{track.cut.id}" will not be mixed in.'
+                    f'{e} MonoCut with id "{track.cut.id}" will not be mixed in.'
```
oplatek, line 2802: @pzelasko I believe the try-except block is not needed here since the energy is checked in the constructor, or do I miss something?

lhotse/cut.py
```diff
@@ -2774,11 +2774,18 @@ def load_audio(self, mixed: bool = True) -> Optional[np.ndarray]:
...
-        )
+        try:
+            mixer = AudioMixer(
+                self.tracks[0].cut.load_audio(),
+                sampling_rate=self.tracks[0].cut.sampling_rate,
+                reference_energy=reference_energy,
+            )
+        except NonPositiveEnergyError as e:
+            logging.warning(
+                f'{e} Cannot mix audios to the reference audio "{self.tracks[0].cut.id}"'
```
pzelasko, line 2785: I see, so in this case you'll return all zeroes. I guess it's as good solution as any -- can you add `"Returning zero signal."` at the end of the warning message?

ngoel17:
Doesn't it make more sense to either crash like the current code does, or to remove that cut from training completely?

pzelasko:
Oh yeah you can try to do `OnTheFlyFeatures(..., fault_tolerant=True)` and LMK if that helped.

pzelasko:
I finished this work in PR #615, it will either crash or skip the cut depending on how `fault_tolerant` arg is set in `OnTheFlyFeatures`/`AudioSamples`. 

Thanks @oplatek!

PR #598, opened 2022-02-23
Modified SpecAugment

luomingshuang:
A suggestion from @danpovey , some entails and experiments results can be found according to https://github.com/k2-fsa/icefall/issues/201 and https://github.com/k2-fsa/icefall/pull/205. 

The results shows that the modified specaugment can accelerate the convergence of the model comparing to the original.


lhotse/dataset/signal_transforms.py
```diff
@@ -124,12 +125,12 @@ class SpecAugment(torch.nn.Module):
...
-        frames_mask_size: int = 70,
-        max_frames_mask_fraction: float = 0.2,
-        p=0.5,
+        num_feature_masks: int = 10,
+        features_mask_size: int = 27,
+        num_frame_masks: int = 2,
+        frames_mask_size: int = 100,
+        max_frames_mask_fraction: float = 0.15,
+        p=0.9,
     ):
```
danpovey, line 134: Incidentally, if we ever end up using a large number of feature masks (e.g. 10), which might happen for long utterances, the SpecAugment code is not efficient as it would copy the data many times.  To fix this problem, it would be necessary to refactor the code, probably so that it would first accumulate a boolean mask vector in the loop, and later apply it.

pzelasko:
It seems the unit tests fail when the number of masks is set to zero, see here for the first failing test https://github.com/lhotse-speech/lhotse/runs/5299393771?check_suite_focus=true#step:7:154

luomingshuang:
Oh, there is a bug. When testing with python 3.6, 3.8, 3.9, it is successful. But it failed with python 3.7.

csukuangfj:
> So how should I solve it? \-_-/

1. Read the error log
2. If you do step 1, you will find that it is a connection timeout error
3. Re-run the tests

pzelasko:
LGTM! At some point we might want to refactor it the way Dan suggested (and restore support for 0 time masks). BTW if we didn't need the time warp factor, probably the whole implementation could be further simplified.

danpovey:
Is there  anyway we can figure out from the Lhotse version number or something similar, whether this patch has been applied?
So we know what value to use for the num_frame_masks in the recipe (since the default of 1 is likely not optimal for previous code, we can't just remove that option).

danpovey:
.. alternatively, this https://stackoverflow.com/questions/12627118/get-a-function-arguments-default-value
may be a way to solve it.

pzelasko:
Maybe a cleaner solution would be introducing a flag that enables/disables the new behavior, or have the new implementation as a separate class?

pzelasko:
You could assume this change is for Lhotse 1.0 but the issue is the dev branch has been 1.0 for a few months now so it might not be that helpful..

danpovey:
Hm.  You're right, it would have been cleaner to have a separate flag.  But I can't think  of a great way to do this, maybe it's too late at night.  I'd like it to have "good" defaults that will work well for the future.

danpovey:
@pzelasko is there any chance we could just bump the version of Lhotse now, to make it easier to test the version?
I know it's ugly, but it's less work and avoids certain other problems.

pzelasko:
Sorry for the late reply -- I think releasing Lhotse right now is even uglier, I think that I made a mistake in setting the next version to 1.0 and it's accumulating a lot of changes, and the high-level docs are fairly outdated right now. I intend to find a bit of time this month to at least refresh the docs and release 1.0 then.

pzelasko:
... I could technically change the release version to 0.13 but people have been using 1.0.dev for some time now, so that would be likely even more confusing.

PR #600, opened 2022-02-23
SPGISpeech recipe

desh2608:
This PR adds data preparation for [SPGISpeech](https://datasets.kensho.com/datasets/spgispeech).
Additionally, functionality is also added for specifying a callable as the `recording_id` parameter in `Recording.from_file()`.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -297,7 +297,7 @@ class Recording:
     @staticmethod
     def from_file(
         path: Pathlike,
-        recording_id: Optional[Union[str, Callable[[Pathlike], str]]] = None,
+        recording_id: Optional[Union[str, Callable[[Path], str]]] = None,
         relative_path_depth: Optional[int] = None,
         force_opus_sampling_rate: Optional[int] = None,
         force_read_audio: bool = False,
@@ -712,14 +712,11 @@ class RecordingSet(Serializable, Sequence[Recording]):
         """
         msg = f"Scanning audio files ({pattern})"
 
-        global file_read_worker
-
-        def file_read_worker(p: Path):
-            return Recording.from_file(
-                p,
-                force_opus_sampling_rate=force_opus_sampling_rate,
-                recording_id=recording_id,
-            )
+        file_read_worker = partial(
+            Recording.from_file,
+            force_opus_sampling_rate=force_opus_sampling_rate,
+            recording_id=recording_id,
+        )
 
         if num_jobs == 1:
             # Avoid spawning process for one job.
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -297,7 +297,7 @@ class Recording:
     @staticmethod
     def from_file(
         path: Pathlike,
-        recording_id: Optional[str] = None,
+        recording_id: Optional[Union[str, Callable[[Pathlike], str]]] = None,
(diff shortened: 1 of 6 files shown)


lhotse/audio.py
```diff
@@ -315,7 +315,7 @@ class Recording:
     @staticmethod
     def from_file(
         path: Pathlike,
-        recording_id: Optional[str] = None,
+        recording_id: Optional[Union[str, Callable[[Pathlike], str]]] = None,
```
pzelasko, line 318:
```suggestion
        recording_id: Optional[Union[str, Callable[[Path], str]]] = None,
```

looking at the code, it will always be of type `Path`

lhotse/audio.py
```diff
@@ -819,22 +828,32 @@ def from_dir(
...
+            specified, the filename will be used as the recording ID.
         :return: a new ``Recording`` instance pointing to the audio file.
         """
         msg = f"Scanning audio files ({pattern})"
-        fn = Recording.from_file
-        if force_opus_sampling_rate is not None:
-            fn = partial(
-                Recording.from_file, force_opus_sampling_rate=force_opus_sampling_rate
+
+        global file_read_worker
```
pzelasko, line 837: How about replacing this with:

```python
from functools import partial

file_read_worker = partial(Recording.from_file, force_opus_sampling_rate=force_opus_sampling_rate, recording_id=recording_id)
```

lhotse/recipes/spgispeech.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    logging.info(
+        "SPGISpeech is not available for direct download. Please fill out the form at"
+        " https://datasets.kensho.com/datasets/spgispeech to download the corpus."
+    )
+
+
+def normalize(text: str) -> str:
+    # Remove all punctuation
+    text = text.translate(str.maketrans("", "", string.punctuation))
```
pzelasko, line 57: are you sure that you also want to remove apostrophes and dashes?

pzelasko (commented):
Thanks! Nice use of lazy manifests, +1 for recording_id callable. Left a bunch of comments.

lhotse/audio.py
desh2608: Good idea, I'll make that change

lhotse/recipes/spgispeech.py
desh2608: I followed the ESPNet recipe for [text normalization](https://github.com/espnet/espnet/blob/9c24b3adddbde3402530080cb58ae08a6f4dd642/egs2/spgispeech/asr1/local/data_prep.sh#L30), where they do:
```
awk '{for(x=2;x<=NF;x++){gsub(/[[:punct:]]/,"",$x)}}1'
```
I believe this removes everything including apostrophes and dashes.

pzelasko: weird but fair enough

pzelasko: can you at least add a comment about it?

desh2608: sure

desh2608:
@pzelasko I have made suggested changes.

pzelasko (approved):
Thanks!

PR #601, opened 2022-02-24
[WIP][POC] make kaldi export support the mutifile sources

jtrmal:
@desh2608 @pzelasko I was thinking about implementing support for exporting multiple files sources RecordingSets (ami recipe is using those) 
Using this code addition I can both successfully export and the exported dir passes kaldi validation. What do you think?
y.

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -169,6 +169,40 @@ def load_kaldi_data_dir(
     return recording_set, supervision_set, feature_set
 
 
+def create_fake_single_channel_sets(
+    recordings: RecordingSet, supervisions: SupervisionSet
+):
+    import copy
+
+    new_recordings_dict = {}
+    for r in recordings:
+        for s in r.sources:
+            assert len(s.channels) == 1, (
+                "This code supports only single channel files"
+                "in the multiple audio sources Recordings"
+            )
+            r_new = copy.copy(r)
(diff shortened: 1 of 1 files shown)


pzelasko:
It sounds interesting to me, but I have one question -- will the combination of (kaldi export -> kaldi import) be an identity operation? I am not 100% sure but I think it mostly is right now, and I think I know of some projects that are depending on it. I think you're changing some recording IDs here which could break this behavior.

As a side note, unfortunately Kaldi import/export was more of an afterthought rather than a key feature for me, so I haven't set key assumptions and written any unit tests when I originally added it. I would be great if we could figure out a more principled approach and test it properly to avoid surprising people when we make changes in Kaldi related code.

jtrmal:
No Its not identity, I don't think that could work without storing some
extra meta data with kaldi dir.

Yes, I'm changing the IDs, but I'm changing it only in case when it wasn't
supported before, so there should not be a collision with expectations.


Y.

jtrmal:
I mean it won't be identity in the previously unsupported case.  The
previously supported cases are staying as they are.
Y.

pzelasko:
cool let's get it merged then :)

pzelasko:
storing extra metadata in Kaldi dir makes sense to me, good idea!

jtrmal:
I will reimplement this for the new kaldi.py structure

PR #602, opened 2022-02-24
Add `WebdatasetWriter` for iterative cut writing

pzelasko:


lhotse/dataset/webdataset.py
```diff
@@ -126,6 +117,129 @@ def export_to_webdataset(
...
+    Export cuts with audio, features, and all custom data to a tarball shards with 500
+    cuts each::
+
+        >>> cuts = CutSet.from_jsonl_lazy("data/cuts-train.jsonl")
+        >>> with WebdatasetWriter("data/tars/shard-%06d.tar", shard_size=500) as writer:
+        ...     for cut in cuts:
+        ...         writer.write(cut)
+        >>> output_paths = writer.output_manifest_paths()
+
+    See also: :func`.export_to_webdataset`
```
danpovey, line 152: For this and the other web-dataset writing functions, it would be nice to have an indication to how one is supposed to read this data.  Since only the tar archive is output, it's not clear to me how one would grab its metadata.

pzelasko:
The tar contains both data and metadata. I’ll write some sort of tutorial once I’m back from holidays.

PR #604, opened 2022-03-01
optimization for specaugment

luomingshuang:
How about this change for specaugment? This change is for reducing the times of coping the whole features. Also, I notice the mask_along_axis in https://pytorch.org/audio/stable/functional.html#torchaudio.functional.mask_along_axis happens a change.

### DIFF
--- a/lhotse/dataset/signal_transforms.py
+++ b/lhotse/dataset/signal_transforms.py
@@ -311,9 +311,15 @@ def mask_along_axis_optimized(
     mask_ends = (min_values.long() + values.long()).squeeze()
 
     if axis == 1:
+        if mask_times == 1:
+            features[:, mask_starts:mask_ends] = mask_value
+            return features.squeeze(0)
         for (mask_start, mask_end) in zip(mask_starts, mask_ends):
             features[:, mask_start:mask_end] = mask_value
     else:
+        if mask_times == 1:
+            features[:, :, mask_starts:mask_ends] = mask_value
+            return features.squeeze(0)
         for (mask_start, mask_end) in zip(mask_starts, mask_ends):
             features[:, :, mask_start:mask_end] = mask_value
 
--- a/lhotse/dataset/signal_transforms.py
+++ b/lhotse/dataset/signal_transforms.py
@@ -259,7 +259,7 @@ class SpecAugment(torch.nn.Module):
             for (mask_start, mask_end) in zip(mask_starts, mask_ends):
                 features[:, :, mask_start:mask_end] = mean
             features = features.squeeze(0)
-            
+
         return features
 
     def state_dict(self) -> Dict:
--- a/lhotse/dataset/signal_transforms.py
+++ b/lhotse/dataset/signal_transforms.py
@@ -243,13 +243,23 @@ class SpecAugment(torch.nn.Module):
                 self.frames_mask_size, _max_tot_mask_frames // num_frame_masks
             )
 
-            for _ in range(num_frame_masks):
-                features = mask_along_axis(
-                    features.unsqueeze(0),
-                    mask_param=max_mask_frames,
-                    mask_value=mean,
-                    axis=1,
-                ).squeeze(0)
+            features = features.unsqueeze(0)
+            features = features.reshape([-1] + list(features.size()[-2:]))
+            values = [torch.rand(1) * max_mask_frames for _ in range(num_frame_masks)]
+            min_values = [
+                torch.rand(1) * (features.size(1) - value) for value in values
+            ]
+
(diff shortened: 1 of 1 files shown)


lhotse/dataset/signal_transforms.py
```diff
@@ -243,13 +243,23 @@ def _forward_single(
...
             )
 
-            for _ in range(num_frame_masks):
-                features = mask_along_axis(
-                    features.unsqueeze(0),
-                    mask_param=max_mask_frames,
-                    mask_value=mean,
-                    axis=1,
-                ).squeeze(0)
+            features = features.unsqueeze(0)
```
danpovey, line 246: This code is too hard to follow, I think you need a bit more structure, like a well-documented function.
Also, I think torch.randint might possibly be useful.

luomingshuang: OK, I will warp these into a function.

luomingshuang:
BTW, I consider there are two methods for it:
the first is:
```
features = features.unsqueeze(0)
features = features.reshape([-1] + list(features.size()[-2:]))
            
values = torch.randint(0,max_mask_frames,(1,num_frame_masks))
min_values = torch.rand(1,num_frame_masks)*(features.size(1) - values)
mask_starts = (min_values.long()).squeeze()
mask_ends = (min_values.long() + values.long()).squeeze()
            
for (mask_start, mask_end) in zip(mask_starts, mask_ends):
      features[:, :, mask_start:mask_end] = mean
features = features.squeeze(0)
```
the second is:
```
features = features.unsqueeze(0)
features = features.reshape([-1] + list(features.size()[-2:]))
            
values = torch.randint(0,max_mask_frames,(1,num_frame_masks))
min_values = torch.rand(1,num_frame_masks)*(features.size(1) - values)
mask_starts = (min_values.long()).squeeze()
mask_ends = (min_values.long() + values.long()).squeeze()
            
mask = torch.arange(0, features.shape[1], device=features.device, dtype=features.dtype)
mask_bool = (mask>=mask_starts[0])&(mask<mask_ends[0])
for (mask_start, mask_end) in zip(mask_starts, mask_ends):
      mask_bool ^= (mask>=mask_start)&(mask<mask_end)
features = features.masked_fill(mask.unsqueeze(-1), mean)

features = features.squeeze(0)
```
For testing their running time, I make them have the same `values` and `min_values`, there are some results:
```
duration for the original method:  0.0012099742889404297
duration for the first method:  0.0001354217529296875
duration for the second method:  0.00043702125549316406

duration for the original method:  0.0014553070068359375
duration for the first method:  0.0001671314239501953
duration for the second method:  0.0005228519439697266

duration for the original method:  0.0012574195861816406
duration for the first method:  0.00013828277587890625
duration for the second method:  0.00044798851013183594

duration for the original method:  0.0013759136199951172
duration for the first method:  0.0001506805419921875
duration for the second method:  0.0004825592041015625

duration for the original method:  0.0013842582702636719
duration for the first method:  0.00014662742614746094
duration for the second method:  0.00044226646423339844

duration for the original method:  0.0013074874877929688
duration for the first method:  0.00015616416931152344
duration for the second method:  0.0004379749298095703

duration for the original method:  0.001360177993774414
duration for the first method:  0.0001709461212158203
duration for the second method:  0.0005428791046142578

duration for the original method:  0.0012598037719726562
duration for the first method:  0.00014519691467285156
duration for the second method:  0.0004558563232421875
```
According to the above results, we can see that the first method use the least running time. So I think the first one is ok for us. (I will warp it into a function.) Of course, if there are any other methods, it will be wonderful.

luomingshuang:
I wrap it into a function called `mask_along_axis_optimized` and also apply it for frequency masking.

pzelasko:
Thanks, that is a nice speed-up. I assume you tested that the output spectrograms look correct? It'd be great if you could apply the transform to a mini-batch from Icefall training and post the screenshot of `lhotse.dataset.vis.plot_batch(batch)` (used like this):

```python
dloader = DataLoader(...)
batch = next(iter(dloader))
lhotse.dataset.vis.plot_batch(batch)
```

pzelasko:
Thanks, that is a nice speed-up. I assume you tested that the output spectrograms look correct? It'd be great if you could apply the transform to a mini-batch from Icefall training and post the screenshot of `lhotse.dataset.vis.plot_batch(batch)` (used like this):

```python
dloader = DataLoader(...)
batch = next(iter(dloader))
lhotse.dataset.vis.plot_batch(batch)
```

luomingshuang:
OK. I will try to do it.

luomingshuang:
I test a mini-batch and it's size is (8, 708, 80). The spectrograms shows as follows:
image

pzelasko:
Thanks!

PR #606, opened 2022-03-02
Output more information from CutSet.describe.

csukuangfj:
The extra output is helpful to select the threshold 
for removing long utterances without removing too much data.


pzelasko:
Thanks!

PR #611, opened 2022-03-09
WIP: Vcc2018

oplatek:


lhotse/bin/modes/recipes/__init__.py
```diff
@@ -56,3 +56,4 @@
 from .voxceleb import *
 from .wenet_speech import *
 from .yesno import *
+from .vcc2018 import *
```
desh2608, line 59: We try to keep the imports sorted in lexicographical order. Could you put this import above `vctk`?

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+
+    converted_speech_dir = corpus_dir / "mnt/sysope/test_files/testVCC2/"
+    assert converted_speech_dir.is_dir(), f"No such directory: {converted_speech_dir}"
+
+    mos_dir = corpus_dir / "vcc2018_listening_test_scores"
+    assert mos_dir.is_dir(), f"No such directory: {mos_dir}"
+    mos_scores = mos_dir / "vcc2018_evaluation_mos.txt"
+    assert mos_scores.is_file()
+    sim_scores = mos_dir / "vcc2018_evaluation_sim.txt"
+    sim_scores.is_file()
```
desh2608, line 170: should this be an `assert`?

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+    assert trn_dir.is_dir(), f"No such directory: {trn_dir}"
+
+    reference_speech_dir = corpus_dir / "vcc2018_reference"
+    assert reference_speech_dir.is_dir(), f"No such directory: {reference_speech_dir}"
+
+    logging.info(
+        f"Collecting reference target recordings for the VCC2018 challange from {reference_speech_dir}"
+    )
+
+    # TODO
```
desh2608, line 182: Can be removed?

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+    if output_dir is not None:
+        output_dir = Path(output_dir)
+        output_dir.mkdir(parents=True, exist_ok=True)
+        recordings.to_jsonl(output_dir / "recordings.jsonl.gz")
+        supervisions.to_jsonl(output_dir / "supervisions.jsonl.gz")
+
+    return {"recordings": recordings, "supervisions": supervisions}
+
+
+def prepare_mos_supervisions(
```
desh2608, line 225: If this function is only intended to be used inside this module, it is recommended to add a single underscore, i.e., `_prepare_mos_supervisions()`. Note that this does not _enforce_ privacy but only indicates that this method should not be called directly.

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+                "system": row.system1_id,
+                "prompt": prompt_id,
+                "reference_path": reference_path,
+            },
+        )
+        supervisions.append(s)
+    return SupervisionSet.from_segments(supervisions)
+
+
+def load_vcc_results(path: Pathlike):
```
desh2608, line 283: Same `_load_vcc_results()`

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+    """
+    Returns pandas.DataFrame
+    """
+    #     """
```
desh2608, line 287: Use either multi-line comment (""" """) or single-line (#)

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+            "src_spk",
+            "tgt_spk",
+            "MOS",
+            "left_audio",
+        ]
+    ]
+    recording_ids = set(mos["left_audio"].tolist())
+    supervisions = []
+    for recording_id_wav in tqdm(recording_ids, desc="Supervision creation"):
+        recording_id = recording_id_wav.rstrip(".wav")
```
desh2608, line 247: If you make the `recording_ids` above as a set of `Path` types, then you could use `.stem` here instead.

lhotse/recipes/vcc2018.py
```diff
@@ -0,0 +1,329 @@
...
+        recordings.to_jsonl(output_dir / "recordings.jsonl.gz")
+        supervisions.to_jsonl(output_dir / "supervisions.jsonl.gz")
+
+    return {"recordings": recordings, "supervisions": supervisions}
+
+
+def prepare_mos_supervisions(
+    mos_results_path, recordings: RecordingSet, id2trn: Dict[str, str]
+) -> SupervisionSet:
+    # TODO very slow -> make it faster it takes ~8min 170it/s
```
desh2608, line 228: This may be because you are manipulating a Pandas dataframe. Pandas does a lot of book-keeping under the hood which makes it good for complicated operations, but here I think it may be better to just use something like a list of namedtuples, and then use `groupby` to group them.

desh2608 (changes requested):
Sorry, we missed reviewing this earlier. I have made some comments.

oplatek:
@desh2608 I forgot about this WIP PR. I will address your comments next week.

lhotse/recipes/vcc2018.py
oplatek: I know… but it is less readible with `Path(recording_id_wav).stem` since I literally need to use both recording_id wit hand without “.wav” for the original data and Lhotse use.

desh2608: `recording_id_wav.name` would give you the file name with the extension.

PR #617, opened 2022-03-14
Refresh tutorials in examples + quality of life improvements in code

pzelasko:


pzelasko:
I removed most of the notebooks which we had since they were 90% duplicated. The `examples` directory now has notebooks that are topic-focused. There are two as of this PR: one shows a "minimal, full, fairly realistic example" of k2 ASR DataLoader with mini LibriSpeech, the other one shows how to use some APIs of Cut and CutSet. We'll be adding more tutorials going forward.

CC @desh2608

PR #621, opened 2022-03-14
Options for handling multi-channel RIR

desh2608:
`rir_channels` option can be specified with `reverb_rir` or `ReverbWithImpulseResponse` to select which channels to use in case of multi-channel RIR. If base cut is MonoCut and RIR is multi-channel, it produces a MixedCut with number of tracks equal to the number of RIR channels.


lhotse/cut.py
```diff
@@ -1976,6 +2015,7 @@ def reverb_rir(
         normalize_output: bool = True,
         early_only: bool = False,
         affix_id: bool = True,
+        keep_mono: bool = False,
```
pzelasko, line 2018: Can we make it `True` by default? For simple augmentation in scenarios like ASR / pre-training, it seems like the right thing to do.

pzelasko:
Looks good to me! Can I merge?

desh2608:
> Looks good to me! Can I merge?

Yeah, it's ready to merge :)

pzelasko:
Thanks!

PR #625, opened 2022-03-16
Fix `.subset(first=...)` for lazy manifests without `len`

pzelasko:


lhotse/cut.py
```diff
@@ -3689,12 +3689,12 @@ def subset(
...
             assert first > 0
-            if first > len(self):
+            out = CutSet.from_cuts(islice(self, first))
+            if len(out) < first:
                 logging.warning(
-                    f"CutSet has only {len(self)} items but first {first} required; not doing anything."
+                    f"CutSet has only {len(out)} items but first {first} were requested."
                 )
                 return self
-            return CutSet.from_cuts(islice(self, first))
```
oplatek, line 3697: @pzelasko  You do not return anything if le(out) == first

i.e. the return statement on line 3697 should be one "tab" less indented

oplatek (approved):

PR #626, opened 2022-03-17
More lazily evaluated methods: `map`, `filter`, `repeat`, `shuffle`

pzelasko:
Recently I've been receiving multiple reports from surprised users that lazy manifests do not do some things lazily at all. This PR addresses the biggest pain points by adding lazy variants of `filter` and `shuffle` when called on CutSet etc., it also adds `map` since it was very similar to `filter`, and a new function `repeat` that can be used to multiply some data when joining datasets. 

These operations are composable and can be chained together (e.g. to achieve multiple levels of shuffling, multiple filters, etc.)


pzelasko:
Also added `to_eager()` to trigger the execution of all lazy operations.

Issue #610, opened 2022-03-09
BucketingSampler with equal_duration drops an arbitrary(?) number of cuts.


m-wiesner:
If you use the BucketingSampler with num_buckets, especially with even number of buckets, though I think you can also get it to work for an odd number of buckets, at the very least the last cut is dropped, but if you choose the durations and num_buckets appropriate you can have many more dropped.

The reason this is not caught in the test is that the line

`assert set(cut_set.ids) == set(c.id for c in sampled_cuts)`

is only called for the equal_length method and not the equal_duration method.

The problem arises from the fact that the loops in _create_buckets_equal_duration_single() are not over the cuts, but are instead over the buckets and the duration over the buckets. Because you pop the last cut off of the odd buckets and put it back at the front, you can be left with a non-empty iterator. 

At the very least the BucketingSampler equal_duration feature should be updated with a warning when using it and print out the number of cuts it has dropped. Many times it will be 0, or 1. 

However, if modify

`def test_bucketing_sampler_single_cuts():
    cut_set = DummyManifest(CutSet, begin_id=0, end_id=1000)
    sampler = BucketingSampler(cut_set, sampler_type=SimpleCutSampler, bucket_method="equal_duration", num_buckets=500)
    print(f'Num cuts orig: {len(cut_set)}, Num cuts in sampler: {sum(len(b[0]) for b in sampler.buckets)}')
    sampled_cuts = []
    for batch in sampler:
        sampled_cuts.extend(batch)
    assert set(cut_set.ids) == set(c.id for c in sampled_cuts)`

in test/dataset/sampling/test_sampling.py to look like what you have above, you will see that in that example it drops 25% of the cuts.

Issue #640, opened 2022-03-24
Duration in bucketing sampler


danpovey:
We have a bucketing sampler like this:
```
            train_sampler = BucketingSampler(
                cuts_train,
                max_duration=300,
                shuffle=True,
                num_buckets=30,
                bucket_method="equal_duration",
                drop_last=True,
            )
```
and are getting OOM on a particular batch... printing out the batch details, it looks like the range of durations is particularly large, from 7.3 to 11.4.  This could be because it's in the right tail of the distribution of lengths of the dataset.  It seems the sum of durations is 299, but [longest-duration * num-cuts] is 399.
Perhaps it would make more sense to have the max-duration constraint apply to the longest-cut times max-duration?

pzelasko:
Good idea. I’ll try to add it tomorrow.

pzelasko:
Added in #653

PR #646, opened 2022-03-28
Fixes Bucketing sampler equal duration method that drops cuts (Issue #610)

m-wiesner:
There are likely better ways of doing this, but here is what I've been using as a fix. It's been working fine so far. I modified to the test to check that the bins never overlap, i.e., max_dur_i <= min_dur_{i+1}, except for possibly in the two "middle" bins which is where the left-over cuts that caused overflow in other bins get dumped. 

They get dumped into the middle bins to avoid any big length mismatches that could occur if dumping them in the last bin. It tries to spread out the bins into which it dumps the extra cuts across 2 or 3 bins to avoid having any one bin that is too large.


pzelasko:
Interesting, @danpovey do you think this could help with the issue you mentioned in #640 (aside from/in addition to adding the change you suggested)?

danpovey:
Ah... I don't understand exactly what the issue was though.  I assumed the problem was that with, say, 30 bins, the right tail of the distribution (30th bin) can still have a big range of lengths in it...

pzelasko:
Hey @m-wiesner, I'd like to merge it because your proposed method sounds very interesting. I suggest the following next steps:
- create a third `bucketing_method` which you can call e.g. `equal_duration_fair` :) so that we can both keep the current version and adopt your new version
- ideally you could show the results of two experiments that compare the previous and the new algorithm (if you can show this then I will feel comfortable enough to make your method the new default)
- you'd also need to run black on the code to fix CI issues.

LMK what you think.

pzelasko:
@danpovey the issue is that Matthew discovered that in some unlucky cases, BucketingSampler's `equal_duration` bucketing can drop a very small amount of cuts (typically 1-2), and he was also able to identify some very extreme (and, if I understood correctly, extremely rare) cases in which it might drop a non-trivial portion of the data.

danpovey:
Ah OK.
In my opinion it would be better to avoid a too-big proliferation of
options and methods (e.g. same with the 'strict' method, I think we chould
just have it always be 'strict', now
that it seems to be stable).  If his method is better, I think it would
make more sense just to switch over.  It will cause brief uncertainty (e.g.
about replication) but in the long term
it will simplify things.

drawfish:
Seems that this bug had not fix in master branch yet?  Where in my case, I use BucketingSampler as my validation set sampler, the set contain about 1000 uttrance. However, in DDP traning(where word_size=8) only a small part of the uttrances (<200) left...

pzelasko:
@drawfish your problem might be unrelated; with 8 GPUs, the sampler will split the data between all nodes. If you don't want that, set `world_size=1` and `rank=0` explicitly in the sampler. Let us know if that helped.

pzelasko:
@danpovey OK, I'm fine with just substituting the `equal_duration` option with the new impl. 

@m-wiesner can you confirm that you're getting similar results when training sth compared to the old impl; and can you run black?

drawfish:
> 
Yes, modifed params ```world_size = 1``` and ```rank = 0``` can get all the uttrances.
But it seems that the larger world_size is, the more uttreance drop.
Here is our test case:
```
from lhotse import CutSet, Fbank, FbankConfig
from lhotse.dataset import (
    BucketingSampler,
    K2SpeechRecognitionDataset,
)
from lhotse.dataset.input_strategies import OnTheFlyFeatures
from torch.utils.data import DataLoader

cutset=CutSet.from_jsonl('cutsets.jsonl.gz')
validate = K2SpeechRecognitionDataset(input_strategy=OnTheFlyFeatures(Fbank(FbankConfig(num_mel_bins=80, sampling_rate=8000))),return_cuts=True)

for world_size  in [1, 2, 4, 8]:
  total_num = 0
  for rank in range(world_size):
    sampler = BucketingSampler(cutset, max_duration=100, shuffle=False, world_size=world_size,rank=rank, num_buckets=10)
    valid_dl = DataLoader(validate, sampler=sampler, batch_size=None,persistent_workers=False,num_workers=2,)
    for i in valid_dl:
      total_num+=len(i['supervisions']['text'])
  print(f"World size: {world_size}, total count: {total_num}.")
```
 logging is:
```
World size: 1, total count: 1234.
World size: 2, total count: 1111.
World size: 4, total count: 630.
World size: 8, total count: 59.
```

pzelasko:
Can you test the branch from this PR? If you get the expected results, I can fix remaining comments and merge.

drawfish:
> Can you test the branch from this PR? If you get the expected results, I can fix remaining comments and merge.

The same problem still exists in the branch of this PR.
```
World size: 1, total count: 1234.
World size: 2, total count: 1128.
World size: 4, total count: 642.
World size: 8, total count: 62.
```

pzelasko:
@drawfish turned out the bug you ran into was a feature I added a long time ago, that ensured the number of batches is identical on each node. I fixed it in a new PR, now if you set `drop_last=False`, no data will be dropped, but that means each node can see a different total number of batches -- make sure your code supports that (if it doesn't, you will know because the script will hang waiting for DDP synchronization; but it should not happen during test).

https://github.com/lhotse-speech/lhotse/pull/695

m-wiesner:
Hey. Sorry it took me a while to get back around to this. I'm tested a simple conformer + ctc training on librispeech 100. It's not at all close to state-of-the-art, but I figured the comparison is what matters. The validation numbers posted below are just decoding using the output of the ctc linear layer with argmax.

image

drawfish:
> @drawfish turned out the bug you ran into was a feature I added a long time ago, that ensured the number of batches is identical on each node. I fixed it in a new PR, now if you set `drop_last=False` (default), no data will be dropped, but that means each node can see a different total number of batches -- make sure your code supports that (if it doesn't, you will know because the script will hang waiting for DDP synchronization; but it should not happen during test). With `drop_last=True`, it is still guaranteed that each node sees an equal number of batches.
> 
> #695

Yes, I confirm that the bug fixed at the latest version of lhotse. I think this feature is important in the verification or testing process because there is only a small part of data involved.
By the way, for the DDP synchronization problem, seem that pytorch community had added a `join` operation to DDP(see pytorch[#33148](https://github.com/pytorch/pytorch/issues/33148) and [here](https://github.com/pytorch/pytorch/commit/4e4626a23d1cf8b2a3c87d2b5f2f63ccd14e44c3) ) even though I haven't tested it yet for the reason that I don't think it's important to lose a small part of data in training mode by setting `drop_last=True` ：）.

pzelasko:
@m-wiesner I'm happy to merge but the automatic checks are failing; can you run `black` on the code?

pzelasko:
Closing and resolving remaining issues in #724

PR #649, opened 2022-03-29
RoundRobinSampler: samples mini-batches in turn from each sub-sampler

pzelasko:
not thoroughly tested yet


pzelasko:
Tested with a real workflow and seems to work.

PR #650, opened 2022-03-30
Update black, pytest, speed up some tests

pzelasko:


test/cut/test_custom_attrs_randomized.py
```diff
@@ -4,8 +4,8 @@
 from lhotse.array import seconds_to_frames
 from lhotse.testing.fixtures import RandomCutTestCase
 
-EXAMPLE_TIMEOUT_MS = 1000
-MAX_EXAMPLES = 500
+EXAMPLE_TIMEOUT_MS = 3000
```
pzelasko, line 7: should fix some flaky tests, sometimes they timed out in GitHub due to random reasons.

PR #653, opened 2022-03-31
Stricter batch constraint exceeding checks

pzelasko:
CC @danpovey this should help you with issue #640, I think it does exactly what you described. I wrote a simple test, but would be good if you could test it with an actual experiment before I merge.

### DIFF
--- a/lhotse/dataset/sampling/base.py
+++ b/lhotse/dataset/sampling/base.py
@@ -2,7 +2,7 @@ import warnings
 from copy import deepcopy
 from dataclasses import asdict, dataclass
 from math import isclose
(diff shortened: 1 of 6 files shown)


pzelasko:
I reworked it a bit so that you can switch between the old and new behavior with `strict=True/False` argument for SimpleCutSampler, CutPairsSampler, BucketingSampler, DynamicCutSampler, and DynamicBucketingSampler. By default it's set to `False` (i.e. old behavior).

danpovey:
Thanks!   Sorry I missed this the first time...

pzelasko:
No problem. Let me know if it solves the issue.

PR #656, opened 2022-04-05
Iwslt 2022 dialect

m-wiesner:
Adds recipe for Tunisian Dialect data used in the iwslt2022 dialectal speech translation task.

### DIFF
--- a/lhotse/recipes/iwslt2022_dialect.py
+++ b/lhotse/recipes/iwslt2022_dialect.py
@@ -11,12 +11,14 @@ from concurrent.futures.thread import ThreadPoolExecutor
 import tqdm
 import soundfile as sf
 import string
+
 try:
     import pyarabic.number as number
     from pyarabic import araby
 except ImportError:
     import pip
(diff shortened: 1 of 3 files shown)


lhotse/recipes/iwslt2022_dialect.py
```diff
@@ -0,0 +1,469 @@
...
+import soundfile as sf
+import string
+
+try:
+    import pyarabic.number as number
+    from pyarabic import araby
+except ImportError:
+    import pip
+
+    pip._internal.main(["install", "pyarabic"])
```
pzelasko, line 21: Smart trick, but as a user I would be extremely surprised if importing Lhotse triggered a pip install command. Please handle missing optional dependencies in the same way as throughout the rest of the code, see e.g. here:

https://github.com/lhotse-speech/lhotse/blob/933fb64834d5bc603db1ef8ffd1d2004ae75497a/lhotse/recipes/gigaspeech.py#L38-L43

PR #664, opened 2022-04-12
Fixes to split-lazy

wgb14:
Add prefix and start from 1: change from `0.jsonl.gz` to `cuts_XL_raw.1.jsonl.gz`

However, it's hard to generate `cuts_XL_raw.0001.jsonl.gz`, because we cannot know how many splits would be in advance while doing lazy-loading.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -933,7 +933,9 @@ class RecordingSet(Serializable, AlgorithmMixin):
             )
         ]
 
-    def split_lazy(self, output_dir: Pathlike, chunk_size: int) -> List["RecordingSet"]:
+    def split_lazy(
+        self, output_dir: Pathlike, chunk_size: int, prefix: str
+    ) -> List["RecordingSet"]:
         """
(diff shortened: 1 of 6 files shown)


lhotse/audio.py
```diff
@@ -933,7 +933,9 @@ def split(
             )
         ]
 
-    def split_lazy(self, output_dir: Pathlike, chunk_size: int) -> List["RecordingSet"]:
+    def split_lazy(
+        self, output_dir: Pathlike, chunk_size: int, prefix: str
```
pzelasko, line 937:
```suggestion
        self, output_dir: Pathlike, chunk_size: int, prefix: str = ""
```

It seems like prefix could be optional here (and in other functions as well)

wgb14: Yes, it looks good to have a default value. But if `prefix=""`, `(output_dir / prefix)` doesn't work, what about setting default `prefix="split"`?

wgb14: Fixed

pzelasko:
Thanks!

PR #667, opened 2022-04-12
Fix reading duration when importing piped input data from kaldi

### LINKED ISSUE #659: Loading data
I'm trying to create a Recording using the same "command" that I used in kaldi. I have a binary that converts speex to wav and in wav.scp I do "wav_id spx2wav /path/to/file - |", here I tried new_wav = Recording.from_file("spx2wav /path/to/file - |" which gave me error (file spx2wav /path/to/file - | doesn't exist). Can someone point me to a document that shows how to this? 
I understand that I can creat the wav.scp file and then use "kaldi.load_kaldi_data_dir" and I was wondering if there is a direct way of doing it here.

csukuangfj:
Fixes #659

The original implementation only reads the header to get the wave info. It is efficient when the input is a file.
However, it causes problems when the input is from a pipe as it closes the pipe as soon as it finishes reading the data, even if the pipe writer is still writing.

This PR should fix that. It requires kaldi_native_io v1.5 (i.e., the latest version as of 2022-04-13)

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -35,10 +35,10 @@ def get_duration(
             )
         import kaldi_native_io
 
-        wave_info = kaldi_native_io.read_wave_info(path)
-        assert wave_info.num_channels == 1, wave_info.num_channels
+        wave = kaldi_native_io.read_wave(path)
+        assert wave.data.shape[0] == 1, f"Expect 1 channel. Given {wave.data.shape[0]}"
 
-        return wave_info.duration
+        return wave.duration
     try:
         # Try to parse the file using pysoundfile first.
         import soundfile


pzelasko:
Can we check if the source is a file or a pipe and either read just the header or the whole thing, accordingly?

csukuangfj:
In lhotse, only when the input is from the output of some command is kaldi_native_io used to get the duration. Shall we change that?

pzelasko:
No, I wouldn't change the code in `audio.py`, it uses torchaudio/sndfile/etc. to determine the duration whenever possible. But in `kaldi.py` the `get_duration` function that you are modifying may receive either filenames or pipes from wav.scp. With your latest changes it would always perform a full read, even if wav.scp had a file and not a pipe. I'd simply check if the line ends with `|` and use the new approach if yes, and the old approach if not.

csukuangfj:
> With your latest changes it would always perform a full read, even if wav.scp had a file and not a pipe.

I don't think so.

Please see the code
https://github.com/lhotse-speech/lhotse/blob/56fcdb12ca00e062f9e8db2337b6d41fb47fe3df/lhotse/kaldi.py#L30-L41

You can see that `kaldi_native_io` is used only when the path ends with `|`

pzelasko:
🤦‍♂️Of course you are right, I skimmed too fast over it before. LGTM and thanks!

PR #668, opened 2022-04-13
Read matrix shape information without reading the whole matrix.

csukuangfj:
Requires kaldi_native_io >= v1.7

It supports reading compressed matrices, float, and double matrices.

---

Note:  Matrices from pipes are not supported since it  closes the pipe as soon as it reads the header, which results in pipe write errors for the producer.

Also, it does not support ranges, i.e., `some.ark[0:3]`

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -141,12 +141,12 @@ def load_kaldi_data_dir(
             feature_set = FeatureSet.from_features(
                 Features(
                     type="kaldi_native_io",
-                    num_frames=mat.shape[0],
-                    num_features=mat.shape[1],
+                    num_frames=mat_shape.num_rows,
+                    num_features=mat_shape.num_cols,
                     frame_shift=frame_shift,
                     sampling_rate=sampling_rate,
                     start=0,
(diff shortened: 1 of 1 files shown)


csukuangfj:
If the above two limitations are significant, I will close this PR.

csukuangfj:
BTW: Seems that there are no tests to cover the changes in `kaldi.py`.

pzelasko:
It looks good to me, I would expect that people rarely have pipes specified in `feats.scp`. Thanks!

About the unit test coverage -- you are right. When we started the project, our approach was to build everything brand-new and I didn't care too much about Kaldi compatibility, and treated these functions more as convenience/proof-of-concept. But now it seems that more people are depending on these capabilities.. it would be great if somebody could help and contribute tests for this module (CC @jtrmal @desh2608 @m-wiesner @ngoel17 I know you're using Kaldi import/export, maybe you or somebody in your groups would like to do that?)

jtrmal:
yeah, I agree it needs unit tests
I'll see what can be done
y.

pzelasko (approved):

PR #669, opened 2022-04-13
Count num_digits before split-lazy

wgb14:
change output from `cuts_XL_raw.1.jsonl.gz` to `cuts_XL_raw.0001.jsonl.gz`

this makes `lhotse split-lazy` slightly slower but consistent with `lhotse split`

Feel free to close this PR if this breaks the nature of `lazy-load`

### DIFF
--- a/lhotse/utils.py
+++ b/lhotse/utils.py
@@ -296,13 +296,21 @@ def split_manifest_lazy(
         prefix = "split"
 
     items = iter(it)
+    num_items = 0
+    for item in items:
+        num_items += 1
+    num_splits = math.ceil(num_items / chunk_size)
+    num_digits = len(str(num_splits))
+
+    items = iter(it)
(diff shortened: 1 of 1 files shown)


pzelasko:
I think it's better to hardcode sth like 8 digits for splits (10M splits) and avoid a potentially costly additional iteration over everything. I'm OK if you also prefer to change `.split()` behavior to be consistent with this format.

wgb14:
> I think it's better to hardcode sth like 8 digits for splits (10M splits) and avoid a potentially costly additional iteration over everything. I'm OK if you also prefer to change `.split()` behavior to be consistent with this format.

Good idea. I would prefer to fix 8 digits for splits. Since changing `split()` may break others' existing work.

pzelasko:
Thanks!

PR #672, opened 2022-04-14
~20x faster speed perturbation

pzelasko:
CC @csukuangfj @danpovey this might help you with speeding up on the fly extraction. A while ago I noticed that resampling was the biggest bottleneck in my dataloading code, and then I found that sox uses a very high quality algorithm that is 20-30x slower than torchaudio's sinc interpolation. Since speed perturbation is just resampling, I figured the same speedup can be achieved here. A simple benchmark of new vs old impl:

image

There are slight differences in the signal, but they are not audible to me (note the Y axis scales).. 
image


danpovey:
cool... are those 2 plots before and after the change?  Because there seems to be a big difference in volume if so.

Issue #666, opened 2022-04-12
Try messing with buffer size?


danpovey:
https://github.com/lhotse-speech/lhotse/blob/b05e344bbaf97db316b4c81d7c6d02b068198a59/lhotse/audio.py#L1703

@csukuangfj this is related to the slowness here https://github.com/k2-fsa/icefall/pull/312, of reading ffmpeg data.  I am trying to debug it by doing `strace` on your ffmpeg commands, like this (as you):
```
for p in $(ps -u kuangfangjun | grep ffmpeg | tail -n 1 | awk '{print $1}'); do strace -T -p $p; done >& foo
```
To see which system calls are slow, we can do:
```
awk '{print $NF, $0}' < foo | sort -r | less
# output is:
<0.014715> write(1, "\355\223H<X2^<\303\n\330<.VH=\353Z\231=\203H\305=\216j\341=\366f\347="..., 1280) = 1280 <0.014715>
<0.007298> write(1, "\332\r)=p9,=>\337\t=\200\335\336<L\331\352<&\222\"=`hh=(\360\233="..., 1280) = 1280 <0.007298>
<0.000146> fstat(3, {st_mode=S_IFREG|0664, st_size=9641836, ...}) = 0 <0.000146>
<0.000099> write(2, "ffmpeg version 3.4.8-0ubuntu0.2", 31) = 31 <0.000099>
<0.000086> openat(AT_FDCWD, "/ceph-fj/fangjun/open-source-2/icefall-pruned-multi-datasets/egs/librispeech/ASR/download/GigaSpeech/audio/podcast/P0036/POD0000003582.opus", O_RDONLY) = 3 <0.000086>
<0.000055> write(1, "lW\20\272\202\372\33\272\236\t\4\272\372H\221\271\36 \313\271\200r\374\271\35#\17\272n\30\371\271"..., 1280) = 1280 <0.000055>
<0.000055> getdents(3, /* 72 entries */, 32768)    = 2208 <0.000055>
<0.000052> write(2, " Copyright (c) 2000-2020 the FFm"..., 46) = 46 <0.000052>
<0.000047> write(1, "*\3314>\313\4\21>\253W\221=\270\350\205\273\244\372\256\275\334 \31\2760\27I\276^\26m\276"..., 1280) = 1280 <0.000047>
[...]

PR #675, opened 2022-04-16
Pipe the output of ffmpeg to cat before reading it.

csukuangfj:
See https://github.com/lhotse-speech/lhotse/issues/666#issuecomment-1100574606

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1703,7 +1703,7 @@ def read_opus_ffmpeg(
         cmd += f" -ar {force_opus_sampling_rate}"
         sampling_rate = force_opus_sampling_rate
     # Read audio samples directly as float32.
-    cmd += " -f f32le -threads 1 pipe:1"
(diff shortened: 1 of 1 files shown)


danpovey:
After looking at the size of reads, it seems that this does not resolve the problem.  I suspect it was not really faster, e.g. we were comparing across machines with different properties.  We should double-check though; if it was faster, it is worth investigating why.

pzelasko:
Let me know once you think it's good to merge it.

pzelasko:
Did you reach any conclusions? I can just merge it as it won't break anything, but it would be good to know if it really helps.

danpovey:
Personally I don't believe this can be helping, even though it appeared to.
From strace output, it doesn't seem to affect the chunk sizes it's read in; and I don't believe cat actually does any buffering.
But if the effect can be replicated I suppose we'd have to find an explanation.

Issue #678, opened 2022-04-18
collect_cuts_in_buckets_ slowness?


danpovey:
From looking at the output of `py-spy dump --pid=<training-process>`, I notice that sometimes it seems to be spending time in _collect_cuts_in_buckets.  I suppose I anticipated that this would perhaps happen in a separate data-loader thread, not in the main thread.   Note, this is a run that was started with:`python3 ./pruned_transducer_stateless3/train.py --world-size 8 --num-epochs 40 --start-epoch 20 --full-libri 1 --exp-dir pruned_transducer_stateless3/exp --max-duration 300 --use-fp16 1 --lr-epochs 4 --num-workers 5 --giga-prob 0.5`.
This doesn't dominate the training-- I get this only maybe 10% or 20% of the time-- but it probably makes an appreciable difference.
It would probably be more ideal to have some more incremental way of doing this.
```
Process 2690310: /ceph-fj/fangjun/software/py38/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=6, pipe_handle=8) --multiprocessing-fork
Python v3.8.0 (/ceph-fj/fangjun/software/py38/bin/python3.8)

Thread 2690310 (active+gil): "MainThread"
    _collect_cuts_in_buckets (lhotse/dataset/sampling/dynamic_bucketing.py:349)
        Arguments:
            self: <DynamicBucketer at 0x7f4d2f514b50>
            n_cuts: 217
        Locals:
            _: 133
            cuts: (<MonoCut at 0x7f4d3ef4d460>)
            duration: 5.75
            bucket_idx: 12
    __iter__ (lhotse/dataset/sampling/dynamic_bucketing.py:334)
        Locals:
            ready_buckets: [<collections.deque at 0x7f4f89df0dc0>, <collections.deque at 0x7f4ce7d94820>, <collections.deque at 0x7f4ce5e21e80>, ...]
            sampling_bucket: <collections.deque at 0x7f4f89df0dc0>
            batcher: <DurationBatcher at 0x7f4d3bb12970>
            batch: <CutSet at 0x7f4d4c3010a0>
            batch_size: 217
            _: 216
    _next_batch (lhotse/dataset/sampling/dynamic_bucketing.py:185)
        Arguments:
            self: <DynamicBucketingSampler at 0x7f4d2b838070>
    __next__ (lhotse/dataset/sampling/base.py:256)
        Arguments:
[...]

PR #680, opened 2022-04-19
PoC for reading cuts in background thread in dynamic bucketing

pzelasko:
@danpovey it may address the issue described in #678; but I haven't tested it beyond running unit tests successfully. I added a background thread for collect_cuts_in_buckets. Threading should be sufficient, as I expect the main process CPU to be mostly idle during forward passes on GPU. This implementation should be stable but I don't think it covers every possible edge case of multithreading hazards.. I might transform this into a full blown thread-safe queue kind of thing if you can confirm this helps with the training speed (or ends up in a deadlock when run in real training...)

### DIFF
--- a/lhotse/dataset/sampling/dynamic_bucketing.py
+++ b/lhotse/dataset/sampling/dynamic_bucketing.py
@@ -2,7 +2,9 @@ import random
 import warnings
 from bisect import bisect_right
 from collections import deque
+from concurrent.futures import ThreadPoolExecutor
 from itertools import islice
+from threading import Lock
 from typing import (
     Deque,
     Generator,
@@ -284,6 +286,9 @@ class DynamicBucketer:
             deque() for _ in range(len(duration_bins) + 1)
         ]
 
+        self._cut_reading_thread = ThreadPoolExecutor(1)
+        self._bucket_mutex = Lock()
+
     def __iter__(self) -> Generator[CutSet, None, None]:
         # Init: sample `buffer_size` cuts and assign them to the right buckets.
         self.cuts_iter = iter(self.cuts)
@@ -302,19 +307,20 @@ class DynamicBucketer:
         # On each step we're sampling a new batch.
         try:
             while True:
-                ready_buckets = [b for b in self.buckets if is_ready(b)]
-                if not ready_buckets:
-                    # No bucket has enough data to yield for the last full batch.
-                    non_empty_buckets = [b for b in self.buckets if b]
-                    if self.drop_last or len(non_empty_buckets) == 0:
-                        # Either the user requested only full batches, or we have nothing left.
-                        raise StopIteration()
-                    else:
-                        # Sample from partial batches that are left.
-                        ready_buckets = non_empty_buckets
-                # Choose a bucket to sample from.
-                # We'll only select from the buckets that have a full batch available.
-                sampling_bucket = self.rng.choice(ready_buckets)
+                with self._bucket_mutex:
+                    ready_buckets = [b for b in self.buckets if is_ready(b)]
+                    if not ready_buckets:
+                        # No bucket has enough data to yield for the last full batch.
+                        non_empty_buckets = [b for b in self.buckets if b]
+                        if self.drop_last or len(non_empty_buckets) == 0:
+                            # Either the user requested only full batches, or we have nothing left.
+                            raise StopIteration()
+                        else:
+                            # Sample from partial batches that are left.
+                            ready_buckets = non_empty_buckets
+                    # Choose a bucket to sample from.
+                    # We'll only select from the buckets that have a full batch available.
+                    sampling_bucket = self.rng.choice(ready_buckets)
                 # Sample one batch from that bucket and yield it to the caller.
                 batcher = DurationBatcher(
                     sampling_bucket,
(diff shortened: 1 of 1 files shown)


csukuangfj:
Thanks! I will test it and post the training time with and without this PR.

csukuangfj:
It throws the following exception
```
2022-04-19 16:48:07,508 INFO [train.py:1069] (2/8) Sanity check -- see if any of the batches in epoch 0 would cause OOM.
2022-04-19 16:48:07,611 INFO [asr_datamodule.py:266] (6/8) About to create dev dataset
2022-04-19 16:48:07,623 INFO [asr_datamodule.py:285] (6/8) About to create dev dataloader
2022-04-19 16:48:07,624 INFO [train.py:1069] (6/8) Sanity check -- see if any of the batches in epoch 0 would cause OOM.
2022-04-19 16:48:24,925 INFO [train.py:1010] (0/8) Loading grad scaler state dict
2022-04-19 16:48:24,928 INFO [train.py:1010] (3/8) Loading grad scaler state dict
2022-04-19 16:48:24,929 INFO [train.py:1010] (6/8) Loading grad scaler state dict
2022-04-19 16:48:24,930 INFO [train.py:1010] (7/8) Loading grad scaler state dict
2022-04-19 16:48:24,930 INFO [train.py:1010] (2/8) Loading grad scaler state dict
2022-04-19 16:48:24,932 INFO [train.py:1010] (4/8) Loading grad scaler state dict
2022-04-19 16:48:24,936 INFO [train.py:1010] (5/8) Loading grad scaler state dict
2022-04-19 16:48:24,977 INFO [train.py:1010] (1/8) Loading grad scaler state dict
  File "./pruned_transducer_stateless3/train.py", line 1123, in <module>
    main()
  File "./pruned_transducer_stateless3/train.py", line 1114, in main
    mp.spawn(run, args=(world_size, args), nprocs=world_size, join=True)
  File "/ceph-fj/fangjun/software/py38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 230, in spawn
    return start_processes(fn, args, nprocs, join, daemon, start_method='spawn')
  File "/ceph-fj/fangjun/software/py38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 188, in start_processes
    while not context.join():
  File "/ceph-fj/fangjun/software/py38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 150, in join
    raise ProcessRaisedException(msg, error_index, failed_process.pid)
torch.multiprocessing.spawn.ProcessRaisedException:

-- Process 7 terminated with the following error:
  File "/ceph-fj/fangjun/software/py38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 59, in _wrap
    fn(i, *args)
  File "/ceph-fj/fangjun/open-source-2/icefall-multi-2/egs/librispeech/ASR/pruned_transducer_stateless3/train.py", line 1023, in run
    train_one_epoch(
  File "/ceph-fj/fangjun/open-source-2/icefall-multi-2/egs/librispeech/ASR/pruned_transducer_stateless3/train.py", line 840, in train_one_epoch
    loss_value = tot_loss["loss"] / tot_loss["frames"]
ZeroDivisionError: division by zero
```

lhotse/dataset/sampling/dynamic_bucketing.py
```diff
@@ -284,6 +286,9 @@ def __init__(
             deque() for _ in range(len(duration_bins) + 1)
         ]
 
+        self._cut_reading_thread = ThreadPoolExecutor(1)
```
csukuangfj, line 289: Any reason to not use a process pool? Due to the global interpreter lock, there can be only one running thread at any given time in Python, I think.

pzelasko: Yes, with some setups that use `IterableDatasetWrapper` you are placing the sampler in a dataloader worker process, and AFAIK you can't spawn a nested process pool there because that process is daemonic.

Anyway thread should be sufficient here as I expect the CPU to be mostly idle when running forward and backward passes on GPUs... The reason it didn't work for you is likely the thread could not populate the buckets fast enough and sampler thought they are depleted (race condition). This can be solved with a proper synchronization mechanism but unfortunately I don't have the time to add it right now. I'll return to it sometime.

danpovey:
So do we know how the num-frames could be zero?
I don't know how the forward() could have succeeded if the tensors were empty.

pzelasko:
I suppose the sampler yielded an empty cutset, the dataset somehow didn't crash and collated an empty tensor. It can be fixed with proper synchronization between threads, but to check really quickly if it works, it could be enough to put `time.sleep(5)` after this line of code to allow the buckets to be populated at the start of `__iter__` before their consumption https://github.com/lhotse-speech/lhotse/blob/a9134b9cc18c17bbf790fd15fe2349676578d94b/lhotse/dataset/sampling/dynamic_bucketing.py#L295

pzelasko:
@danpovey @csukuangfj please check if it is faster now (I checked that it does synchronize correctly with the latest changes). In quick local testing I could not see any difference, but maybe you will notice some in your setup.

csukuangfj:
> @danpovey @csukuangfj please check if it is faster now (I checked that it does synchronize correctly with the latest changes). In quick local testing I could not see any difference, but maybe you will notice some in your setup.

I will test it when we have free GPUs.

SongLi89:
Hi, Is this PR merged? Maybe I have similar problems, that reading the Cuts is not so fast.

pzelasko:
No, it hasn’t been merged — I didn’t find any difference with this implementation in quick testing. Can you describe your environment a bit more? What’s your sampler, max_duration, num_workers, data size, are you reading audio or features, etc. Also I recommend running `py-spy` on your script (or dataloading worker processes) to understand where the time is being spent.

PR #693, opened 2022-05-02
WIP kaldi import/export refactoring proof of concept/design

jtrmal:
This is just an attempt at refactoring the code into a few classes to enable override/specify input or output format for specifically formatted corpora. For now, no customization is enabled, it's just the refactoring itself to discuss the design.

Passes the kaldi tests, tho

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -1,4 +1,3 @@
-import warnings
 from collections import defaultdict
 from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


jtrmal:
seems like the test fail is because of something wrong with wpe

lhotse/kaldi.py
```diff
@@ -15,6 +14,247 @@
     fastcopy,
     is_module_available,
 )
+import contextlib
+
+
+class KaldiFormatterBase:
```
pzelasko, line 20: It seems like both input and output formatters are context managers and don't have a common interface otherwise; not sure we need a common base class for that.

lhotse/kaldi.py
```diff
@@ -50,6 +290,22 @@ def get_duration(
     return info.duration
 
 
+class KaldiDirectory:
+    def __init__(self, path: Pathlike, options: dict):
```
pzelasko, line 294: options as dict is are a little hacky, why not define keyword arguments?

pzelasko (commented):
Sorry for commenting on it so late; first I postponed it and then I forgot about it for a while.

Why do you want to refactor this? I'm not opposed to it but as-is I don't see the benefit. Are you intending to add more types of input / output formatters? What would they support?

jtrmal:
I plan to get back to this once I have a little bit more bandwidth available

jtrmal:
let me try to revive this today... I feel like I might have different
opinion on how I did stuff after I had it on backburner for a few months
But feel free to comment on this if you feel you have a better idea(s)
y.

Issue #701, opened 2022-05-06
HTTP Error 404 when using "Lhotse download ali-meeting --force-download download/"


luomingshuang:
When I want to use "Lhotse download ali-meeting --force-download download/" to download the ali-meeting data, there is a bug showing as follows:
```
(k2-python) luomingshuang@de-74279-k2-train-1-0307195509-54c966b95f-rtpfq:~/codes/icefall-ali-meeting/egs/ali_meeting/ASR$ lhotse download ali-meeting --force-download download/
Downloading Train_Ali_far.tar.gz: 0.00B [00:01, ?B/s]
Traceback (most recent call last):
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/bin/lhotse", line 33, in <module>
    sys.exit(load_entry_point('lhotse==1.2.0.dev0+git.cda63b5.clean', 'console_scripts', 'lhotse')())
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/lib/python3.8/site-packages/click-8.0.1-py3.8.egg/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/lib/python3.8/site-packages/click-8.0.1-py3.8.egg/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/lib/python3.8/site-packages/click-8.0.1-py3.8.egg/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/lib/python3.8/site-packages/click-8.0.1-py3.8.egg/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/lib/python3.8/site-packages/click-8.0.1-py3.8.egg/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/ceph-meixu/luomingshuang/anaconda3/envs/k2-python/lib/python3.8/site-packages/click-8.0.1-py3.8.egg/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
[...]

PR #702, opened 2022-05-06
Minor fix in base url for AliMeeting download

desh2608:
This should fix https://github.com/lhotse-speech/lhotse/issues/701


pzelasko:
Thanks!

PR #703, opened 2022-05-07
Avoid being converted to ASCII when preparing manifest

luomingshuang:
When I use the lhotse to prepare manifest for aidatatang_200zh, the text in manifest is converted to ASCII. So I add `ensure_ascii=False` to avoid this case. And after adding it, the text in manifest is normal (char).

### DIFF
--- a/lhotse/serialization.py
+++ b/lhotse/serialization.py
@@ -61,7 +61,7 @@ class YamlMixin:
 def save_to_json(data: Any, path: Pathlike) -> None:
     """Save the data to a JSON file. Will use GZip to compress it if the path ends with a ``.gz`` extension."""
     with open_best(path, "w") as f:
-        json.dump(data, f, indent=2)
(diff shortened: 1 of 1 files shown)


csukuangfj:
The following is what I have:
```
$ head -n15  supervisions_train.json
[
  {
    "id": "T0055G3010S0310",
    "recording_id": "T0055G3010S0310",
    "start": 0.0,
    "duration": 2.142875,
    "channel": 0,
    "text": "\u5176\u5b9e \u597d\u73a9 \u7684 \u8f6f\u4ef6",
    "language": "Chinese",
    "speaker": "G3010"
  },
```

What is your output when `ensure_ascii=False`?

luomingshuang:
After using `ensure_ascii=False`, my output shows as follows:
972c35ac21452ec53a4caca92795ab7

csukuangfj:
Ok, I think the two are equivalent, but  using `ensure_ascii=False` produces a more readable output.

PR #709, opened 2022-05-13
adding Earnings-21 dataset from rev-dot-com

jtrmal:
I'm ready to add Earnings-22 as well, as soon as this will pass the review (the corpus structure is similar)

PR #712, opened 2022-05-16
Normalize output path names for recipes

desh2608:
This PR "normalizes" the output manifests for the recipes:
* All JSON type outputs are changed to JSONL (ses [this comment](https://github.com/lhotse-speech/lhotse/pull/709#discussion_r872698200))
* Manifests are named as `<corpus-name>_<manifest-type>_<part>.jsonl`, e.g., "ami_recordings_dev.jsonl". This is to ensure that manifests from multiple corpora can be saved to same directory without name-based confusion as would occur with names like "recordings_dev.jsonl".

Obviously, we would have to modify the data preparation in icefall to be compatible with these changes, but I feel it's better to switch to this normalized naming scheme now rather than when it would break a lot more recipes.

### DIFF
--- a/docs/corpus.rst
+++ b/docs/corpus.rst
@@ -168,6 +168,14 @@ Adding new corpora
     ``{'train': {'recordings': <RecordingSet>, 'supervisions': <SupervisionSet>}, 'dev': ...}``
 
 .. hint::
+    **Manifest naming convention.** The default naming convention is ``<corpus-name>_<manifest-type>_<split>.jsonl.gz``,
+    i.e., we save the manifests in a compressed JSONL file. Here, ``<manifest-type>`` can be ``recordings``, 
+    ``supervisions``, etc., and ``<split>`` can be ``train``, ``dev``, ``test``, etc. In case the corpus
+    has no such split defined, we can use ``all`` as default. Other information, e.g., mic type, language, etc. may 
+    be included in the ``<corpus-name>``. Some examples are: ``cmu-indic_recordings_all.jsonl.gz``,
+    ``ami-ihm_supervisions_dev.jsonl.gz``, ``mtedx-english_recordings_train.jsonl.gz``.
+
+.. hint::
     **Isolated utterance corpora.** Some corpora (like LibriSpeech) come with pre-segmented recordings.
     In these cases, the :class:`~lhotse.supervision.SupervisionSegment` will exactly match the
     :class:`~lhotse.recording.Recording` duration
--- a/lhotse/recipes/adept.py
+++ b/lhotse/recipes/adept.py
@@ -147,7 +147,7 @@ def prepare_adept(
 
     if output_dir is not None:
         output_dir = Path(output_dir)
-        supervisions.to_file(output_dir / "adept_supervisions.json")
-        recordings.to_file(output_dir / "adept_recordings.json")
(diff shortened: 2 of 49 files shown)


desh2608:
BTW about 1/4th of the recipes already follow this naming (e.g. gigaspeech, spgispeech, swbd, ...)

pzelasko:
A big +1 from me, thank you for this effort 🙏🏻 

Some suggestions:
- we could adopt a convention that corpora without splits would be named `<corpus-name>_<manifest-type>_all.jsonl`.
- when a corpus specifies multiple types of sub-splits (e.g. mtedx has language) put these extra parts before `<manifest-type>` so that if we do `parts = path.stem.split("_")` then the following two are always true: `parts[-1] == '<train/dev/test/all-split>'`, and `parts[-2] == <manifest-type>`, maybe that can help some automation if people combine more data together.

desh2608:
@pzelasko @jtrmal both of those suggestions make sense to me. I didn't add any gzipping since my understanding was we mainly do it for large(ish) manifests, but I can add that too if it's the consensus.

jtrmal:
I defer to Piotr's opinion :)
y.

pzelasko:
I don't see anything to lose and I guess it makes people's lives easier.

desh2608:
Done :)

pzelasko:
Awesome!

ngoel17:
@desh2608 - Just adding my comment that I sent to Piotr. I did this several weeks ago - and the motivation was that I wanted to use recipes/utils on kaldi imported data.

I found that if I do the following modifications to the recipes/utils.py my life becomes super easy.  But this will not work with the lhotse recipes because the naming convention for kaldi imports is different than the convention for lhotse recipes. 

 How about making both the same? i.e. dataset_name/supervisons.jsonl.gz  etc.  Please look.
---------------------------
:param prefix: Optional common prefix for the manifest files (underscore is automatically added).

@@ -39,8 +42,6 @@ def read_manifests_if_cached(

     """

     if output_dir is None:

         return {}

-    if prefix and not prefix.endswith("_"):

-        prefix = f"{prefix}_"

     if suffix.startswith("."):

         suffix = suffix[1:]

     if lazy and not suffix.startswith("jsonl"):

@@ -48,17 +49,19 @@ def read_manifests_if_cached(

             f"Only JSONL manifests can be opened lazily (got suffix: '{suffix}')"

         )

     manifests = defaultdict(dict)

-    for part in dataset_parts:

-        for manifest in types:

-            path = output_dir / f"{prefix}{manifest}_{part}.{suffix}"

-            if not path.is_file():

-                continue

-            if lazy:

-                manifests[part][manifest] = TYPES_TO_CLASSES[manifest].from_jsonl_lazy(

-                    path

-                )

-            else:

-                manifests[part][manifest] = load_manifest(path)

+    for dataset in datasets:

+        for part in dataset_parts:

+            for manifest in types:

+                path = output_dir / f"./{prefix}/{dataset}/{manifest}{part}.{suffix}"

+                if not path.is_file():

+                    [logging.info](http://logging.info/)(f'{path} not found')

+                    continue

+                if lazy:

+                    manifests[dataset][manifest] = TYPES_TO_CLASSES[manifest].from_jsonl_lazy(

+                        path

+                    )

+                else:

+                    manifests[dataset][manifest] = load_manifest(path)

     return dict(manifests)

PR #713, opened 2022-05-16
Adding the second revdotcom's earnings corpus

jtrmal:
For now, let's ignore download of the corpus, it needs advanced git features that I don't feel I would be able to implement reliably.
Uses desh's naming policy (I think/hope)
y.


pzelasko:
Thanks!

PR #715, opened 2022-05-17
[ali-meeting] Change jsonl to json

luomingshuang:
I think changing jsonl to json is better for preparing lang files when using.

### DIFF
--- a/lhotse/recipes/ali_meeting.py
+++ b/lhotse/recipes/ali_meeting.py
@@ -172,12 +172,8 @@ def prepare_ali_meeting(
         validate_recordings_and_supervisions(recording_set, supervision_set)
 
         if output_dir is not None:
-            supervision_set.to_file(
-                output_dir / f"alimeeting_supervisions_{part.lower()}.jsonl.gz"
-            )
-            recording_set.to_file(
-                output_dir / f"alimeeting_recordings_{part.lower()}.jsonl.gz"
-            )
+            supervision_set.to_json(output_dir / f"supervisions_{part.lower()}.json")
+            recording_set.to_json(output_dir / f"recordings_{part.lower()}.json")
 
         manifests[part.lower()] = {
             "recordings": recording_set,


lhotse/recipes/ali_meeting.py
```diff
@@ -172,12 +172,8 @@ def prepare_ali_meeting(
         validate_recordings_and_supervisions(recording_set, supervision_set)
 
         if output_dir is not None:
-            supervision_set.to_file(
-                output_dir / f"alimeeting_supervisions_{part.lower()}.jsonl.gz"
-            )
-            recording_set.to_file(
-                output_dir / f"alimeeting_recordings_{part.lower()}.jsonl.gz"
-            )
+            supervision_set.to_json(output_dir / f"supervisions_{part.lower()}.json")
```
danpovey, line 175: There are various things that we can't do so easily with json files, including splitting.
Reading json files uses much more memory, you cannot just read a subset or read items one by one.

luomingshuang: OK.

luomingshuang:
There is a question. If the files in data/manifest for recording and supervision are jsonl.gz, it seems that I can not use `CutSet.from_manifests` to read it. And also, I can not use `CutSet.from_file` to read the two jsonl.gz at the same time.

csukuangfj:
> There is a question. If the files in data/manifest for recording and supervision are jsonl.gz, it seems that I can not use `CutSet.from_manifests` to read it. And also, I can not use `CutSet.from_file` to read the two jsonl.gz at the same time.

You can find example code in icefall for how to handle jsonl.gz files.

See  https://github.com/k2-fsa/icefall/blob/master/egs/gigaspeech/ASR/local/preprocess_gigaspeech.py#L57
```
    manifests = read_manifests_if_cached(
        dataset_parts=dataset_parts,
        output_dir=src_dir,
        prefix="gigaspeech",
        suffix="jsonl.gz",
    )
```

You can look into how `read_manifests_if_cached` is implemented in lhotse.

luomingshuang:
I add suffix="jsonl.gz" for read_manifests_if_cached, it works well. Thanks.

desh2608:
@luomingshuang `suffix="jsonl.gz"` is also set as the default after [this PR](https://github.com/lhotse-speech/lhotse/pull/712), since all recipes export files with this suffix now.

PR #716, opened 2022-05-17
Update examples of combining datasets with RoundRobinSampler and add `stop_early` option.

pzelasko:


jtrmal:
That reviewNB thing is so pretty/useful.

PR #725, opened 2022-05-28
Update mgb2.py

AmirHussein96:
extra new line causes error with load_kaldi_data_dir()

### DIFF
--- a/lhotse/recipes/mgb2.py
+++ b/lhotse/recipes/mgb2.py
@@ -130,7 +130,7 @@ def prepare_mgb2(
             ) as f_out:
                 for line in f_in:
                     f_out.write(line.replace("wav/", f"{corpus_dir}/{part}/wav/"))
-                    f_out.write("\n")
+                    
 
             recordings, supervisions, _ = load_kaldi_data_dir(
                 (output_dir / part), 16000
--- a/lhotse/recipes/mgb2.py
+++ b/lhotse/recipes/mgb2.py
@@ -183,16 +183,18 @@ def prepare_mgb2(
     return manifests
 
 
-_unicode = "\u0622\u0624\u0626\u0628\u062a\u062c\u06af\u062e\u0630\u0632\u0634\
-    \u0636\u0638\u063a\u0640\u0642\u0644\u0646\u0648\u064a\u064c\u064e\u0650\u0652\u0670\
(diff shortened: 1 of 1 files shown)


csukuangfj:
Should `load_kaldi_data_dir()` skip empty lines?

AmirHussein96:
> Should `load_kaldi_data_dir()` skip empty lines?

Yes, I think so. The way it works now is that It assumes that each line in the wav will be split into list of two items [id, text]. The additional empty line adds third item and causes an error. One can also fix this issue from the `load_kaldi_data_dir()` by skipping/removing the empty lines.

lhotse/recipes/mgb2.py
```diff
@@ -130,7 +130,7 @@ def prepare_mgb2(
             ) as f_out:
                 for line in f_in:
                     f_out.write(line.replace("wav/", f"{corpus_dir}/{part}/wav/"))
-                    f_out.write("\n")
```
pzelasko, line 133: Wouldn’t the script write everything to a single line after this change?

pzelasko (commented):
Why don’t we fix Kaldi utilities to skip empty lines instead?

danpovey:
> Should `load_kaldi_data_dir()` skip empty lines?

Kaldi data dirs have a well defined format that does not allow empty lines.

lhotse/recipes/mgb2.py
AmirHussein96: The lines read from the wav.scp file already contain a new line, so the additional "\n" adds empty line. However I agree that it will be cleaner to just filter out the empty lines from any files that is being read from kaldi folder.

pzelasko: As Dan pointed out, Kaldi doesn’t allow empty lines, so let’s strive to produce correct outputs. Please fix style issues and LGTM.

AmirHussein96: The new line was coming from the line `f_out.write("\n")` which I removed now. I tested it and now it works perfectly with load_kaldi_data_dir().

pzelasko: Cool. Please fix black issues and LGTM.

AmirHussein96: Done

pzelasko:
Thanks!

PR #734, opened 2022-06-03
Fix heroico regex strings

jtrmal:
I noticed a couple of warnings, so I fixed the strings. The grid seems offline now, so I cannot connect and rebase, so feel free to review and then wait, I will rebase so that the commit history is clean


pzelasko:
Thanks! I’ll just squash and merge.

PR #735, opened 2022-06-05
[Ready to merge] Update lhotse/recipes/aishell4.py

luomingshuang:
I am trying to build an Aishell4 recipe for icefall. But when using the function `read_manifests_if_cached` to load recordings and supervisions, it happens an error. The error shows that it is due to the improper name of the recording and supervision jsonl.gz file. So I did a change to it. And it runs normally.

### DIFF
--- a/lhotse/recipes/aishell4.py
+++ b/lhotse/recipes/aishell4.py
@@ -144,9 +144,7 @@ def prepare_aishell4(
         validate_recordings_and_supervisions(recording_set, supervision_set)
 
         if output_dir is not None:
-            supervision_set.to_file(
-                output_dir / f"supervisions_{part}.jsonl.gz"
-            )
+            supervision_set.to_file(output_dir / f"supervisions_{part}.jsonl.gz")
             recording_set.to_file(output_dir / f"recordings_{part}.jsonl.gz")
 
         manifests[part] = {"recordings": recording_set, "supervisions": supervision_set}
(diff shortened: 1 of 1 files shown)


csukuangfj:
Have you tried to add `prefix=aishell4` and `suffix="jsonl.gz"` while invoking `read_manifests_if_cached`?

lhotse/recipes/aishell4.py
```diff
@@ -144,10 +144,8 @@ def prepare_aishell4(
         validate_recordings_and_supervisions(recording_set, supervision_set)
 
         if output_dir is not None:
-            supervision_set.to_file(
-                output_dir / f"aishell4_supervisions_{part}.jsonl.gz"
-            )
-            recording_set.to_file(output_dir / f"aishell4_recordings_{part}.jsonl.gz")
+            supervision_set.to_file(output_dir / f"supervisions_{part}.jsonl.gz")
```
csukuangfj, line 147: I think it is intentional to use `aishell4_supervions_xxx.jsonl.gz` instead of `supervisions_xxx.jsonl.gz`.

Please modify your code to add `prefix` and `suffix` to `read_manifests_if_cached()`.

luomingshuang: Oh, I noticed that the Lhotse recipes has been modified comparing with before. (All add the prefix.)

luomingshuang: Thanks.

Issue #740, opened 2022-06-09
Error while preparing CommonVoice 


mohsen-goodarzi:
Hi,
I manually downloaded the latest CommonVoice dataset for Dutch (cv-corpus-9.0-2022-04-27-nl), and then tried to prepare the manifest using lhotse by following command:

`lhotse prepare commonvoice -l nl ./cv-corpus-9.0-2022-04-27 ./manifests`

But I get the following error for every entry in the dataset and final manifest files are empty:

```
2022-06-09 16:08:02,713 ERROR [commonvoice.py:247] Error when processing TSV file: line no. 10551: 'client_id     d09ab3464067c2c3d93c079cbe442fa0f029f8e391d382...
path                               common_voice_nl_26946122.mp3
sentence      Ik hoop, beste collega's, dat er morgen nieman...
up_votes                                                      2
down_votes                                                    0
age                                                    twenties
gender                                                     male
accents                                   Nederlands Nederlands
locale                                                       nl
segment                                                     NaN
Name: 10551, dtype: object'.
Original error type: '<class 'AttributeError'>' and message: 'Series' object has no attribute 'accent'
```

I used a fresh installation of lhotse using following commands in a new venv, so it must be easily reproducible:

```
pip install git+https://github.com/lhotse-speech/lhotse
pip install pandas
```

mohsen-goodarzi:
@csukuangfj 
I noticed that the `accent` column in  `cv-corpus-5.1-2020-06-22` contains just single word values but in `cv-corpus-9.0-2022-04-27-nl`, they are two words separated by space like:  `Nederlands Nederlands` or `Frans Nederlands`.
Could it be the reason?

pzelasko:
It seems they changed `accent` column to `accents` column in the new release; it seems like it should be fairly easily fixable if we assume we don't support the older CV versions, or we can add some logic to resolve the version and parse the right column. Do you mind submitting a PR?

mohsen-goodarzi:
Of course not. 
Please see PR #743.

desh2608:
Fixed in https://github.com/lhotse-speech/lhotse/pull/743.

PR #743, opened 2022-06-10
fix CommonVoice prepare

mohsen-goodarzi:
This PR will solve the prepare step of recent versions of CommonVoice mentioned here #740.
In recent CommonVoice, they changed the `accent `column to `accents`.
This solution should be backward compatible, but the corresponding field in supervision json will be `accents `instead of `accent`.


lhotse/recipes/commonvoice.py
```diff
@@ -263,6 +263,11 @@ def parse_utterance(
         raise ValueError(f"No such file: {audio_path}")
     recording_id = Path(row.path).stem
     recording = Recording.from_file(audio_path, recording_id=recording_id)
+    # Handling accent(s) in different versions of CommonVoice
+    if "accents" in row:
```
csukuangfj, line 267: I usually see people using `hasattr(row, 'accents')`.
The operator `in` is commonly used together with `list`, `set` and`dict`.

Not sure whether using the operator `in` is appropriate here.

pzelasko: AFAIK `in` checks for column existence in pandas objects, so it looks OK to me.

pzelasko: Actually `hasattr` might not work, I think pandas "columns-as-members" thing works only dynamically via `getattr` (unless they explicitly added support for `hasattr`)

csukuangfj: Ah, thanks. I was not aware that it is a pandas object.

PR #757, opened 2022-06-23
Prototyping scaling down batch sizes with very long cuts

pzelasko:
The main idea is to automatically decrease the batch size in presence of "very long" examples. Two main questions are:
- how much to decrease the batch size
- how long is "very long"

The goal is to use GPU memory more efficiently (i.e. not have to decrease `max_duration` too much because of 1% of batches with long examples).

I'll test a few variants and see if it helps.


pzelasko:
I forgot that I even started to prototype this solution before. Solved by #950

PR #764, opened 2022-07-04
Add support for Python 3.10 and PyTorch 1.12

pzelasko:


pzelasko:
There is a "hacky" workaround for changes in torchaudio 0.12 -- we are reading full MP3 files to get their number of samples. See issue https://github.com/pytorch/audio/issues/2524. If we can find a better workaround I'll update the code.

PR #773, opened 2022-07-08
Fix restoring sampler state with `world_size>1`

pzelasko:


pzelasko:
Merging this late but better than never :)

PR #789, opened 2022-08-19
Support also kaldi data dierectories without segments file.

MartinKocour:
Now, when kaldi data directory does not contain `segments` file, supervisions (text, speakers, etc.) are not processed.

This pull request basicaly solves this issue by assuming that provided supervision (text, speakers, gender, etc.) correspondents to whole recordings.


jtrmal:
I like

pzelasko:
LGTM! Thanks @MartinKocour!

PR #797, opened 2022-08-29
add test data to bvcc

oplatek:
Adding processing for the test split of the BVCC VoiceMOS Challange data


pzelasko:
LMK when you think it's good to merge

oplatek:
Few experiments finished without a problem with the data. They seem valid. Feel free to merge.

PR #799, opened 2022-09-01
Add reverb with fast RIR generator

desh2608:
This PR adds the functionality to synthesize a room impulse response on-the-fly for using the `reverb_rir()` methods in the `Cut` or `Recording` (see discussion in https://github.com/lhotse-speech/lhotse/discussions/787). The algorithm is based on [this paper](https://arxiv.org/abs/2208.04101). Here is an audio sample: [Google drive link](https://drive.google.com/file/d/1RGF5rIc66lkWVye100Mi1mhu-lzzxwkT/view?usp=sharing).

**Usage:** 
Basically, all the same methods would still work, but will use the random RIR generator if no RIR recording is explicitly provided.

### DIFF
--- a/lhotse/augmentation/utils.py
+++ b/lhotse/augmentation/utils.py
@@ -2,8 +2,6 @@ from typing import List
 
 import numpy as np
 import torch
-from torchaudio.functional import highpass_biquad
-from torchaudio.transforms import Resample
 
 try:
     # Pytorch >= 1.7
@@ -76,7 +74,7 @@ def convolve1d(signal: torch.Tensor, kernel: torch.Tensor) -> torch.Tensor:
     return result[:padded_size]
 
 
-# The following is taken from: https://github.com/yluo42/FRA-RIR/blob/main/FRA-RIR.py
+# The following is based on: https://github.com/yluo42/FRA-RIR/blob/main/FRA-RIR.py
 def generate_fast_random_rir(
     nsource: int = 1,
     sr: int = 16000,
@@ -98,6 +96,16 @@ def generate_fast_random_rir(
     :param tau: controlling the relationship between the distance and the number of reflections of each virtual sound source. Default: 0.25.
     :return: simulated RIR filter for all sources, shape: (nsource, nsample)
     """
+    from torchaudio.functional import highpass_biquad
+    from lhotse.augmentation.torchaudio import get_or_create_resampler
+
+    # the sample rate at which the original RIR filter is generated
+    ratio = 64
+    sample_sr = sr * ratio
+
+    # two resampling operations
+    resample1 = get_or_create_resampler(sample_sr, sample_sr // int(np.sqrt(ratio)))
+    resample2 = get_or_create_resampler(sample_sr // int(np.sqrt(ratio)), sr)
 
     eps = np.finfo(np.float16).eps
 
@@ -113,10 +121,6 @@ def generate_fast_random_rir(
     # number of virtual sound sources
     image = sr * 2
 
-    # the sample rate at which the original RIR filter is generated
-    ratio = 64
-    sample_sr = sr * ratio
-
     # sound velocity
     velocity = 340.0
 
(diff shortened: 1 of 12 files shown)


lhotse/augmentation/utils.py
```diff
@@ -1,4 +1,9 @@
+from typing import List
+
+import numpy as np
 import torch
+from torchaudio.functional import highpass_biquad
+from torchaudio.transforms import Resample
```
pzelasko, line 6: Can you remove torchaudio imports from the global namespace into the local function you added? They tend to crash documentation builds last time I checked, because they tried to load some shared library that is not available in that env.

Also, I suggest importing `get_or_create_resampler` from `lhotse.augmentation.torchaudio` (locally in the function) as it will create a global resampler, caching the filter coefficients (I got a big speedup from this previously).

lhotse/augmentation/utils.py
```diff
@@ -69,3 +74,134 @@ def convolve1d(signal: torch.Tensor, kernel: torch.Tensor) -> torch.Tensor:
     result = irfft(f_result, n=fast_ftt_size)
 
     return result[:padded_size]
+
+
+# The following is taken from: https://github.com/yluo42/FRA-RIR/blob/main/FRA-RIR.py
+def generate_fast_random_rir(
```
pzelasko, line 80: If it turns out to be too slow, I would check if it's faster with pure `numpy` that with CPU `torch`, that often turned out to be true for me. If it's too much hassle it's OK to leave as is.

lhotse/augmentation/utils.py
desh2608: Done. Thanks for the tip!

desh2608:
Some time comparisons:

1. Reverb with RIR specified
```python
In [14]: %timeit reverb_cut_rir(c,r)
The slowest run took 8.14 times longer than the fastest. This could mean that an intermediate result is being cached.
148 ms ± 88.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

2. Reverb with generated RIR (torchaudio `Resample`)
```python
In [9]: %timeit reverb_cut(c)
179 ms ± 57.7 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

3. Reverb with generated RIR (cached `Resample`)
```python
In [15]: %timeit reverb_cut(c)
138 ms ± 59.8 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

4. Using Numpy instead of torch
```python
In [6]: %timeit reverb_cut(c)
281 ms ± 121 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Based on above, I'm keeping the torch operations with cached resamplers.

pzelasko:
Wow, nice speed up on PyTorch side. Let’s fix the style checks and LGTM!

pzelasko:
Desh do you have any insight how much simulated vs real RIR matters for tasks such as ASR or diarization? I wonder if given this implementation it still makes sense to use RIR datasets. The audio sample you uploaded sounds nice actually.

desh2608:
> Desh do you have any insight how much simulated vs real RIR matters for tasks such as ASR or diarization? I wonder if given this implementation it still makes sense to use RIR datasets. The audio sample you uploaded sounds nice actually.

I haven't run comparisons myself yet, but at least in the paper it seems to match the results using real RIRs on speech enhancement tasks. Also, older studies have found that at least for ASR, simulated vs real RIRs do not make much difference, so I would imagine that this method should be comparable with those. 

In any case, I am hoping to prepare an AMI recipe for icefall soon, and I can conduct such a comparison then.

pzelasko:
Thanks!

lhotse/augmentation/utils.py
mthrok: > Can you remove torchaudio imports from the global namespace into the local function you added? They tend to crash documentation builds last time I checked, because they tried to load some shared library that is not available in that env.

@pzelasko Do you have a log of the crash? We make sure that PyTorch is the only hard requirement. I am curious to see what it is.

pzelasko: Sure, let me check it again in a spare moment. I think it was related to loading libsndfile.so which was not present in that system. It was also a while ago so maybe the current versions don't have that issue.

PR #807, opened 2022-09-13
Fix read and write in piped CLI

desh2608:


desh2608:
After these changes, the comparison works, but there are still some issues. For example, the following works fine:
```
lhotse subset --first 5 data/librispeech/raw_cuts_dev-clean.jsonl.gz - | cat > tmp.jsonl
```
But the following does not:

Command: `cat tmp.jsonl | lhotse subset --first 10 - tmp2.jsonl`
Desired result: Create `tmp2.jsonl` with first 10 cuts
Actual result: `tmp2.jsonl` is created correctly, but first 10 cuts are also printed to stdout (with blank line between them).

desh2608:
The following work well now:
**Example 1:** `lhotse cut truncate -d 10 --preserve-id data/librispeech/raw_cuts_dev-clean.jsonl.gz - | cat > tmp.jsonl`
**Example 2:** `cat tmp.jsonl | lhotse subset --first 10 - tmp2.jsonl`
**Example 3:** `lhotse cut truncate -d 10 --preserve-id data/librispeech/raw_cuts_dev-clean.jsonl.gz - | lhotse subset --first 10 - tmp2.jsonl`

There is still an error in the `split` command but I don't think that it is related to piping; I will create a separate issue for that.

pzelasko:
Thanks!

pzelasko (approved):

PR #809, opened 2022-09-14
Default behavior of CutSet.mix

ZuoyunZheng:
Change the default behavior of `CutSet.mix` where when no `duration` is passed, as many `Cut`s from the other `CutSet` will be sampled until the complete source `Cut` is mixed.


pzelasko:
Looks good to me! Can you please fix the formatting checks and the unit test that asserted the old behavior of `CutSet.mix`?

pzelasko:
Thanks!

PR #814, opened 2022-09-20
Audio backend refactoring and a workaround for FLAC reading from/writing to in-memory buffers

pzelasko:
This PR went a little out of scope, I didn't originally intend the refactoring but the audio errors were too confusing even for me at this point. I think I will follow-up with a similar refactoring that @desh2608 did in #820 to improve the code readability. I might also further refactor the "audio backend" thing in Lhotse, apparently it grew quite complicated as it tries to support multiple torchaudio versions, multiple torchaudio backends, libsoundfile, audioread, and custom hacks for things like OPUS and SPHERE, and now also custom hacks for in-memory buffers. I'll need to think if the solution can be made a bit more elegant.

BTW I tried the suggested workaround solution to https://github.com/pytorch/audio/issues/2662 by using the new torchaudio ffmpeg streamer, but I had some issues with some (not all) in-memory buffers saying that seek operation is not permitted, but I could not easily create a reproducible example yet. Maybe I'll revisit this in the future.


desh2608:
Nice! I think most users (me, for one) would care more about what functionalities a `Recording` or `RecordingSet` provide, compared to nitty-gritties of how the audio is loaded in the backend. Probably best to separate these 2 concepts.

PR #815, opened 2022-09-20
Adding more info about resampling options

RuABraun:
Could add it also to other locations but am uncertain where else could be appropriate.

I'm assuming it only makes sense to talk about resample() for CutSet/RecordingSet (not sure if there are other data constructs that could call resample() ?).


pzelasko:
Thanks, looks good! Technically you can call it on single `Recording` or `Cut` objects but I think the message is good enough.

Issue #779, opened 2022-07-26
Pad CutSet with silence?


danpovey:
I'm wondering whether there is an easy way to pad all the cuts in a CutSet with a fixed amount of silence on the sides, as in extend()?  This is for things like debugging edge issues with models, or decoding and making sure final symbols are output.
I tried extend_by(), but it appears to be a no-op, e.g.
`a.extend_by(0.05, preserve_id=True).fill_supervisions()`
seemed to return the original CutSet (possibly because it only extended to the edge of the original recordings?)
I ended up doing this with something like:
```
 load_manifest(self.args.manifest_dir / "cuts_test-clean.json.gz").pad(direction='both').fill_supervisions()
```
just for testing purposes, which isn't ideal because it make all utterances as long as the longest one, which is not what I really wanted.

pzelasko:
Try this one https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/cut_transforms/extra_padding.py by passing it to the Dataset as a cut transform.

danpovey:
When I add the cut_transforms line below:
```
        test = K2SpeechRecognitionDataset(
            input_strategy=OnTheFlyFeatures(Fbank(FbankConfig(num_mel_bins=80)))
            if self.args.on_the_fly_feats
            else eval(self.args.input_strategy)(),
            cut_transforms=[ExtraPadding(extra_frames=10)],
            return_cuts=self.args.return_cuts,
        )
```
I get the following error, likely because it is not doing fill_supervisions():
```

/ceph-dan/.local/lib/python3.8/site-packages/lhotse-1.3.0.dev0+git.4198446.clean-py3.8.egg/lhotse/dataset/sampling/dynamic_bucketing.py:121: UserWarning: You are using DynamicBucketingSampler with an eagerly read CutSet. You won\
't see any memory/speed benefits with that setup. Either use 'CutSet.from_jsonl_lazy' to read the CutSet lazily, or use a BucketingSampler instead.
Traceback (most recent call last):
  File "./pruned_transducer_stateless7/decode.py", line 635, in <module>
    main()
  File "/ceph-dan/.local/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 26, in decorate_context
    return func(*args, **kwargs)
  File "./pruned_transducer_stateless7/decode.py", line 617, in main
    results_dict = decode_dataset(
  File "./pruned_transducer_stateless7/decode.py", line 385, in decode_dataset
[...]

PR #816, opened 2022-09-22
Add `pad_silence` option to `extend_by`

desh2608:
This address https://github.com/lhotse-speech/lhotse/issues/779.

Currently, the `extend_by()` method only extends the cut to at most the boundaries of the recording. With this option set to true, we can now pad with silence to extend beyond the recording's boundaries.

### DIFF
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -1375,7 +1375,7 @@ class MonoCut(Cut):
         duration: Seconds,
         direction: str = "both",
         preserve_id: bool = False,
-        pad_silence: bool = False,
+        pad_silence: bool = True,
     ) -> "MonoCut":
         """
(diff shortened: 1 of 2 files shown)


pzelasko (approved):
Great work, thanks!! Do you think it maybe makes sense to enable this option by default? Based on the issue (our single data point about user expectations) it looks like the users might expect that behavior by default.

desh2608:
> Great work, thanks!! Do you think it maybe makes sense to enable this option by default? Based on the issue (our single data point about user expectations) it looks like the users might expect that behavior by default.

Yeah, I suppose that makes sense. I'll make the change.

pzelasko (approved):

pzelasko:
The unit test fail is a random flake, merging. Thanks!

PR #817, opened 2022-09-24
Message when calling len() on LazyFilter

desh2608:
Currently, when `len()` is called on an object of type `LazyFilter` (which is the case when a filter is applied on a lazily loaded manifest), it will throw a `TypeError`:
```python
TypeError: object of type 'LazyFilter' has no len()
```

After this change, we instead show an informative message:
```python
NotImplementedError: LazyFilter does not support __len__ because it would require iterating over the whole iterator, which is not possible in a lazy fashion. If you really need to know the length, convert
 to eager mode first using `.to_eager()`. Note that this will require loading the whole iterator into memory.
 ```

### DIFF
--- a/lhotse/features/base.py
+++ b/lhotse/features/base.py
@@ -18,7 +18,7 @@ from lhotse.audio import Recording
 from lhotse.augmentation import AugmentFn
 from lhotse.features.io import FeaturesWriter, get_reader
 from lhotse.lazy import AlgorithmMixin
-from lhotse.serialization import Serializable, load_yaml, save_to_yaml
+from lhotse.serialization import LazyMixin, Serializable, load_yaml, save_to_yaml
 from lhotse.utils import (
     Pathlike,
     Seconds,
@@ -594,10 +594,12 @@ class FeatureSet(Serializable, AlgorithmMixin):
         return self.features
 
     @staticmethod
-    def from_features(features: Iterable[Features]) -> "FeatureSet":
-        # NOTE: here we use list comprehension instead of list() because the latter
-        # would not work if features is a lazy generator.
-        return FeatureSet([f for f in features])
+    def from_features(features: Union[Iterable[Features], LazyMixin]) -> "FeatureSet":
+        return (
+            FeatureSet([f for f in features])
+            if isinstance(features, LazyMixin)
+            else FeatureSet(list(features))
+        )
 
     from_items = from_features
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lhotse/features/base.py [hunk 1]
```diff
@@ -593,12 +593,13 @@ def data(self) -> Union[Dict[str, Features], Iterable[Features]]:
...
-    def to_eager(self) -> "FeatureSet":
-        return FeatureSet(list(self))
-
     @staticmethod
-    def from_features(features: Iterable[Features]) -> "FeatureSet":
-        return FeatureSet(list(features))  # just for consistency with other *Sets
+    def from_features(features: Union[Iterable[Features], LazyMixin]) -> "FeatureSet":
+        return (
+            FeatureSet([f for f in features])
+            if isinstance(features, LazyMixin)
```
pzelasko, line 600: Curious why not use one of these two options for both cases? I recall list has some weird behavior with regard to whether len is implemented, but maybe we can just always use comprehension?

lhotse/lazy.py
```diff
@@ -363,11 +363,17 @@ def __init__(self, iterator: Iterable, predicate: Callable[[Any], bool]) -> None
     def __iter__(self):
         return filter(self.predicate, self.iterator)
 
-    # note: no __len__
-
     def __add__(self, other) -> "LazyIteratorChain":
         return LazyIteratorChain(self, other)
 
+    def __len__(self) -> int:
+        raise NotImplementedError(
```
pzelasko, line 370: Can you check other lazy iterator classes and see if they can also use this error message? At least one that comes to my mind is LazyFlatten.

pzelasko (approved):
Thanks, great work! It should be very helpful for the users.

lhotse/features/base.py
desh2608: Yeah, `list()` requires `__len__` to be implemented, which is why I had to use list comprehension. However, it seems like list comprehension may be slower than list() (see [this](https://stackoverflow.com/questions/29356846/python-list-vs-list-comprehension-building-speed)), which is why I didn't make it the default.

lhotse/lazy.py
desh2608: Added to `LazyFlattener`. All other lazy iterator classes already have `__len__` defined.

pzelasko (approved):

PR #818, opened 2022-09-24
WIP: `MultiCut` to store multi-channel inputs with shared supervision

desh2608:
This PR will extend supervisions to be shared across channels.


lhotse/utils.py
```diff
@@ -609,6 +609,10 @@ def is_none_or_gt(value, threshold) -> bool:
     return value is None or value > threshold
 
 
+def is_equal_or_contains(value, other) -> bool:
+    return value == other or (isinstance(value, Iterable) and other in value)
```
pzelasko, line 613: I'm sure Iterable is not the right type here, my best guess is Sequence, but you should check the documentation of `typing` to make sure.

```suggestion
    return value == other or (isinstance(value, Sequence) and other in value)
```

pzelasko: would be nice to add types e.g. `value: Any, other: Union[Any, Sequence[Any]]` (or if you feel adventurous, add proper type vars to indicate preserving types across arguments, see TypeVar in `typing` IIRC)

desh2608:
Will open PR from new branch.

PR #820, opened 2022-09-24
Refactor cut and retain `git blame` history

desh2608:
This should hopefully retain all (most?) of the git blame history. The trick is to create a new branch for each split file, and create the new file using `git mv` so that it retains the history of the original file. Once all the branches are created, we merge them together.


pzelasko:
Thanks, amazing work! Please fix the failing style check and LGTM

desh2608:
> Thanks, amazing work! Please fix the failing style check and LGTM

I think adding the imports required to resolve the failing checks would cause a circular dependency. See the dependency graph in `__init__.py`. I am not sure what's a good way to resolve this.

pzelasko:
IMO the easiest way to fix is to change the types from specific (like MixedCut) to generic (like Cut). It's still truthful even if it misses on some auto-hint capabilities if we know that there will be only a specific type returned. I think maybe this problem can be fixed on a higher level by adopting typing stub files (.pyi) but that's probably a considerably larger effort.

desh2608:
Okay, I changed the MonoCut and MixedCut types to Cut. There were still 2 instances of CutSet so I add noqa to ignore them.

desh2608:
It seems the git history was not retained. If this is an issue, we can revert this merge and try to figure out how to do it.

pzelasko:
Don't worry about it. I probably messed up because I squashed then merged.

PR #822, opened 2022-09-26
`MultiCut` to store multi-channel recordings with shared supervision

desh2608:
This PR adds a `MultiCut` which is a new type of Cut that represents multi-channel data. We also define a `DataCut` as a Cut which contains actual data (as opposed to a PaddingCut or MixedCut which are empty or operations on other cuts). This DataCut is the abstract base class from which MonoCut and MultiCut inherit.

```
 >>> 2-channel telephone recording with 2 supervisions, one for each channel (e.g., Switchboard):

                  ╔══════════════════════════════  MultiCut  ═════════════════╗
                  ║ ┌──────────────────────────┐                              ║
     Channel 1  ──╬─│   Hello this is John.    │──────────────────────────────╬────────
                  ║ └──────────────────────────┘                              ║
                  ║                               ┌──────────────────────────┐║
     Channel 2  ──╬───────────────────────────────│ Hey, John. How are you?  │╠────────
                  ║                               └──────────────────────────┘║
                  ╚═══════════════════════════════════════════════════════════╝

   
>>> Multi-array multi-microphone recording with shared supervisions (e.g., CHiME-6),
    along with close-talk microphones (A and B are distant arrays, C is close-talk):

               ╔═══════════════════════════════════════════════════════════════════════════╗
               ║ ┌───────────────────┐                         ┌───────────────────┐       ║
       A-1   ──╬─┤                   ├─────────────────────────┤                   ├───────╬─
               ║ │ What did you do?  │                         │I cleaned my room. │       ║
       A-2   ──╬─┤                   ├─────────────────────────┤                   ├───────╬─
               ║ └───────────────────┘  ┌───────────────────┐  └───────────────────┘       ║
       B-1   ──╬────────────────────────┤Yeah, we were going├──────────────────────────────╬─
               ║                        │   to the mall.    │                              ║
       B-2   ──╬────────────────────────┤                   ├──────────────────────────────╬─
               ║                        └───────────────────┘        ┌───────────────────┐ ║
        C    ──╬─────────────────────────────────────────────────────┤      Right.       ├─╬─
               ║                                                     └───────────────────┘ ║
               ╚════════════════════════════════  MultiCut  ═══════════════════════════════╝
```


lhotse/cut/multi.py
```diff
@@ -0,0 +1,1128 @@
...
+    uuid4,
+)
+
+
+@dataclass
+class MultiCut(Cut):
+    """
+    :class:`~lhotse.cut.MultiCut` is a :class:`~lhotse.cut.Cut` that is analogous to the MonoCut.
+    While MonoCut represents a single channel of a recording, MultiCut represents multiple channels
+    that share the same supervision. It is intended to be used to store, for example, segments
```
pzelasko, line 41: Just skimming through for now, but are you sure you want to the supervisions to always be shared across channels? Can we envision a MultiCut that has different supervisions on each channel?

lhotse/cut/multi.py
```diff
@@ -0,0 +1,1128 @@
...
+    of a microphone array recording.
+
+    By definition, a MultiCut has the same attributes as a MonoCut. The key difference is that
+    the Recording object has multiple channels, and the Supervision object is shared across
+    multiple channels. The channels that the MultiCut pertains to is defined by the channels
+    in the Supervision object, and these are the same as the ``channel`` attribute of the
+    MultiCut. These channels may be a subset of the channels in the Recording object. For
+    example, if the Recording object has 4 channels (0, 1, 2, 3), we may construct a MonoCut
+    object with a supervision for channel 0, and a MultiCut object with a supervision for
+    channels 1 and 2.
```
pzelasko, line 51: I think that MultiCut channels should be interpreted separately from Supervision channels. E.g.:

3-channel MultiCut that consists of
- 4-channel Recording (one channel is not used)
- a 2-channel Supervision (e.g. stereo mics for conversation in XY setting)
- a 1-channel Supervision (e.g. from a different mic)

Does that make sense? It seems (too?) flexible but maybe it's a good opportunity to account for these possibilities if you're doing serious revisions to the code :)

desh2608: Yeah that makes sense. I was coming in more from a utility perspective instead of a representation perspective. However, one aspect we might want to think is how would such a broader definition of MultiCut interact with things like mixing.

desh2608: Perhaps this is what you are proposing?
```
    While MonoCut represents a single channel of a recording, MultiCut represents multi-channel
    recordings where supervisions may or may not be shared across channels. 
    It is intended to be used to store, for example, segments of a microphone array recording.
    The following diagrams illustrate some examples for MultiCut usage:

    >>> 2-channel telephone recording with 2 supervisions, one for each channel (e.g., Switchboard):


                  ╔══════════════════════════════  MultiCut  ═════════════════╗
                  ║ ┌──────────────────────────┐                              ║
     Channel 1  ──╬─│   Hello this is John.    │──────────────────────────────╬────────
                  ║ └──────────────────────────┘                              ║
                  ║                               ┌──────────────────────────┐║
     Channel 2  ──╬───────────────────────────────│ Hey, John. How are you?  │╠────────
                  ║                               └──────────────────────────┘║
                  ╚═══════════════════════════════════════════════════════════╝

    >>> Multi-array multi-microphone recording with shared supervisions (e.g., CHiME-6),
    along with close-talk microphones (A and B are distant arrays, C is close-talk):

               ╔═══════════════════════════════════════════════════════════════════════════╗
               ║ ┌───────────────────┐                         ┌───────────────────┐       ║
       A-1   ──╬─┤                   ├─────────────────────────┤                   ├───────╬─
               ║ │ What did you do?  │                         │I cleaned my room. │       ║
       A-2   ──╬─┤                   ├─────────────────────────┤                   ├───────╬─
               ║ └───────────────────┘  ┌───────────────────┐  └───────────────────┘       ║
       B-1   ──╬────────────────────────┤Yeah, we were going├──────────────────────────────╬─
               ║                        │   to the mall.    │                              ║
       B-2   ──╬────────────────────────┤                   ├──────────────────────────────╬─
               ║                        └───────────────────┘        ┌───────────────────┐ ║
        C    ──╬─────────────────────────────────────────────────────┤      Right.       ├─╬─
               ║                                                     └───────────────────┘ ║
               ╚════════════════════════════════  MultiCut  ═══════════════════════════════╝
```

pzelasko: Now I have to learn how to make these diagrams from you 🤯 yes, this is what I had in mind.

desh2608:
This is a list of additional functionalities that should be added to use the power of MultiCut (in future PRs). I am just putting this here for now so I don't forget.

- [ ] Multi-channel feature extractors (like IPDs, can be based on [this implementation](https://github.com/desh2608/css/blob/main/css/executor/feature.py))
- [ ] MVDR workflow using torchaudio ([link](https://pytorch.org/audio/0.10.0/tutorials/mvdr_tutorial.html))
- [ ] Multi-channel ASR dataset

lhotse/cut/mixed.py
```diff
@@ -38,31 +39,36 @@
...
     offset: Seconds = 0.0
     snr: Optional[Decibels] = None
 
+    def __post_init__(self):
+        self.type = type(self.cut).__name__
+
     @staticmethod
     def from_dict(data: dict):
         raw_cut = data.pop("cut")
+        raw_cut_type = eval(data.pop("type"))
```
pzelasko, line 57: in `CutSet.from_dicts` there's an close called `deserialize_one` that can disambiguate which type of cut to load. Can you promote it to a standalone function (or create `Cut.from_dict()` with that so that it just instantiates the right child class)? I would rather avoid an `eval()` here because what if somebody sends you a manifest with `MixTrack(type="import os; os.system('sudo rm -rf /')")`? 😀

lhotse/cut/mixed.py
```diff
@@ -147,6 +159,11 @@ def num_samples(self) -> Optional[int]:
     def num_features(self) -> Optional[int]:
         return self.tracks[0].cut.num_features
 
+    @property
+    def num_channels(self) -> Optional[int]:
+        # PaddingCut has 0 channels, MonoCut has 1, MultiCut has N. We don't support MixedCut with MixedCut tracks.
```
pzelasko, line 164: shouldn't PaddingCut have a single channel too?

lhotse/cut/mono.py
```diff
@@ -198,6 +198,11 @@ def num_samples(self) -> Optional[int]:
     def num_features(self) -> Optional[int]:
         return self.features.num_features if self.has_features else None
 
+    @property
+    def num_channels(self) -> Optional[int]:
+        # MonoCut is always single-channel, if it has a recording.
+        return int(self.has_recording)
```
pzelasko, line 204: By definition MonoCut has a single channel regardless of the recording's presence, right?

```suggestion
        return 1
```

lhotse/cut/mixed.py
desh2608: Yeah will do.

lhotse/cut/mono.py
desh2608: But then what does this "ghost" channel refer to?

lhotse/cut/padding.py
```diff
@@ -70,6 +70,10 @@ def has_features(self) -> bool:
     def has_recording(self) -> bool:
         return self.num_samples is not None
 
+    @property
+    def num_channels(self) -> int:
+        return 0
```
pzelasko, line 75: I think 1 is technically correct, it's all zeros but it's a channel.

```suggestion
        return 1
```

lhotse/cut/multi.py
```diff
@@ -0,0 +1,1155 @@
+import logging
```
pzelasko, line 1: There is a lot of duplicated code in this file, what are the key differences between Mono and Multi cuts? If there's not too many, maybe there is a way to make them share common codebase? In the current shape, the codebase will be a bit more difficult to maintain.

lhotse/cut/multi.py
```diff
@@ -0,0 +1,1155 @@
...
+    @property
+    def sampling_rate(self) -> int:
+        return (
+            self.features.sampling_rate
+            if self.has_features
+            else self.recording.sampling_rate
+        )
+
+    @rich_exception_info
+    def load_features(self, channel: Optional[int] = None) -> Optional[np.ndarray]:
```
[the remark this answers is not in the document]
desh2608: I think they can be either 2D or 3D. But even 2D features can refer to multiple channels.

lhotse/cut/multi.py
desh2608: My original plan was to inherit MultiCut from MonoCut and only implement the methods where they differ. The best way would probably be to inherit both from a common base class though.

pzelasko: I think a common base class does make sense here (even though I tend to be cautious around object hierarchies with more than 2 levels). Not sure how to name it though. The key common property is that both Mono and Multi cuts refer to specific data rather than to an operation on other cuts (like MixedCut does). `DataCut`? `OriginalCut`? `SourceCut`? Maybe you can think of something better..

lhotse/cut/mono.py
pzelasko: If you have features instead, then they represent that single channel. If you have neither recording or features, then the object is sort of invalid anyway.

lhotse/cut/multi.py
desh2608: I think `DataCut` seems most intuitive.

lhotse/cut/multi.py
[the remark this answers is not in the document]
pzelasko: Yeah, that's what I was suspecting. E.g. if you reduce to one channel via beamforming and then compute the features. Probably worth supporting both cases here. Actually it seems it would have been easier to represent those things using `TemporalArray` rather than `Features`: I was vaguely intending to replace `Features` everywhere with temporal arrays, but I never actually got to it... it's also an extra effort to preserve backwards compatibility with manifests created before such change. Maybe in the future..

lhotse/cut/mono.py
desh2608: Ahh, makes sense.

lhotse/cut/multi.py
desh2608: Here's what I'm thinking. We can have 2 base classes: DataCut and DerivedCut. DataCut refers to cuts that point to actual data, while DerivedCut refers to an operation on other cuts. MonoCut and MultiCut will inherit from DataCut, and MixedCut will inherit from DerivedCut. I'm not sure if it's still useful to retain the original Cut, though.

desh2608: That might be too complicated, however. Perhaps best to just create a base class for Mono and Multi, for now.

pzelasko: > That might be too complicated, however. Perhaps best to just create a base class for Mono and Multi, for now.

Yes, the other base class serves no real purpose. We want to have a common base class even if just for easier type annotations.

desh2608: Done.

lhotse/cut/set.py
```diff
@@ -2776,6 +2753,24 @@ def merge(values: Iterable[str]) -> Optional[str]:
...
+    if cut_type == "Cut":
+        warnings.warn(
+            "Your manifest was created with Lhotse version earlier than v0.8, when MonoCut was called Cut. "
+            "Please re-generate it with Lhotse v0.8 as it might stop working in a future version "
+            "(using manifest.from_file() and then manifest.to_file() should be sufficient)."
+        )
+        return MonoCut.from_dict(raw_cut)
+    if cut_type == "MixedCut":
+        return MixedCut.from_dict(raw_cut)
+    raise ValueError(f"Unexpected cut type during deserialization: '{cut_type}'")
```
pzelasko, line 2771: What about PaddingCut?

lhotse/cut/multi.py
```diff
@@ -84,172 +73,10 @@ class MultiCut(Cut):
...
-            else None
-        )
-
-    @property
-    def num_features(self) -> Optional[int]:
-        return self.features.num_features if self.has_features else None
-
     @property
     def num_channels(self) -> Optional[int]:
         return len(self.channel) if self.has_recording else None
```
pzelasko, line 78: This should always be `len(self.channel)`, similarly like it's always `1` in MonoCut. Even if you drop the recording, the precomputed features represent these channels.

desh2608:
I think this one's ready for review once the tests pass.

csukuangfj:
```
 >>> 2-channel telephone recording with 2 supervisions, one for each channel (e.g., Switchboard):

                  ╔══════════════════════════════  MultiCut  ═════════════════╗
                  ║ ┌──────────────────────────┐                              ║
     Channel 1  ──╬─│   Hello this is John.    │──────────────────────────────╬────────
                  ║ └──────────────────────────┘                              ║
                  ║                               ┌──────────────────────────┐║
     Channel 2  ──╬───────────────────────────────│ Hey, John. How are you?  │╠────────
                  ║                               └──────────────────────────┘║
                  ╚═══════════════════════════════════════════════════════════╝

```

The visualization looks great!

lhotse/audio.py
```diff
@@ -1215,6 +1215,7 @@ def __init__(
         self.tracks = [base_audio]
         self.offsets = [0]
         self.sampling_rate = sampling_rate
+        self.num_channels = base_audio.shape[0]
```
pzelasko, line 1218: Can you update the documentation of `AudioMixer` to explain the new multi channel behavior? E.g. for K channel base_audio, do you require every mixed track to have K channels as well, or can they also have 1 channel, etc?

lhotse/audio.py
```diff
@@ -1264,12 +1265,38 @@ def unmixed_audio(self) -> List[np.ndarray]:
...
+            elif track.shape[0] > 1 and self.num_channels == 1:
+                # Sum all channels of the track
+                track = np.sum(track, axis=0, keepdims=True)
+            else:
+                raise ValueError("Incompatible number of channels")
+            mixed[:, offset : offset + track.shape[1]] += track
+        return mixed
+
+    @property
+    def mixed_and_flattened_audio(self) -> np.ndarray:
```
pzelasko, line 1289: Do I understand correctly this could be just named `mixed_mono_audio`?

lhotse/cut/__init__.py
```diff
@@ -7,28 +7,32 @@
 
 """
```
pzelasko, line 8: This docstring should be moved to the top of the file (I think then it becomes a module-level docstring). Also--great visualisation!

lhotse/cut/mixed.py
```diff
@@ -704,8 +710,12 @@ def load_features(self, mixed: bool = True) -> Optional[np.ndarray]:
...
         Loads the features of the source cuts and mixes them on-the-fly.
 
-        :param mixed: when True (default), returns a 2D array of features mixed in the feature domain.
-            Otherwise returns a 3D array with the first dimension equal to the number of tracks.
+        :param mixed: when True (default), the features are mixed together (as defined in
+            the mixing function for the extractor). This could result in either a 2D or 3D
+            array. For example, if all underlying tracks are single-channel, the output
+            will be a 2D array of shape (num_frames, num_features). If any of the tracks
+            are multi-channel, the output may be a 3D array of shape (num_frames, num_features,
+            num_channels).
```
pzelasko, line 718: any specific reason to go with channels last format?

lhotse/cut/mixed.py
```diff
@@ -765,7 +786,7 @@ def load_features(self, mixed: bool = True) -> Optional[np.ndarray]:
                 )
             except NonPositiveEnergyError as e:
                 logging.warning(
-                    str(e) + f' MonoCut with id "{track.cut.id}" will not be mixed in.'
+                    str(e) + f' DataCut with id "{track.cut.id}" will not be mixed in.'
```
pzelasko, line 789: how about sth like `type(track.cut).__name__`?

lhotse/cut/mixed.py
```diff
@@ -846,13 +877,26 @@ def load_audio(self, mixed: bool = True) -> Optional[np.ndarray]:
                 )
             except NonPositiveEnergyError as e:
                 logging.warning(
-                    f'{e} MonoCut with id "{track.cut.id}" will not be mixed in.'
+                    f'{e} DataCut with id "{track.cut.id}" will not be mixed in.'
```
pzelasko, line 880: how about sth like type(track.cut).__name__?

lhotse/cut/mixed.py
```diff
@@ -795,12 +816,22 @@ def load_features(self, mixed: bool = True) -> Optional[np.ndarray]:
...
-        :param mixed: When True (default), returns a mono mix of the underlying tracks.
-            Otherwise returns a numpy array with the number of channels equal to the number of tracks.
+        :param mixed: When True (default), returns a mix of the underlying tracks. This will
+            return a numpy array with shape ``(num_channels, num_samples)``, where ``num_channels``
+            is determined by the ``num_channels`` property of the MixedCut. Otherwise returns a
+            numpy array with the number of channels equal to the total number of channels
+            across all tracks in the MixedCut. For example, if it contains a MultiCut with 2
+            channels and a MonoCut with 1 channel, the returned array will have shape
+            ``(3, num_samples)``.
+        :param flattened: If the MixedCut contains > 1 channels (for e.g. when one of its tracks
```
pzelasko, line 832: can we call it `mono_downmix` instead?

lhotse/cut/mixed.py
```diff
@@ -846,13 +877,26 @@ def load_audio(self, mixed: bool = True) -> Optional[np.ndarray]:
                 )
             except NonPositiveEnergyError as e:
                 logging.warning(
-                    f'{e} MonoCut with id "{track.cut.id}" will not be mixed in.'
+                    f'{e} DataCut with id "{track.cut.id}" will not be mixed in.'
                 )
 
+        if flattened and not any(track.type == "MultiCut" for track in self.tracks):
+            warnings.warn(
```
pzelasko, line 884: I don't think that's a serious enough issue to annoy the user with a warning (a lesson learned the hard way for me)

lhotse/cut/mixed.py
```diff
@@ -846,13 +877,26 @@ def load_audio(self, mixed: bool = True) -> Optional[np.ndarray]:
...
+                "Flattening a MixedCut without MultiCut tracks has no effect.",
+                UserWarning,
+            )
+            flattened = False
+
+        if flattened and not mixed:
+            warnings.warn(
```
pzelasko, line 891: I would say same for this warning, we can just mention it in documentation.

lhotse/cut/multi.py
```diff
@@ -73,16 +72,21 @@ class MultiCut(DataCut):
         - :class:`lhotse.cut.MixedCut`
     """
 
+    channel: List[int]
```
pzelasko, line 75: if it's a list, should it be plural (`channels`) instead?

lhotse/cut/multi.py
```diff
@@ -347,6 +354,28 @@ def from_mono(*cuts: Cut) -> "MultiCut":
             }
         )
 
+    def to_mono(self) -> List["DataCut"]:
```
pzelasko, line 357:
```suggestion
    def to_mono(self) -> List["MonoCut"]:
```

lhotse/cut/set.py
```diff
@@ -2638,127 +2658,14 @@ def _takewhile(
     return collected, iterable
 
 
-def merge_supervisions(
```
pzelasko, line 2589: Do I get it right -- you moved `merge_supervisions` to each of the cut classes definition because it's cleaner to implement them separately. In that case, can you keep `merge_supervisions` for CutSet that simply returns `self.map(lambda cut: cut.merge_supervisions(...))`?

lhotse/audio.py
```diff
@@ -1264,12 +1265,38 @@ def unmixed_audio(self) -> List[np.ndarray]:
...
+        total = self.num_samples_total
+        mixed = np.zeros((self.num_channels, total), dtype=self.dtype)
+        for offset, track in zip(self.offsets, self.tracks):
+            if track.shape[0] == self.num_channels:
+                # Do nothing
+                ...
+            elif track.shape[0] == 1 and self.num_channels > 1:
+                # Duplicate the track for all channels
+                track = np.tile(track, (self.num_channels, 1))
+            elif track.shape[0] > 1 and self.num_channels == 1:
```
pzelasko, line 1280: is it possible to improve the test coverage of AudioMixer (see codecov comments here)? Historically, it's a class that required many fixes over time.

lhotse/cut/data.py
```diff
@@ -0,0 +1,960 @@
...
+    overlaps,
+    overspans,
+    perturb_num_samples,
+    rich_exception_info,
+    uuid4,
+)
+
+
+@dataclass
+class DataCut(Cut, metaclass=ABCMeta):
```
pzelasko, line 38: I don't know how you made abstract classes work with dataclasses and inheritance, but you seem to have achieved something I could not :) kudos

lhotse/cut/multi.py
```diff
@@ -0,0 +1,405 @@
...
+                    fastcopy(s, channel=channel)
+                    for s in self.supervisions
+                    if channel in s.channel
+                ],
+            )
+            for channel in self.channel
+        ]
+
+    @staticmethod
+    def from_dict(data: dict) -> "MultiCut":
```
pzelasko, line 380: It looks like serialization of MultiCut is missing test coverage, can you pls add some tests for this?

pzelasko (commented):
Wow, very impressive work!! I left just a few comments, mostly minor ones + asking for a bit more test coverage, otherwise it looks great!

lhotse/audio.py
desh2608: Yes, that is correct. I will rename it.

lhotse/cut/mixed.py
desh2608: Mainly because I could use ellipsis in the tensor indexing and then the same code would work for both 2D and 3D features.

lhotse/cut/multi.py
desh2608: This would cause a flake8 issue since we cannot import MonoCut (to avoid circular import).

lhotse/cut/set.py
desh2608: Yeah, that's right. Will do.

desh2608: Actually, the `CutSet` already has merge_supervisions(). I only removed the global utility function which implemented the merging logic and moved the implementation to each Cut type.

lhotse/cut/data.py
desh2608: I think a bit of help from Github copilot XD

lhotse/cut/multi.py
desh2608: Done.

lhotse/audio.py
desh2608: Done.

desh2608:
@pzelasko Addressed all comments.

pzelasko:
LGTM! Again, great work.

PR #824, opened 2022-09-27
Experimental Lhotse feature: corpus creation tools (``workflows``), starting with OpenAI Whisper support

pzelasko:
It's actually a very experimental feature, but people seem to be getting satisfying results with running Whisper on podcasts, videos, their own speech, etc. So maybe it can be used for weak labeling of unlabeled datasets. I added a "dummy-friendly" interface in Lhotse that can create full CutSet manifest with multiple supervisions per cut from just a directory of recordings. It doesn't try to be super efficient (no dataloading, batching, etc.). 

Usage:

```bash
lhotse workflows annotate-with-whisper -r my_recordings -e flac cuts.jsonl.gz 
```

The results are saved to `cuts.jsonl.gz` (can be also `cuts.jsonl`).

### DIFF
--- a/lhotse/workflows/whisper.py
+++ b/lhotse/workflows/whisper.py
@@ -1,5 +1,4 @@
 import logging
-from itertools import pairwise
 from typing import Generator, List, Optional
 
 import torch
@@ -77,10 +76,12 @@ def _postprocess_timestamps(supervisions: List[SupervisionSegment]):
     Under a strong assumption that the input speech is non-overlapping, we can fix that
     by always truncating to the start timestamp of the next segment.
     """
+    from cytoolz import sliding_window
+
     if len(supervisions) < 2:
         return supervisions
     out = []
-    for cur, nxt in pairwise(supervisions):
+    for cur, nxt in sliding_window(2, supervisions):
         if cur.end > nxt.start:
             cur = cur.trim(end=nxt.start)
         out.append(cur)
--- a/lhotse/workflows/whisper.py
+++ b/lhotse/workflows/whisper.py
@@ -1,0 +1,88 @@
+import logging
+from itertools import pairwise
+from typing import Generator, List, Optional
+
+import torch
+
+from lhotse import MonoCut, RecordingSet, SupervisionSegment
+from lhotse.qa import trim_supervisions_to_recordings
+from lhotse.utils import is_module_available
+
+
+def annotate_with_whisper(
+    recordings: RecordingSet,
+    language: Optional[str] = None,
+    model_name: str = "base",
+    device: str = "cpu",
+) -> Generator[MonoCut, None, None]:
+    """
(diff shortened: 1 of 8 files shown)


mthrok:
Question out of curiosity: Is alignment part of this?

pzelasko:
> Question out of curiosity: Is alignment part of this?

Not right now, but after some quick skimming I think that info is available inside ``whisper.transcribe``, so that could be added too.

danpovey:
Cool!

desh2608:
Perhaps we can also add a function that takes a CutSet as input instead of a RecordingSet, and modifies the supervisions. I am thinking of cases when we have long recordings with segments (e.g., obtained from a VAD), and we want to fill in the `text` attribute for the supervisions.

desh2608:
> > Perhaps we can also add a function that takes a CutSet as input instead of a RecordingSet, and modifies the supervisions. I am thinking of cases when we have long recordings with segments (e.g., obtained from a VAD), and we want to fill in the `text` attribute for the supervisions.
> 
> Makes sense. If it's something that will help you feel free to take it, my main aim for now is to find some time to dig into that word-level alignment thing, a lot of people seem interested in that.

Sure, will create a PR.

pzelasko:
OK I mis-read the Whisper paper, they only provide segment-level start/end timestamps, I thought they have word-level timestamps too. I only figured it out while spending more time with the internals. Still, it would be very cool to provide a robust out-of-the-box word-level forced aligner for data annotation. Maybe some Icefall models can be used for that? @danpovey @csukuangfj what do you think is the best approach here?

csukuangfj:
We have force-alignment code in icefall with RNN-T.
https://github.com/k2-fsa/icefall/pull/239

But the results may not be as accurate as the one obtained using HMM-GMM.

---

I am not sure whether it is worthwhile to extract the HMM-GMM part from kaldi as an external project, rewrite it with libtorch , and wrap it to python.

huangruizhe:
Yes, but it is still ongoing.

lhotse/workflows/whisper.py
```diff
@@ -0,0 +1,89 @@
...
+                language=result["language"],
+            )
+            for segment in result["segments"]
+        ]
+        cut = recording.to_cut()
+        if supervisions:
+            supervisions = _postprocess_timestamps(supervisions)
+            cut.supervisions = list(
+                trim_supervisions_to_recordings(
+                    recordings=recordings, supervisions=supervisions, verbose=False
```
desh2608, line 67: Shouldn't this be `recordings=recording` instead of `recordings=recordings`?

desh2608: I guess it would still work since internally it checks only the corresponding recording_id, but we are passing around the full manifest for no reason.

pzelasko: Good point!

PR #827, opened 2022-09-28
[workflow] Word-level forced alignment with pretrained models from Torchaudio

pzelasko:
As a follow-up to #824 I'm adding another workflow for forced word-level alignment. It may be interesting to combine the two workflows together. In the end I used the torchaudio model because the tutorial example looks quite convincing with regard to word timestampt accuracy, but I did not evaluate if it's better or worse than any other approach. We can add other workflows for forced alignment later to create some choice.

@desh2608 I also refactored the AlignmentItem thing a bit so that it's NamedTuple again, for the sake of efficiency. I don't like this approach though, after using it for some time. I think we should eventually change it to something like a numpy array (or a collection of them) and maybe not store them in the manifest. Anyway, at least the size of the manifests is greatly reduced with the NamedTuple approach.

### DIFF
--- a/lhotse/bin/modes/workflows.py
+++ b/lhotse/bin/modes/workflows.py
@@ -104,7 +104,7 @@ def annotate_with_whisper(
     "-n",
     "--bundle-name",
     default="WAV2VEC2_ASR_BASE_960H",
-    help="One of Torchaudio pretrained 'bundle' variants (see: https://pytorch.org/audio/stable/pipelines.html)",
+    help="One of torchaudio pretrained 'bundle' variants (see: https://pytorch.org/audio/stable/pipelines.html)",
 )
 @click.option(
     "-d", "--device", default="cpu", help="Device on which to run the inference."
--- a/lhotse/bin/modes/workflows.py
+++ b/lhotse/bin/modes/workflows.py
@@ -6,6 +6,7 @@ from tqdm import tqdm
 
 from lhotse import CutSet, RecordingSet
 from lhotse.bin.modes.cli_base import cli
+from lhotse.serialization import load_manifest_lazy_or_eager
 from lhotse.utils import exactly_one_not_null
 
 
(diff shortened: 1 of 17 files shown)


desh2608 (dismissed):
Super-fast development on this one! LGTM, in general.

Regarding alignments, I agree it's not very efficient. Perhaps treating them in the same way as Features would be a better option.

desh2608:
Here's one approach off the top of my head. We can have an `AlignmentObject` that contains 2 things:
1. A dict which is just a mapping from a symbol (e.g., word, phone) to an int.
2. A path to a numpy file, containing rows of the form (symbol_idx, start_time, end_time, score).

lhotse/bin/modes/workflows.py
```diff
@@ -92,3 +93,61 @@ def annotate_with_whisper(
...
+@workflows.command()
+@click.argument(
+    "in_cuts", type=click.Path(exists=True, dir_okay=False, allow_dash=True)
+)
+@click.argument("out_cuts", type=click.Path(allow_dash=True))
+@click.option(
+    "-n",
+    "--bundle-name",
+    default="WAV2VEC2_ASR_BASE_960H",
+    help="One of Torchaudio pretrained 'bundle' variants (see: https://pytorch.org/audio/stable/pipelines.html)",
```
csukuangfj, line 107:
```suggestion
    help="One of torchaudio pretrained 'bundle' variants (see: https://pytorch.org/audio/stable/pipelines.html)",
```

just to be consistent with other usages of `torchaudio` in this file.

lhotse/supervision.py
```diff
@@ -24,22 +34,27 @@
...
-    We use dataclasses instead of namedtuples (even though they are potentially slower)
-    because of a serialization bug in nested namedtuples and dataclasses in Python 3.7
-    (see this: https://alexdelorenzo.dev/programming/2018/08/09/bug-in-dataclass.html).
-    We can revert to namedtuples if we bump up the Python requirement to 3.8+.
     """
 
     symbol: str
     start: Seconds
     duration: Seconds
+    score: Optional[float] = None
```
csukuangfj, line 47: What does the score indicate?

pzelasko: It's aligner-specific "confidence score", in case of torchaudio they suggested to provide the average token probability for a given segment in their tutorial; but it could be also computed differently. I'll add a proper note.

PR #829, opened 2022-09-29
Drop support for Python 3.6

### LINKED ISSUE #828: [feedback requested] Dropping support for Python 3.6
> Python 3.6.0
> Release Date: Dec. 23, 2016
> Note: The release you are looking at is Python 3.6.0, the initial feature release for the legacy 3.6 series which has now reached end-of-life and is no longer supported. See the [downloads page](https://www.python.org/downloads/) for currently supported versions of Python. The final source-only security fix release for 3.6 was [3.6.15](https://www.python.org/downloads/release/python-3615/) and the final bugfix release was [3.6.8](https://www.python.org/downloads/release/python-368/).

Source: https://www.python.org/downloads/release/python-360/

It would make my life a bit simpler (and less work in CI) if we drop the support for Python 3.6 in Lhotse 1.8 and onwards. Before doing that, I'd like to see if anybody in the community really needs Python 3.6 to be supported for some reason.

pzelasko:
Closes #828

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -16,8 +16,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - python-version: "3.6"
-            torch-install-cmd: "pip install torch==1.8.2+cpu torchaudio==0.8.2 -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html"
           - python-version: "3.7"
             torch-install-cmd: "pip install torch==1.8.2+cpu torchaudio==0.8.2 -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html"
           - python-version: "3.8"
--- a/setup.py
+++ b/setup.py
@@ -202,7 +202,7 @@ if os.environ.get("READTHEDOCS", False):
 setup(
     name="lhotse",
     version=LHOTSE_VERSION,
-    python_requires=">=3.6.0",
+    python_requires=">=3.7.0",
     description="Data preparation for speech processing models training.",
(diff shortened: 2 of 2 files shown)


csukuangfj:
Shall we display an explicit message saying that Python 3.6 is no longer supported?

In sherpa, we use the following code

https://github.com/k2-fsa/sherpa/blob/master/setup.py#L44

```python3
if sys.version_info < (3,):
    # fmt: off
    print(
        "Python 2 has reached end-of-life and is no longer supported by sherpa."
    )
    # fmt: on
    sys.exit(-1)

if sys.version_info < (3, 7):
    print(
        "Python 3.6 has reached end-of-life on December 31st, 2021 "
        "and is no longer supported by sherpa."
    )
    sys.exit(-1)

```

csukuangfj:
In PyTorch, they use the following code
https://github.com/pytorch/pytorch/blob/master/setup.py#L205

```python3
import sys
if sys.version_info < (3,):
    print("Python 2 has reached end-of-life and is no longer supported by PyTorch.")
    sys.exit(-1)
if sys.platform == 'win32' and sys.maxsize.bit_length() == 31:
    print("32-bit Windows Python runtime is not supported. Please switch to 64-bit Python.")
    sys.exit(-1)

import platform
python_min_version = (3, 7, 0)
python_min_version_str = '.'.join(map(str, python_min_version))
if sys.version_info < python_min_version:
    print("You are using Python {}. Python >={} is required.".format(platform.python_version(),
                                                                     python_min_version_str))
    sys.exit(-1)
```

PR #831, opened 2022-09-30
create proper wav_id in the segments file for multichannel recording

jtrmal:
found a bug... Might be the same thing Nagendra found, I'm not sure ATM. But this is definitely something that is fixable without breaking anything that worked previously (this produced just plainly wrong segment file)


jtrmal:
I think I wanna fix this before we dabble into refactoring the export

pzelasko (approved):

pzelasko:
Thanks!

PR #832, opened 2022-09-30
kaldi: add an switch/option to read the durations from kaldi utt2dur …

### LINKED ISSUE #788: Support reading from `reco2dur` in `lhotse.kaldi.load_kaldi_data_dir`
For now in `lhotse.kaldi.load_kaldi_data_dir`, it reads the original audio file, and get the duration info. but this is not friendly to long recodings with `sox` or `ffmpeg` conversion, when `reco2dur` already exists in the kaldi data directory.

Will you support loading `reco2dur` and get duration information from this file?

jtrmal:
…instead of touching the audio file
should resolve #788
also, let me know if -d is the option that feels right for it, to me it does not exactly do, but didn't feel another is better


jtrmal:
I tested:
```
 lhotse kaldi import -d   . 16000 kaldi2/
 ```
 and compared the duration with original files. They weren't equal to the durations in utt2dur but I assume it's just float representation issue

**Comment by jtrmal** [comment]
😱

File: lhotse/kaldi.py [hunk 1]
```diff
@@ -371,18 +381,20 @@ def make_wavscp_channel_string_map(
             # used in the sph files
             audios = dict()
             for channel in source.channels:
-                audios[
-                    channel
-                ] = f"sph2pipe {source.source} -f wav -c {channel+1} -p | ffmpeg -threads 1 -i pipe:0 -ar {sampling_rate} -f wav -threads 1 pipe:1 |"
+                audios[channel] = (
+                    f"sph2pipe {source.source} -f wav -c {channel+1} -p | ffmpeg -threads 1 "
+                    "-i pipe:0 -ar {sampling_rate} -f wav -threads 1 pipe:1 |"
```
pzelasko, line 386:
```suggestion
                    f"-i pipe:0 -ar {sampling_rate} -f wav -threads 1 pipe:1 |"
```

lhotse/kaldi.py
```diff
@@ -371,18 +381,20 @@ def make_wavscp_channel_string_map(
...
         else:
             # Handles non-WAVE audio formats and multi-channel WAVEs.
             audios = dict()
             for channel in source.channels:
-                audios[
-                    channel
-                ] = f"ffmpeg -threads 1 -i {source.source} -ar {sampling_rate} -map_channel 0.0.{channel}  -f wav -threads 1 pipe:1 |"
+                audios[channel] = (
+                    f"ffmpeg -threads 1 -i {source.source} -ar {sampling_rate} "
+                    "-map_channel 0.0.{channel}  -f wav -threads 1 pipe:1 |"
```
pzelasko, line 396:
```suggestion
                    f"-map_channel 0.0.{channel}  -f wav -threads 1 pipe:1 |"
```

pzelasko (commented):
I found some issues, could you add a unit test to cover those? The existing tests didn't catch them.

Regarding reco2dur, IIRC it only represents durations up to two decimal points so the tolerance is +/- 5ms. [Default Lhotse tolerance for mismatch is +/-0.25s](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/audio.py#L70) (which makes Lhotse happy with how most MP3 decoders are diverging in terms of the total number of samples), so it should be OK to set this option to true as a default as save everybody's time. Historically this code used reco2dur already, but Lhotse tolerated zero mismatch between manifest and audio duration at the time, so I removed it (and much later realized there is no way that approach can work with various audio codecs and added the tolerance).

jtrmal:
yeah, sorry, didn't realize these...
working on the unit test for the new feature (ETA 5 mins)
y.

jtrmal:
Added tests for my incorrect formatting and added test for the new feature. Lets wait. But tests pass fine on my setup
y.

pzelasko:
Can you also change to use reco2dur by default, if present? It should speed up things and shouldn't hurt anybody.

jtrmal:
the default behavior is to ignore the reco2dur existence and read the files
fully, as it has been done before, only when the import is called with the
proper parameter, the durations are read from that file.
y.

jtrmal:
our emails just crossed...
ok, if you are fine with it, happily.

Just FYI -- it also implicitly tests for the existence of the file (because
it will complain if the file does not exist0, so it might cause some
confusion
y.

jtrmal:
OK, I'm done with the changes (including making the option enabled by default). 
Just waiting for the tests

jtrmal:
I'm investigating. I assume its something with my cast/converstion to float

jtrmal:
OK, should be fixed now. It was just an issue in the test, due to change of defaults (I was much stricter than I should have been)

jtrmal:
OK, we survived. Ready for review :)

test/test_kaldi_dirs.py
```diff
@@ -220,6 +221,7 @@ def test_ok_on_file_singlechannel_sph_source_type(tmp_path, channel):
     assert list(out.keys()) == [channel]
     assert out[channel].startswith("sph2pipe")
     assert "nonexistent.sph" in out[channel]
+    assert "{" not in out[channel]
```
desh2608, line 224: Just curious to know what's this `assert` verifying?

lhotse/bin/modes/kaldi.py
```diff
@@ -42,7 +42,7 @@ def kaldi():
 @click.option(
     "-d",
     "--use-reco2dur",
```
pzelasko, line 44: I think with default=True and is_flag=True it is impossible to turn this off. The idiomatic click way would be `@click.option("--use-reco2dur/--compute-durations", default=True, help="...")` and then name the Python function argument `use_reco2dur: bool` (the same as the first option before slash, i.e. the same as right now).

lhotse/kaldi.py
```diff
@@ -97,8 +107,10 @@ def fix_id(t: str) -> str:
                 )
             ],
             sampling_rate=sampling_rate,
-            num_samples=compute_num_samples(durations[recording_id], sampling_rate),
-            duration=durations[recording_id],
+            num_samples=compute_num_samples(
+                float(durations[recording_id]), sampling_rate
```
pzelasko, line 111: I think it could be cleaner to do sth like `load_kaldi_text_mapping(path, float_vals: bool = False)` and when `float_vals=True`, cast the second column to float in the function. Otherwise you have to cast in multiple places, looks like a nasty surprise for anybody modifying these things in the future :)

jtrmal:
just to detect improperly formatted strings  (I split f"" string into to
and forgot the second string prefix with "f"). So it will try to detect
left-over "{placeholders}". It's only in a test so I didn't do it more
extensive/foolproof.
y.

jtrmal:
good idea, will fix

test/test_kaldi_dirs.py
jtrmal: it's just detect someone has goofed up and didn't fully formatted the f-string (or forgot to add f"")

jtrmal:
@pzelasko  your comments are addressed. I don't understand the 3.7 unit tests failing reason

jtrmal:
could this be an issue? https://stackoverflow.com/questions/73929564/entrypoints-object-has-no-attribute-get-digital-ocean

pzelasko:
> could this be an issue? https://stackoverflow.com/questions/73929564/entrypoints-object-has-no-attribute-get-digital-ocean

🤦🏻‍♂️ 

do you mind adding `importlib-metadata<5.0.0` to requirements and checking if it helps? Thanks...

jtrmal:
hm, that wasn't it :(
the last errors come from hypothesis, but I'm not sure if thats the
culprit :(
y.

setup.py
```diff
@@ -141,6 +141,7 @@ def mark_lhotse_version(version: str) -> None:
 
 
 install_requires = [
+    "importlib-metadata<5.0.0",
```
pzelasko, line 144: if these lines don't help, let's remove them

setup.py
```diff
@@ -180,6 +181,7 @@ def mark_lhotse_version(version: str) -> None:
 
 docs_require = (project_root / "docs" / "requirements.txt").read_text().splitlines()
 tests_require = [
+    "importlib-metadata<5.0.0",
```
pzelasko, line 184: if these lines don't help, let's remove them

pzelasko (commented):
LGTM, just one minor comment. Don't worry about failing tests, it'll either sort itself out (pypi/packaging flake for some third party dependency) or I will look into it later.

jtrmal:
will remove. Thanks!
y.

jtrmal:
ah, sorry, fixing this

PR #834, opened 2022-10-03
Use CutSet for whisper annotation workflow

desh2608:
This PR allows transcribing a CutSet with the whisper workflow. It is currently WIP since it relies on https://github.com/lhotse-speech/lhotse/pull/832 and https://github.com/lhotse-speech/lhotse/pull/822. But I have verified that it works.


lhotse/workflows/whisper.py
```diff
@@ -39,32 +40,35 @@ def annotate_with_whisper(
...
         ]
-        cut = recording.to_cut()
+        cut = item.to_cut() if isinstance(item, Recording) else item
+        recording = item if isinstance(item, Recording) else item.recording
         if supervisions:
             supervisions = _postprocess_timestamps(supervisions)
             cut.supervisions = list(
                 trim_supervisions_to_recordings(
-                    recordings=recordings, supervisions=supervisions, verbose=False
+                    recordings=recording, supervisions=supervisions, verbose=False
```
pzelasko, line 71: This won't do unfortunately. If you have 1h Recording with 10s cuts, the segments can go beyond cut's end (which doesn't make sense).

Unfortunately I think it might be cleaner to write a separate variant of this function for cuts... I don't like how the control flow became quite complex.

desh2608:
@pzelasko I have separated it into two helper functions. Let me know if it looks reasonable now.

lhotse/workflows/whisper.py
```diff
@@ -64,12 +79,44 @@ def annotate_with_whisper(
...
+            )
+            continue
+        audio = torch.from_numpy(cut.resample(16000).load_audio()).squeeze(0)
+        result = whisper.transcribe(model=model, audio=audio, language=language)
+        supervisions = [
+            SupervisionSegment(
+                id=f"{cut.id}-{segment['id']:06d}",
+                recording_id=cut.recording_id,
+                start=round(segment["start"], ndigits=8),
+                duration=round(segment["end"], ndigits=8),
```
pzelasko, line 110: I think with this single change we can prevent timestamps going out of cut's duration; can you also include `_postprocess_timestamps` in the cut-based workflow?

```suggestion
                duration=max(cut.duration, round(segment["end"], ndigits=8)),
```

desh2608: Done!

pzelasko: Because in my own testing I discovered that their "end" is actually a "duration" 😂 but if you could triple-check that I got it right, maybe using some longer recording, that'd be great.

pzelasko:
Thanks, LGTM

lhotse/workflows/whisper.py
desh2608: So I ran the model on one of the AMI headset recordings (~5000s) and it seems like the "end" actually shows the end of the segment, not the duration. Here is the JSON containing the `results["segments"]`: https://drive.google.com/file/d/169igkcDY2SmMs5k3hOhHip89T4MQDnKs/view?usp=sharing

PR #836, opened 2022-10-04
make regexp string raw

jtrmal:
please let me guys know what you think... I know there's been some fixes to ami and maybe some other recipes, so I don't want to end up undoing your changes again
/cc @desh2608 @pzelasko


pzelasko:
Can you test the affected recipes? Did it actually work before (or did sth break in this PR)?

pzelasko:
Thanks!

PR #838, opened 2022-10-05
Update supervisions channels in multi-channel recipes

desh2608:
In https://github.com/lhotse-speech/lhotse/pull/822, we updated the `channel` attribute of `SupervisionSegment` to be able to represent a single channel or a list of channels. This PR updates the existing multi-channel recipes to use this capability. For example, in the AMI "mdm" setting, each supervision segment is shared across all the channels.


desh2608:
> Looks good to me, did you test the recipes? If yes, LGTM!

Not extensively. Give me a couple of days to make sure they are good to go. I'll let you know once I have verified them.

desh2608:
I verified that the changes work for some of the recipes (LibriCSS, AMI). I think we can merge now and fix later if there are any issues.

pzelasko:
Great, thanks!

Issue #835, opened 2022-10-04
Error during fbank computing after the 1.9 release


teowenshen:
When computing fbank, the newer version of lhotse tripped at `compute_and_store_features`, inside the compute_fbank_xxx.py file. I suspect it might have something to do with the cut.py refactor. 

The error message is below. 

> root@2313535c79e8:/workspace/icefall/egs/csj/ASR# python local/compute_fbank_tedxjp10k.py --manifest-dir data/manifests --fbank-dir /mnt/minami_data_server/t2131178/corpus/TEDxJP-10K_v1.1/fbank
> 2022-10-04 10:22:37,936 INFO [compute_fbank_tedxjp10k.py:37] Manifests read.
> 2022-10-04 10:22:37,951 INFO [compute_fbank_tedxjp10k.py:45] Processing new
> /opt/conda/lib/python3.7/site-packages/lhotse/lazy.py:395: UserWarning: A lambda was passed to LazyMapper: it may prevent you from forking this process. If you experience issues with num_workers > 0 in torch.utils.data.DataLoader, try passing a regular function instead.
>   "A lambda was passed to LazyMapper: it may prevent you from forking this process. "
> Traceback (most recent call last):
>   File "local/compute_fbank_tedxjp10k.py", line 114, in <module>
>     main()  
>   File "local/compute_fbank_tedxjp10k.py", line 106, in main
>     num_mel_bins = 80
>   File "local/compute_fbank_tedxjp10k.py", line 65, in compute_fbank_csj
>     storage_type=ChunkedLilcomHdf5Writer
>   File "/opt/conda/lib/python3.7/site-packages/lhotse/cut/set.py", line 1455, in compute_and_store_features
>     cut_sets = self.split(num_jobs, shuffle=True)
>   File "/opt/conda/lib/python3.7/site-packages/lhotse/cut/set.py", line 541, in split
>     self, num_splits=num_splits, shuffle=shuffle, drop_last=drop_last
>   File "/opt/conda/lib/python3.7/site-packages/lhotse/utils.py", line 331, in split_sequence
>     seq = list(seq)
>   File "/opt/conda/lib/python3.7/site-packages/lhotse/cut/set.py", line 1988, in __len__
>     return len(self.cuts)
>   File "/opt/conda/lib/python3.7/site-packages/lhotse/lazy.py", line 231, in __len__
>     return sum(len(it) for it in self.iterators)
[...]

PR #844, opened 2022-10-10
Fix CutSet.compute_and_store_features support for lazy CutSets

pzelasko:
Resolves https://github.com/lhotse-speech/lhotse/issues/835

It should also help with https://github.com/k2-fsa/icefall/issues/608

@csukuangfj can you check it out?

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -1409,14 +1409,13 @@ class CutSet(Serializable, AlgorithmMixin):
             for parallel computation).
         :return: Returns a new ``CutSet`` with ``Features`` manifests attached to the cuts.
         """
-        from cytoolz import identity
-
+        from lhotse.lazy import LazySlicer
         from lhotse.manipulation import combine
 
         # Pre-conditions and args setup
         progress = (
-            identity  # does nothing, unless we overwrite it with an actual prog bar
-        )
+            lambda x: x
+        )  # does nothing, unless we overwrite it with an actual prog bar
         if num_jobs is None:
             num_jobs = 1
         if num_jobs == 1 and executor is not None:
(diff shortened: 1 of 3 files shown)


csukuangfj:
Thanks! I am testing it.

csukuangfj:
@pzelasko 

It throws the following error with this PR:

Screen Shot 2022-10-10 at 10 50 20
Screen Shot 2022-10-10 at 10 50 37

csukuangfj:
Also, the process gets stuck until I process `ctrl + C` to kill it:
Screen Shot 2022-10-10 at 10 53 24


Screen Shot 2022-10-10 at 10 53 41

pzelasko:
I see. When calling `.filter()`, you would have to change the lambda into a function or `pip install dill` to allow pickling lambdas (lhotse auto-magically supports this); there is no other way if you want to keep the manifest lazy, because it's using a lambda in `.filter` and attempting to pass it to subprocesses, so it must be pickled.

csukuangfj:
Thanks!

```
pip install dill
```

works perfectly. 

I don't use `pip install` to install lhotse so I did not have `dill` in my environment. After installing `dill` manually, everything works fine.

csukuangfj (approved):

pzelasko:
@csukuangfj `dill` is an optional dependency for Lhotse, as I am trying to limit the scope the amount of core dependencies as much as possible. That means unless it becomes a core dependency of Icefall, the users will run into the same issue as you with hanging training due to lambda. If the solution in this PR is acceptable, then I suggest to add `dill` to Icefall requirements.

csukuangfj:
> @csukuangfj `dill` is an optional dependency for Lhotse, as I am trying to limit the scope the amount of core dependencies as much as possible. That means unless it becomes a core dependency of Icefall, the users will run into the same issue as you with hanging training due to lambda. If the solution in this PR is acceptable, then I suggest to add `dill` to Icefall requirements.

Thanks! Added in https://github.com/k2-fsa/icefall/pull/613

PR #847, opened 2022-10-10
Fix timestamps in Whisper annotation workflow

### LINKED ISSUE #839: Whisper workflow supervision `end` may be incorrect
So I ran the model on one of the AMI headset recordings (~5000s) and it seems like the "end" actually shows the end of the segment, not the duration. Here is the JSON containing the `results["segments"]`: https://drive.google.com/file/d/169igkcDY2SmMs5k3hOhHip89T4MQDnKs/view?usp=sharing

_Originally posted by @desh2608 in https://github.com/lhotse-speech/lhotse/pull/834#discussion_r988376898_

pzelasko:
Closes #839

### DIFF
--- a/lhotse/workflows/whisper.py
+++ b/lhotse/workflows/whisper.py
@@ -3,7 +3,14 @@ from typing import Any, Generator, List, Optional, Union
 
 import torch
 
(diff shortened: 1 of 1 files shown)


desh2608 (approved):
LGTM, thanks!

PR #849, opened 2022-10-11
forced alignment: use `num2words` to get word timestamps for numbers

eschmidbauer:


pzelasko:
Cool! Can you move the num2words import to the local scope of `_normalize_text` and add a check at the top of `align_with_torchaudio` that does sth like:

```python
from lhotse.utils import is_module_available

assert is_module_available("num2words"), "To align with torchaudio, please run 'pip install num2words' for number to word normalization."
```

.github/workflows/unit_tests.yml
```diff
@@ -57,6 +57,7 @@ jobs:
         pip install kaldi_native_io  # for running kaldi_native_io tests
         pip install webdataset==0.2.5 # for running webdataset tests
         pip install git+https://github.com/fgnt/nara_wpe  # for running WPE tests
+        pip install num2words
```
pzelasko, line 60: I'd rather not add it to the CI, since we are not testing torchaudio forced alignment here. I don't want to make `num2words` a dependency of Lhotse, i.e. the code has to be able to import correctly without `num2words` being installed. I'd like the CI to test that.

pzelasko:
Thanks again!

Issue #846, opened 2022-10-10
Draft specification of a new, extensible sequential storage format 


pzelasko:
# Motivation

Currently Lhotse offers two types of data storage: 
- random-access (HDF5, lilcom-chunky, individual files, etc.) with manifests that are fast to read and easy to extend with new fields;
- sequential (tarfiles with WebDataset) with manifests stored together with data, which maximizes I/O but makes metadata modification painful (basically each modification requires a new copy of the whole thing)

I'd like to have a format that is both efficient to read (comparable to WebDataset tarfile efficiency) and easy to modify the metadata (like editing a JSONL file is) or extend with new pieces of metadata or custom data fields (such as other types of features, quantized codebook indexes, etc.).

# Implementation

## Core idea

My idea is to keep the sequential storage, but keep the manifest and the data stored in separate files. That means we could have `cuts.jsonl.gz` for the manifest, and `cuts.tar` for audio/features/etc. At runtime, we read both of them in sync and populate the cuts with in-memory raw data (e.g. compressed lilcom binary blobs for features, encoded audio data, etc.) which is decoded into numpy arrays only when we call `.load_X` on the cut. Lhotse would need to offer an API that reads both of these files (or a directory with a standardized file layout) to make sure iteration over metadata and data is kept in sync. 

The above expressed in pseudo-code:

```python
# variant 1:
cuts = CutSet.from_new_format("cuts.jsonl.gz", data="cuts.tar")

# variant 2:
# Expected contents of my_data_dir:
# $ ls my_data_dir/
# cuts.jsonl.gz cuts.tar
cuts = CutSet.from_new_format("my_data_dir/")

# The underlying CutSet reads manifest lines and the corresponding data sequentially in sync from the two files
# The following conditions are the same as in WebDataset based approach:
# - the data discarded by .filter is read into memory anyway
# - the shuffle operation keeps a buffer in memory that contains the actual data in addition to the metadata
for cut in cuts.filter(lambda c: c.duration >= 5).shuffle().subset(first=1000):
    audio = cut.load_audio()  # the data was already in memory, but it's an implementation detail
    ...
```

## Modularity extension

The idea above can be extended to support multiple sequentially read tarfiles with data, each for a single data field type, e.g.: 

```bash
$ ls my_data_dir/
cuts.jsonl.gz audio.tar features.tar codebook_indexes.tar ...
```

[...]

PR #850, opened 2022-10-11
Sharded tar writers for Lhotse Shar format

pzelasko:
This is part of the work needed to implement #846 

To make it easier to refer to this new data format I decided to name it `shar` (https://en.wikipedia.org/wiki/Lhotse_Shar), which can also be interpreted as "sharded archives".

I added tar writers that support saving recordings, features, arrays, and temporal arrays in a sharded manner. There is a high level `SharWriter` that shards the cutset and all of its relevant data files. It creates relevant writer for all data fields (supports wav/flac/mp3 for audio and numpy/lilcom for arrays). We remove any references to data in the cutset itself -- that data will be re-populated later by combining cutset+multiple tar files in a special reader class, and wrapped into `CutSet.from_shar` or something like that. I am also considering an option where it's no longer a CutSet, but I haven't decided yet.


danpovey:
Apparently Lhotse Shar is the most deadly of the 8-thousander peaks, even worse than k2 :-(

pzelasko:
That brings a new meaning to the expression "I will die on this hill" :)

pzelasko:
I wrote two versions of readers: one uses ``torchdata``, the other doesn't use any external dependency. Also, I added some documentation for the code written so far. 

Technically this works now, but to be usable it requires at least the following features:
- [ ] Support for multi-worker and multi-node training, either one (or both) of these variants:
    - Data/shard de-duplication for DataLoader (WebDataset approach)
    - Different (but deterministic) shuffle of shards for every node+worker (Dan's suggestion; will make each epoch artificially longer by the factor of num_workers*num_gpus, but as Dan noted the user can select to checkpoint every N steps instead)

In addition to be more useful, it needs:
- [ ] Support for non-local filesystem storage (e.g. cloud)
- [ ] Support for reading just a subset of fields (as an option in readers)
- [ ] Support for extending an existing Shar archive with a new field (as an option in the writer or as a separate writer)
- [ ] Benchmarking vs WebDataset/single tar archive on slow storage so see how much behind it is due to multiple archive

Then, I would like to investigate how to extend/adapt this work to support:
- [ ] Advanced dataloader checkpointing for resuming training by skipping shards
- [ ] (maybe) ensuring equal number of batches from each dataloader worker

lhotse/audio.py
```diff
@@ -470,13 +475,7 @@ def move_to_memory(
...
-            "wav",
-            "ogg",
-            "vorbis",
-            "flac",
-            "sph",
-        ):
+        if torchaudio_soundfile_supports_format() and format == "flac":
             # Prefer saving with soundfile backend whenever possible to avoid issue:
             # https://github.com/pytorch/audio/issues/2662
             # Saving with sox_io backend to FLAC may corrupt the file, IDK about other
```
desh2608, line 481: So, other formats (except flac) can be safely saved with the sox_io backend?

pzelasko: I believe so, at least I never experienced an issue with a different format.

pzelasko:
I did a basic benchmark reading a single WebDataset tar file, and a collection of Shar archives from S3, and achieved ~3x greater throughput with Shar. This is probably because it could read the data in parallel using multiple connections, as it was using pipe input to read from `pipe:aws s3 cp <url> -`. I don't have any magnetic HDD around to test it with that, but I expect Shar at least shouldn't be (much) worse than WDS. I'll merge this as is with all tests passing. I might tweak some things later, and will try to figure out how to implement the training checkpointing mechanism.

PR #851, opened 2022-10-12
Prepare CSJ

teowenshen:
# Prepare for CSJ 

This pull request diverges from the standard prepare of other recipes by a few points. I was hoping to get your input on these points in order to standardise my `prepare_csj` before the merge. 

1. **Train-Valid Splitting** 
CSJ does not define a separate validation dataset, so I follow the splitting scheme in espnet2 where 4000 utterances from the training datasets are picked as validation. Because one recording corresponds to multiple supervisions, I first create a concatenated cutset, trim the cutsets to its supervisions, then split that resulting cutset. The train cutset is further perturbed in speed here.  Then, I returned the cuts instead in the return value of `prepare_csj`. 
   - Is there a recommended way to split the resulting cutset into its constituent recordingset and supervisionset so that it can returned?
   - Alternatively, is there a recommended way or an existing recipe to refer to for this train-valid splitting?

2. **Cutset saved to disc** Since the cutsets are already created in this step, I saved the cutsets directly to disc so that compute_fbank_csj.py just loads the cuts_XXX.jsonl.gz. What do you think about this flow? (Once I can extract the recordingset and supervisionset from the cutset, I will save those to disc too.)

3. **Transcript preparation** Generating the transcript from the raw tsv file took me a separate python program. I start `prepare_csj` with the assumption that the transcript has been generated using that program, which I plan to include in the icefall recipe. What do you think? 

Also, please let me know if I need to change anything else in my codes. It's my first time using click, so I am not confident there.

### DIFF
--- a/lhotse/bin/modes/recipes/__init__.py
+++ b/lhotse/bin/modes/recipes/__init__.py
@@ -15,8 +15,8 @@ from .cmu_arctic import *
 from .cmu_indic import *
 from .cmu_kids import *
 from .commonvoice import *
-from .cslu_kids import *
 from .csj import *
+from .cslu_kids import *
 from .daily_talk import *
 from .dihard3 import *
 from .earnings21 import *
--- a/lhotse/bin/modes/recipes/__init__.py
+++ b/lhotse/bin/modes/recipes/__init__.py
@@ -16,6 +16,7 @@ from .cmu_indic import *
 from .cmu_kids import *
 from .commonvoice import *
 from .cslu_kids import *
+from .csj import *
 from .daily_talk import *
 from .dihard3 import *
 from .earnings21 import *
--- a/lhotse/recipes/csj.py
+++ b/lhotse/recipes/csj.py
@@ -39,7 +39,6 @@ Notice that the 'D' transcripts are split into respective (L)eft and (R)ight cha
 
 """
 
-from glob import glob
 import logging
 from concurrent.futures.thread import ThreadPoolExecutor
 from pathlib import Path
@@ -66,14 +65,14 @@ def parse_transcript_header(line : str):
     return (sgid, float(start), float(end), line)
 
 def parse_one_recording(
-    template : str,
+    template : Path,
     wavlist_path : Path, 
     recording_id : str
 ) -> Tuple[Recording, List[SupervisionSegment]]:
     transcripts = []
     
-    for trans in glob(template + '*.txt'):
-        trans_type = trans.replace(template + '-', '').replace(".txt", '')
+    for trans in template.glob(f'{recording_id}*.txt'):
+        trans_type = trans.stem.replace(recording_id + '-', '')
         transcripts.append([(trans_type, t) for t in Path(trans).read_text().split('\n')])
     
     assert all(len(c) == len(transcripts[0]) for c in transcripts), transcripts
@@ -156,8 +155,9 @@ def prepare_csj(
             futures = []
             
             for wavlist in part_path.glob("*/*-wav.list"):
-                template = wavlist.as_posix().rstrip('-wav.list')
                 spk = wavlist.name.rstrip('-wav.list')
+                template = wavlist.parent
+                
                 futures.append(
                     ex.submit(parse_one_recording, template, wavlist, spk)
                 )
--- a/lhotse/recipes/csj.py
+++ b/lhotse/recipes/csj.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
 # Copyright    2022  The University of Electro-Communications  (Author: Teo Wen Shen)  # noqa
 #
-# Adapted from kaldi-asr/kaldi/egs/csj/s5/local/csj_make_trans/{csj_autorun.sh,csj2kaldim.pl,csjconnect.pl}  # noqa
+# Adapted from kaldi-asr/kaldi/egs/csj/s5/local/csj_make_trans/{csj_autorun.sh,csj2kaldi4m.pl,csjconnect.pl}  # noqa
 #
-# See ../LICENSE for clarification regarding multiple authors
+# See ../../../LICENSE for clarification regarding multiple authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -1282,8 +1282,8 @@ def create_trans_dir(corpus_dir: Path, trans_dir: Path):
     logging.info("Transcripts have been moved.")
 
 
-def parse_sdb_process(
-    jobs_queue: Queue,
+def parse_one_sdb(
+    sdb: Path,
     gap: float,
     maxlen: float,
     minlen: float,
@@ -1292,39 +1292,32 @@ def parse_sdb_process(
     use_segments: bool,
     write_segments: bool,
 ):
-    def parse_one_sdb(sdb: Path):
-        with sdb.open("r", encoding="shift_jis") as fin:
-            result = CSJSDB_Word.from_file(fin)
+    with sdb.open("r", encoding="shift_jis") as fin:
+        result = CSJSDB_Word.from_file(fin)
 
-        if not use_segments:
-            transcripts = make_text(result, gap, maxlen, minlen, gap_sym)
-        else:
-            channels = (
-                ["-L-segments", "-R-segments"] if sdb.name[0] == "D" else ["-segments"]
+    if not use_segments:
+        transcripts = make_text(result, gap, maxlen, minlen, gap_sym)
+    else:
+        channels = (
+            ["-L-segments", "-R-segments"] if sdb.name[0] == "D" else ["-segments"]
+        )
+        transcripts = []
+        for channel in channels:
+            segments = Path(sdb.as_posix()[:-4] + channel).read_text().split("\n")
+            assert segments, segments
+            transcripts.append(modify_text(result, segments, "", 0.5))
+
+    for transcript in transcripts:
+        spk_id = transcript.pop("spk_id")
+        segments = transcript.pop("segments")
+        (sdb.parent / f"{spk_id}-{trans_mode}.txt").write_text(
+            "\n".join(transcript["text"]), encoding="utf8"
+        )
+        if write_segments:
+            (sdb.parent / f"{spk_id}-segments").write_text(
+                "\n".join(f"{s[0]} {s[1]} {s[2]}" for s in segments),
+                encoding="utf8",
             )
-            transcripts = []
-            for channel in channels:
-                segments = Path(sdb.as_posix()[:-4] + channel).read_text().split("\n")
-                assert segments, segments
-                transcripts.append(modify_text(result, segments, "", 0.5))
-
-        for transcript in transcripts:
-            spk_id = transcript.pop("spk_id")
-            segments = transcript.pop("segments")
-            (sdb.parent / f"{spk_id}-{trans_mode}.txt").write_text(
-                "\n".join(transcript["text"]), encoding="utf8"
-            )
-            if write_segments:
-                (sdb.parent / f"{spk_id}-segments").write_text(
-                    "\n".join(f"{s[0]} {s[1]} {s[2]}" for s in segments),
-                    encoding="utf8",
-                )
-
-    while True:
-        job = jobs_queue.get()
-        if not job:
-            break
-        parse_one_sdb(sdb=job)
 
 
 def load_config(config_file: str):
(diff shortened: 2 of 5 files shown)


jtrmal:
I think fwiw not making the normalization part of this makes sense. But
please document it somewhere (maybe even print it after data preparation)
so that it's easy to figure out.
y.

lhotse/recipes/csj.py
```diff
@@ -0,0 +1,246 @@
...
+    return (sgid, float(start), float(end), line)
+
+def parse_one_recording(
+    template : str,
+    wavlist_path : Path, 
+    recording_id : str
+) -> Tuple[Recording, List[SupervisionSegment]]:
+    transcripts = []
+    
+    for trans in glob(template + '*.txt'):
```
pzelasko, line 96: I recommend using `pathlib.Path` for path operations (then you can write sth like `template.glob("*.txt")`)

desh2608:
> # Prepare for CSJ
> This pull request diverges from the standard prepare of other recipes by a few points. I was hoping to get your input on these points in order to standardise my `prepare_csj` before the merge.
> 
> 1. **Train-Valid Splitting**
>    CSJ does not define a separate validation dataset, so I follow the splitting scheme in espnet2 where 4000 utterances from the training datasets are picked as validation. Because one recording corresponds to multiple supervisions, I first create a concatenated cutset, trim the cutsets to its supervisions, then split that resulting cutset. The train cutset is further perturbed in speed here.  Then, I returned the cuts instead in the return value of `prepare_csj`.
>    
>    * Is there a recommended way to split the resulting cutset into its constituent recordingset and supervisionset so that it can returned?
>    * Alternatively, is there a recommended way or an existing recipe to refer to for this train-valid splitting?

In general, if the corpus does not provide a data split, we just return the consolidated recordings and supervisions instead of creating our own splits, and the user can then split it in the way they want depending on their use case. See the `cmu_kids` recipe for example.

> 2. **Cutset saved to disc** Since the cutsets are already created in this step, I saved the cutsets directly to disc so that compute_fbank_csj.py just loads the cuts_XXX.jsonl.gz. What do you think about this flow? (Once I can extract the recordingset and supervisionset from the cutset, I will save those to disc too.)

You can use the `CutSet.decompose()` function to get the consitituents of the cut set. But you won't need this if you only return the full recordings and supervisions.

> 3. **Transcript preparation** Generating the transcript from the raw tsv file took me a separate python program. I start `prepare_csj` with the assumption that the transcript has been generated using that program, which I plan to include in the icefall recipe. What do you think?

I think this is reasonable. Some corpora require a bit of preprocessing to get them in a suitable format to read with a python program. CHiME-5, for example, requires array synchronization, and LibriCSS requires running a segmentation script provided with the official data. You can just mention the details about the assumption in the header so that the user can prepare it in the format required.
> 
> Also, please let me know if I need to change anything else in my codes. It's my first time using click, so I am not confident there.

lhotse/bin/modes/recipes/__init__.py
```diff
@@ -16,6 +16,7 @@
 from .cmu_kids import *
 from .commonvoice import *
 from .cslu_kids import *
+from .csj import *
```
desh2608, line 19: Should be above `cslu_kids`

lhotse/recipes/csj.py
```diff
@@ -0,0 +1,246 @@
...
+
+from lhotse import validate_recordings_and_supervisions, CutSet
+from lhotse.audio import Recording, RecordingSet
+from lhotse.recipes.utils import manifests_exist, read_manifests_if_cached
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike
+
+import random
+
+"""
```
desh2608, line 18: I think we usually put this at the top of the file before the imports.

lhotse/recipes/csj.py
```diff
@@ -0,0 +1,246 @@
...
+    
+    logging.info(f"Creating train cuts")
+    train_set = CutSet.from_cuts(islice(cut_set, split, None))
+    
+    train_set = (
+        train_set
+        + train_set.perturb_speed(0.9)
+        + train_set.perturb_speed(1.1)
+    )
+    train_set.to_jsonl(output_dir / "cuts_train.jsonl.gz")
```
desh2608, line 228: Use `to_file` instead of `to_jsonl`.

lhotse/recipes/csj.py
```diff
@@ -0,0 +1,246 @@
...
+    
+    logging.info("Creating eval cuts.")
+    # Create eval datasets
+    for i in range(1, 4):
+        cut_set = CutSet.from_manifests(
+            recordings=RecordingSet.from_file(output_dir / f"recordings_eval{i}.jsonl.gz"),
+            supervisions=SupervisionSet.from_file(output_dir / f"supervisions_eval{i}.jsonl.gz")
+        )
+        cut_set = cut_set.trim_to_supervisions(keep_overlapping=False)
+        cut_set.to_jsonl(output_dir / f"cuts_eval{i}.jsonl.gz")
```
desh2608, line 243: Same as above.

desh2608 (changes requested):

pzelasko (commented):
Thanks, the recipe looks good overall! Can you also update the table in `docs/corpus.rst`? 

Regarding the transcripts, Lhotse recipes are generally supposed to work with raw data distribution as-is and be the very first step of the pipeline. I would suggest to incorporate your separate script for preparing the transcript as a function that's called inside this data prep recipe. I don't know CSJ, maybe the transcript preprocessing work is very involved and not suitable for Lhotse recipe, but I'd like you to at least attempt to convince me first :) 

Honestly, I would avoid returning cuts as done here. I agree with Desh this is something that is typically done in downstream applications, if there's no dev set, we just don't provide it. The code you wrote for cuts could be moved to the Icefall recipe.

jtrmal:
I think we do, desh -- thats even the change you were doing -- saving as
"all_supervisions" vs "{dev,test,train}_supervisions"?

desh2608:
@jtrmal Yeah, in such cases we would save as `recordings_<corpus>_all.jsonl.gz` instead.

lhotse/recipes/csj.py
teowenshen: After the latest commit, this part was removed to return `SupervisionSet` and `RecordingSet` directly.

lhotse/recipes/csj.py
teowenshen: Because `template` is an incomplete path, I can't make `Path.glob` find the files appended with transcript type. 
Another reason is that I extract the transcript type from the resulting filename string. From a `Path` object, I will need to first convert it the path to a string. 

Though admittedly, I think globbing like this might break on Windows. 

```python
> from pathlib import Path
> partial_filename = Path("/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051/A04M0051")
> list(partial_filename.glob("*.txt"))
[]
> intended_result = Path("/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051")
> list(intended_result.glob("*.txt"))
[PosixPath('/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051/A04M0051-fluent.txt'), PosixPath('/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051/A04M0051-disfluent.txt'), PosixPath('/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051/A04M0051-number.txt'), PosixPath('/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051/A04M0051-symbol.txt')]
```

lhotse/recipes/csj.py
```diff
@@ -0,0 +1,246 @@
...
+                recordings.append(recording)
+                supervisions.extend(segments)
+            
+            recording_set = RecordingSet.from_recordings(recordings)
+            supervision_set = SupervisionSet.from_segments(supervisions)
+            validate_recordings_and_supervisions(recording_set, supervision_set)
+            
+            if output_dir:
+                supervision_set.to_file(output_dir / f"supervisions_{part}.jsonl.gz")
+                recording_set.to_file(output_dir / f"recordings_{part}.jsonl.gz")
```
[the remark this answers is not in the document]
teowenshen: Yes. I followed this suggestion and removed all `CutSet`-related operations. I will implement the splitting in another script.

lhotse/bin/modes/recipes/__init__.py
teowenshen: done

lhotse/recipes/csj.py
teowenshen: done

lhotse/recipes/csj.py
teowenshen: Removed too.

lhotse/recipes/csj.py
desh2608: How about the following:
```
dir = Path("/mnt/minami_data_server/t2131178/corpus/CSJ/retranscript_new/eval1/A04M0051")
partial_filename = dir.stem
list(dir.glob(f"{partial_filename}*.txt"))
```

teowenshen:
Added some changes:-

- Return supervisions and recordings instead of cutset 
- Parsing dataset parts as-is. (train dataset actually is a concatenation of core and noncore parts. Here, I return them as separate dataset parts.)
- Added gender info in `SupervisionSegment` 
- Updated docstrings

> In general, if the corpus does not provide a data split, we just return the consolidated recordings and supervisions instead of creating our own splits, and the user can then split it in the way they want depending on their use case. See the cmu_kids recipe for example.

Yes. Probably the best way is to have a separate program for the dataset preparation before compute_fbank_csj.py. 

> Can you also update the table in `docs/corpus.rst`?

Done. 

> Regarding the transcripts, Lhotse recipes are generally supposed to work with raw data distribution as-is and be the very first step of the pipeline. I would suggest to incorporate your separate script for preparing the transcript as a function that's called inside this data prep recipe. I don't know CSJ, maybe the transcript preprocessing work is very involved and not suitable for Lhotse recipe, but I'd like you to at least attempt to convince me first :)

It's a 700-liner program that reads in a config file and optionally a segment file, and my `prepare.sh` runs it recursively for each transcript mode. 😂 However, I do understand the motivation behind consolidating all transcript preparation scripts into lhotse. 

I plan to update my Icefall recipe and pull in a working version in Icefall. I will link it here once it's done. 

Ideally, I was thinking that maybe a separate command in lhotse could be used for these processes. `lhotse parse` ?

lhotse/recipes/csj.py
teowenshen: Thanks! It slipped my mind that I already have the spk_id. 
I've removed the reliance on `glob.glob`.

teowenshen:
I have just sent in the pull request to Icefall for the data preparation part of the recipe. 

This is the file that generates the transcripts from the raw tsv file: [csj_make_transcript.py](https://github.com/Minami-Lab-UEC/icefall/blob/ef6cfc34ef9b2385b08e4f7aac9a6d3241a492f4/egs/csj/ASR/local/csj_make_transcript.py#L734)

The transcript generation is so involved, partly because there are many tags that mark slurring, laughing, clapping, stuttering, orthography variation, etc. 

Another Japanese corpus that we plan to work on (not so soon) also requires an external script to parse the transcript from raw subtitle files. That corpus probably will have the same issue.

desh2608:
Can you fix the style checks? You can use:
```
pre-commit install
pre-commit run
```

teowenshen:
Okay, I think I have done cleared the style checks.

desh2608:
> Okay, I think I have done cleared the style checks.

Black seems to be failing still.

teowenshen:
Somehow pre-commit shows that there are no files to check. 
Anyhow I ran black again. I was splitting the long lines, which was why black complained.

> check that executables have shebangs.................(no files to check)Skipped
> fix end of files.....................................(no files to check)Skipped
> mixed line ending....................................(no files to check)Skipped
> trim trailing whitespace.............................(no files to check)Skipped
> flake8...............................................(no files to check)Skipped
> isort................................................(no files to check)Skipped
> black................................................(no files to check)Skipped

desh2608 (dismissed):

jtrmal:
makes sense to me -- I agree
y.

desh2608:
I second Piotr. If the conf.ini files are needed for the corpus preparation, they can be put on a server (e.g. openslr) and downloaded inside the preparation recipe.

teowenshen:
Yes. I would also like to standardise this corpus with other corpora too.  

Is it okay to add more optional arguments behind lhotse prepare? I will include a default config in the codes. If the user wants to change the behaviour, they can do so by supplying the path to their own modified config file.

pzelasko:
Yes, various recipe have their own recipe-specific optional arguments.

teowenshen:
Great! I will update my codes and send in a pull request soon.

teowenshen:
I have moved the transcript generation part over to Lhotse, so now this recipe should be performing the same functions as other recipes. I included a default .ini file directly inside the codes so that this recipe is as self sufficient as possible.

pzelasko:
Thanks! Great work.

PR #852, opened 2022-10-12
Allow retaining or trimming channels in trim_to_supervisions

desh2608:
Consider AMI IHM as an example. For each session, we create a multi-cut with 4 channels (1 channel per speaker), and each supervision segment corresponds to one of the 4 channels (based on who the speaker is). If we then want to use this data to train an ASR system, we can call `trim_to_supervisions()`. Currently, the resulting cuts would still contain 4 channels. This PR allows the user to trim the channels to the supervisions (by default), or ignore channel while trimming.

### DIFF
--- a/lhotse/cut/base.py
+++ b/lhotse/cut/base.py
@@ -470,17 +470,17 @@ class Cut:
             if not keep_overlapping:
                 # Ensure that there is exactly one supervision per cut.
                 trimmed = trimmed.filter_supervisions(lambda s: s.id == segment.id)
-            if not keep_all_channels:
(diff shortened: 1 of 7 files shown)


lhotse/cut/base.py
```diff
@@ -394,6 +394,7 @@ def trim_to_supervisions(
         keep_overlapping: bool = True,
         min_duration: Optional[Seconds] = None,
         context_direction: Literal["center", "left", "right", "random"] = "center",
+        ignore_channel: bool = False,
```
pzelasko, line 397: What do you think about `keep_all_channels: bool = False`? `ignore_channel` is a little vague for me, but I'm not sure.

PR #853, opened 2022-10-15
Match `cut_id` to `utt_id` if there is exactly one supervision per cut

wgb14:


desh2608:
I am a little skeptical about enforcing this in the function by default.

wgb14:
> I am a little skeptical about enforcing this in the function by default.

See discussion here: https://github.com/k2-fsa/icefall/issues/620. what do you think would be a better way?

desh2608:
After reading the linked discussion, I think it makes sense in view of ease of use. But please add a sentence in the function's docstring explaining this behavior.

lhotse/cut/base.py
```diff
@@ -431,7 +431,9 @@ def trim_to_supervisions(
 
         :param keep_overlapping: when ``False``, it will discard parts of other supervisions that overlap with the
             main supervision. In the illustration above, it would discard ``Sup2`` in ``Cut1`` and ``Sup1`` in ``Cut2``.
-            In this mode, we guarantee that there will always be exactly one supervision per cut.
+            In this mode, we guarantee that there will always be exactly one supervision per cut, and we set the ``id``
```
wgb14, line 434: Does this look good?

desh2608: Yeah, perfect.

desh2608 (approved):

desh2608:
Let's get this checked in since it is a small change.

lhotse/cut/base.py
```diff
@@ -470,6 +472,7 @@ def trim_to_supervisions(
             if not keep_overlapping:
                 # Ensure that there is exactly one supervision per cut.
                 trimmed = trimmed.filter_supervisions(lambda s: s.id == segment.id)
+                trimmed.id = segment.id
```
pzelasko, line 475: This is a bit un-intuitive that the IDs will have one format or another depending on the setting `keep_overlapping`. Maybe we could always set the cut ID to supervision ID regardless of this flag?

desh2608: Oh, good catch. There could be a single supervision even when keep_overlapping is set to True.

wgb14: I see. I didn't do overlapping ASR corpus before, and didn't want to break any pipeline.
If it's still good to set `cut_id` to `utt_id` while there is overlapping, we can revert this PR and modify it, or just open a new one.

pzelasko: Feel free to open a new PR

desh2608: @wgb14 What I mean (and changed in https://github.com/lhotse-speech/lhotse/pull/855) was that even if `keep_overlapping` is set to True, it may happen that a trimmed cut has only 1 supervision, if there was no overlapping supervision originally. In such case, it makes sense to set the trimmed cut id to that of the supervision.

Now, even if the cut has multiple overlapping supervisions, one could argue that its boundaries are defined by 1 "main" supervision, and the others just happen to be included. So it would still make sense to change the cut id to that of the main supervision.

wgb14: Make sense, thanks

PR #855, opened 2022-10-17
Small changes in `trim_to_supervisions()`

desh2608:
Makes 2 changes:
* If trimmed cut has exactly 1 supervision, cut id is set to supervision id (see [this comment](https://github.com/lhotse-speech/lhotse/pull/853#discussion_r997447028) for reference).
* If MultiCut is trimmed and resulting cut contains 1 channel, it is returned as MonoCut.


pzelasko (approved):
Thanks!

PR #863, opened 2022-10-25
More similar mean batch duration across nodes with DynamicBucketingSampler in multi-GPU training

pzelasko:
Resolves https://github.com/lhotse-speech/lhotse/discussions/857

I don't think it's really possible to have the same duration buckets used on each GPU node, because every node may see a subset of data and have a different set of buckets ready for sampling (ready == able to yield at least one batch):

https://github.com/lhotse-speech/lhotse/blob/ff29065b76415e5c7221ca49578cff3101c6f887/lhotse/dataset/sampling/dynamic_bucketing.py#L360-L372

But I figured out that we can do the next best thing and instead of sampling the bucket index directly, we would sample a random value `v` in [0.0, 1.0) and quantize it to the first integer `k` that satisfies `k >= v * num_ready_buckets` in each node (note that each node may have a different `num_ready_buckets`). Then we select the bucket at index `k` -- since the buckets are sorted by duration, we will have a better chance than random at selecting buckets with similar durations on every node. It's still not a guarantee, and I'm not sure if it's possible to select the optimal buckets on every node without introducing inter-node communication (which gets ugly). Note this whole approach can work because the RNGs are seeded with the combination of seed + epoch, and we assume the sampler seed is identical on every node (usually true e.g. in Icefall recipes).

@david20181 would you be willing to give it a run and let me know if this accelerated your training? Since you already have the setup to measure this, it would make my life easier..

### DIFF
--- a/lhotse/dataset/sampling/dynamic_bucketing.py
+++ b/lhotse/dataset/sampling/dynamic_bucketing.py
@@ -16,7 +16,7 @@ from lhotse.dataset.sampling.base import (
     TimeConstraint,
 )
 from lhotse.dataset.sampling.dynamic import DurationBatcher, Filter
-from lhotse.utils import ifnone, streaming_shuffle
+from lhotse.utils import ifnone, quantize, streaming_shuffle
 
 
 class DynamicBucketingSampler(CutSampler):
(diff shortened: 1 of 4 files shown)


pzelasko:
Superseded by #1341

PR #870, opened 2022-10-27
Shar: tarfiles now also contain metadata

pzelasko:
This change is related to continued development of Lhotse Shar. It allows to add new fields to a CutSet dynamically. E.g.:

- when somebody converts cuts with audio to Shar, they will have a directory like:

```
cuts.000000.jsonl.gz cuts.000001.jsonl.gz recording.000000.tar recording.000001.tar
```

- then if somebody reads these cuts using `LazySharIterator`, computes some custom features `my_features`, and saves to Shar again, they will get:

```
cuts.000000.jsonl.gz cuts.000001.jsonl.gz my_features.000000.tar my_features.000001.tar
```

This change allows to use the cuts from the step 1 to be read together with `recordings` from step 1 and `my_features` from step 2. Previously the code would complain about missing `my_features` manifest in cuts.

### DIFF
--- a/lhotse/array.py
+++ b/lhotse/array.py
@@ -82,14 +82,14 @@ class Array:
         """
         return fastcopy(self, storage_path=str(Path(path) / self.storage_path))
 
-    def move_to_memory(self) -> "Array":
+    def move_to_memory(self, lilcom: bool = False) -> "Array":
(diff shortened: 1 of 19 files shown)


danpovey:
cool!

PR #874, opened 2022-10-31
Fixed an erroneous assertion

JinZr:
Fixed an erroneous assertion in SpecAugment which prevents ``num_frame_masks`` to be set to ``0`` to be disabled.


pzelasko:
Thanks!

PR #875, opened 2022-11-01
Small changes to make channel attribute hashable

desh2608:
Some comparisons require the channel attribute to be hashable (of `Tuple` type). This PR adds a utility function to convert channel to tuple. Other small changes, e.g., CLI option for keeping all channels in trim_to_supervisions.


pzelasko:
I’m OK to merge, but it would be useful if you also added a test that would have been broken before the PR and fixed with these changes.

desh2608:
> I’m OK to merge, but it would be useful if you also added a test that would have been broken before the PR and fixed with these changes.

Good point. Done!

pzelasko (approved):
Thanks!

PR #889, opened 2022-11-12
Add option to split AMI segments similar to Kaldi

desh2608:
Kaldi AMI data preparation includes splitting long segments based on full-stops and commas. This is useful for training ASR models since otherwise we would throw away a lot of data. Here are some statistics before and after adding this option:

| **# segments**         | **Train** | **Dev** | **Test** |
|------------------------|:---------:|:-------:|:--------:|
| Kaldi                  |   105149  |  13059  |   12612  |
| Lhotse (w/o splitting) |   65557   |   8665  |   7490   |
| Lhotse (w/ splitting)  |   103377  |  12614  |   12121  |

The remaining difference may be because we don't "over-segment" the segments if they are already under the max words.

### DIFF
--- a/lhotse/bin/modes/recipes/ami.py
+++ b/lhotse/bin/modes/recipes/ami.py
@@ -14,7 +14,9 @@ __all__ = ["ami"]
     "--annotations",
     type=click.Path(),
     default=None,
-    help="Provide if annotations are download in a different directory than corpus.",
+    help=(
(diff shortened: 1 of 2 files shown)


desh2608:
(@HuangZiliAndy and I have both observed that decoding WERs can be significantly better with the segment splitting.)

pzelasko (approved):

pzelasko:
Great!

PR #898, opened 2022-11-21
Fix an edge case with BucketingSampler and a small amount data/buckets

pzelasko:
@m-wiesner could you review for me to be sure that that the fix makes sense?


pzelasko:
I'll merge and fix if needed later -- for me it solves the issue (I'm just not sure if the buckets are still optimal in these edge cases, but maybe that's not too critical).

PR #903, opened 2022-11-27
Fix audio save for parallel workers

pkufool:


desh2608 (dismissed):

desh2608:
Could you add a test which would have failed before this change? (I think simply adding an additional `pytest.mark.parameterize("num_jobs")` in the `test_store_audio` test method should be fine.)

test/cut/test_cut_set.py
```diff
@@ -521,8 +521,10 @@ def test_map_cut_set_rejects_noncut(cut_set_with_mixed_cut):
         cut_set = cut_set_with_mixed_cut.map(lambda cut: "not-a-cut")
 
 
-def test_store_audio():
-    cut_set = CutSet.from_json("test/fixtures/libri/cuts.json")
+@pytest.mark.parametrize("num_jobs", [1, 2])
+def test_store_audio(num_jobs):
+    cut_set = CutSet.from_json("test/fixtures/ljspeech/cuts.json")
```
pkufool, line 526: libri only has one cut, so I use ljspeech for this test.

desh2608 (approved):

PR #909, opened 2022-12-05
AliMeeting recipe enhancement

desh2608:
This PR makes 2 enhancements to the AliMeeting recipe:
1. Add text normalization based on the original M2MeT challenge baseline.
2. Add option to save single-channel recording when preparing "SDM" mic. This can address the slow feature extraction issue due to loading 8-channel recordings every time (see https://github.com/lhotse-speech/lhotse/issues/911).


pzelasko (approved):
Thanks!

PR #912, opened 2022-12-07
Fix snr error in CutSet.mix

pkufool:
In current  code, the `snr` in `CutSet.mix` will be overwritten and make it same for all cuts in the CutSet, I think it is not expected when passing a range of snr.

@pzelasko I am not sure if it is enough to fix this by modifying these lines, please have a look. And I have not came up with a good way to test it, do you have any suggestions.


pzelasko:
I added the RNG seed and unit tests in #915 -- closing this one. Thanks @pkufool!

PR #919, opened 2022-12-09
Downloading the AMI dataset using CLI will trigger TypeError

JinZr:
Using the wrong keyword when calling the "download_ami" function triggers this error:

```
TypeError: download_ami() got an unexpected keyword argument 'annotations_dir'
```


pzelasko (approved):
Thanks!

PR #920, opened 2022-12-11
Add room and source RNG seeds option for `reverb_rir`

desh2608:
Currently when calling `MixedCut.reverb_rir()` without passing an explicit RIR, each track in the MixedCut would get convolved with an RIR generated with different room configurations. This PR fixes that issue by allowing an `rir_generator` Callable to be passed to the underlying cuts.

### DIFF
--- a/lhotse/cut/mixed.py
+++ b/lhotse/cut/mixed.py
@@ -649,6 +649,7 @@ class MixedCut(Cut):
         early_only: bool = False,
         affix_id: bool = True,
         rir_channels: List[int] = [0],
+        rir_generator: Optional[Callable] = None,
     ) -> "MixedCut":
         """
         Return a new ``MixedCut`` that will convolve the audio with the provided impulse response.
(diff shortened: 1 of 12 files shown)


lhotse/cut/mixed.py
```diff
@@ -688,10 +693,19 @@ def reverb_rir(
         if len(rir_channels) == 1:
             rir_channels = rir_channels * len(self.tracks)
 
-        # NOTE: Currently, if no RIR is provided, this method will generate a
-        # random one for each track in the MixedCut. This is not ideal since the room
-        # configuration for all the RIRs should be the same. But we ignore this for now
-        # since it simplifies the implementation considerably.
+        source_rng_seeds = [source_rng_seed] * len(self.tracks)
```
desh2608, line 696: This is the core part of the change. Earlier, a random RIR would be generated when `load_audio()` was called on the reverberated cut for each track. Now, we have 2 RNGs -- one for the room and another for the source location. We use the same room RNG seed for all tracks but different source RNG seed for each track. In this way, we can ensure that all tracks are reverberated with RIRs that are generated using the same room configuration but different source locations.

pzelasko (approved):
Thanks!

Issue #917, opened 2022-12-09
Unable to guarantee that the noise added audio energy is greater than 0.


drawfish:
Lhotse AudioMixer make an assertion that the energy of the  audio to mix in must greater than 0.0. However, in some rare case, this assertion cannot be satisfied.
For example, the case  below.
During noise augmention training where I mix a speech audio with two noise audio which were extracted from musan. 
The first noise audio is OK but for the second noise audio, there are only a small part of the samples is extracted(from 0.0 to 0.0001875 second). Because it's at the begining of the audio, the enery is 0.0.
I know that the target of assertion is to to ensure that the signal-to-noise ratio does not cause error. However, it's hard for user to guarantee this because the length and the samples to mix can not predict in advance.
So maybe we should solve this problem ‘smoothly’ instread of throw exception. 

```
WARNING:root:To perform mix, energy must be non-zero and non-negative (got 0.0).  MonoCut with id "noise-free-sound-0275-275-0" will not be mixed in.
112779 112782
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
File /data/k2/miniconda3/envs/k2-latest/lib/python3.8/site-packages/lhotse/utils.py:727, in rich_exception_info.<locals>.wrapper(*args, **kwargs)
    726 try:
--> 727     return fn(*args, **kwargs)
    728 except Exception as e:

File /data/k2/miniconda3/envs/k2-latest/lib/python3.8/site-packages/lhotse/cut/mixed.py:906, in MixedCut.load_audio(self, mixed, mono_downmix)
    905     print(audio.shape[1], self.num_samples)
--> 906     assert audio.shape[1] == self.num_samples, (
    907         f"Inconsistent number of samples in a MixedCut: please report "
    908         f"this issue at https://github.com/lhotse-speech/lhotse/issues "
    909         f"showing the cut below. MixedCut:\n{self}"
    910     )
    911 else:

AssertionError: Inconsistent number of samples in a MixedCut: please report this issue at https://github.com/lhotse-speech/lhotse/issues showing the cut below. MixedCut:
[...]

PR #922, opened 2022-12-12
Remove 'NonPositiveEnergyError' exception when mix audio

drawfish:
Remove 'NonPositiveEnergyError' exception when mix audio, still check if energy<=0.0 but just print warning.
For more detail reason, please see this [issue](https://github.com/lhotse-speech/lhotse/issues/917).

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1322,7 +1322,7 @@ class AudioMixer:
         """
         if audio.size == 0:
             return  # do nothing for empty arrays
-        
+
         if self.reference_energy <= 0.0:
             return
 
@@ -1341,7 +1341,9 @@ class AudioMixer:
                 # we need to take a square root of the energy ratio.
                 gain = sqrt(target_energy / added_audio_energy)
             else :
-                logging.warning("Non-positive energy audio track to mix, will not scale it with snr.")
+                logging.warning(
+                    "Non-positive energy audio track to mix, will not scale it with snr."
+                )
         self.tracks.append(gain * audio)
         self.offsets.append(num_samples_offset)
         # We cannot mix 2 multi-channel audios with different number of channels.
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1235,11 +1235,6 @@ class AudioMixer:
         else:
             self.reference_energy = reference_energy
 
-        if self.reference_energy <= 0.0:
-            logging.warning(
-                f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy}), ignore mix operation."
-            )
-
     def _pad_track(
         self, audio: np.ndarray, offset: int, total: Optional[int] = None
     ) -> np.ndarray:
@@ -1323,16 +1318,13 @@ class AudioMixer:
         if audio.size == 0:
             return  # do nothing for empty arrays
 
-        if self.reference_energy <= 0.0:
-            return
-
         assert offset >= 0.0, "Negative offset in mixing is not supported."
 
         num_samples_offset = compute_num_samples(offset, self.sampling_rate)
 
         # When SNR is requested, find what gain is needed to satisfy the SNR
         gain = 1.0
-        if snr is not None:
+        if snr is not None and self.reference_energy > 0:
             added_audio_energy = audio_energy(audio)
             if added_audio_energy > 0.0:
                 target_energy = self.reference_energy * (10.0 ** (-snr / 10))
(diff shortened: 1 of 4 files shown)


desh2608:
You may need to modify the tests in accordance with the new behavior.

lhotse/audio.py
```diff
@@ -1236,8 +1236,8 @@ def __init__(
             self.reference_energy = reference_energy
 
         if self.reference_energy <= 0.0:
-            raise NonPositiveEnergyError(
-                f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy})"
+            logging.warning(
```
pzelasko, line 1239: I would avoid the warning at all -- it will just annoy the users.

lhotse/audio.py
```diff
@@ -1323,6 +1323,9 @@ def add_to_mix(
         if audio.size == 0:
             return  # do nothing for empty arrays
 
+        if self.reference_energy <= 0.0:
```
pzelasko, line 1326: hmmm in that case we have audio data with all zeroes, and want to mix something in -- the result should be mixed in audio rather than all zeros. I would remove this `if` here

lhotse/audio.py
```diff
@@ -1331,16 +1334,16 @@ def add_to_mix(
         gain = 1.0
         if snr is not None:
```
pzelasko, line 1335: I would change this to 
```suggestion
        if snr is not None or self.reference_energy == 0:
```

lhotse/audio.py
```diff
@@ -1331,16 +1334,16 @@ def add_to_mix(
...
-                raise NonPositiveEnergyError(
-                    f"To perform mix, energy must be non-zero and non-negative (got {added_audio_energy}). "
+            if added_audio_energy > 0.0:
+                target_energy = self.reference_energy * (10.0 ** (-snr / 10))
+                # When mixing time-domain signals, we are working with root-power (field) quantities,
+                # whereas the energy ratio applies to power quantities. To compute the gain correctly,
+                # we need to take a square root of the energy ratio.
+                gain = sqrt(target_energy / added_audio_energy)
+            else:
+                logging.warning(
```
pzelasko, line 1344: We don't need the warning, it will just annoy users.

lhotse/features/mixer.py
```diff
@@ -58,9 +59,10 @@ def __init__(
             self.reference_energy = feature_extractor.compute_energy(base_feats)
         else:
             self.reference_energy = reference_energy
-        assert (
-            self.reference_energy > 0.0
-        ), f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy})"
+        if self.reference_energy <= 0.0:
+            logging.warning(
+                f"To perform mix, energy must be non-zero and non-negative (got {self.reference_energy}), ignore mix operation."
```
pzelasko, line 64: Same as above, I'd just remove this.

pzelasko: The rest of the comments I made above in `audio.py` are also relevant for this file

pzelasko (commented):
Thanks @drawfish, I left some comments, would you be able to address those?

drawfish:
> Thanks @drawfish, I left some comments, would you be able to address those?
Done~

drawfish:
> Thanks @drawfish, I left some comments, would you be able to address those?

Done~

test/known_issues/test_mixing_zero_energy_cuts.py
```diff
@@ -17,8 +17,8 @@ def test_mix_zero_energy_cut_raises(self, snr):
 
         mixed = zero_cut.mix(rand_cut, snr=snr)
 
-        with pytest.raises(NonPositiveEnergyError):
-            mixed.load_audio()
+        mix_cut_samples = mixed.load_audio()
+        assert mix_cut_samples.shape[1] == sr
```
pzelasko, line 21: can you additionally assert that `np.testing.assert_equal(rand_cut.load_audio(), mix_cut_sampler)`?

test/known_issues/test_mixing_zero_energy_cuts.py
```diff
@@ -17,8 +17,8 @@ def test_mix_zero_energy_cut_raises(self, snr):
...
 
     @pytest.mark.parametrize("snr", [None, 10])
     def test_fault_tolerant_loading_skips_cut(self, snr):
```
pzelasko, line 24: Can you remove this test? It does not do anything right now since mixing zero cuts cannot trigger failure in audio loading.

pzelasko (commented):
Thanks, one last round of review and LGTM!

test/known_issues/test_mixing_zero_energy_cuts.py
drawfish: OK~

pzelasko (approved):

pzelasko:
Thanks!

Issue #923, opened 2022-12-12
Using forced alignment for segmentation


lubacien:
Hello,

I have a dataset made of short utterances spaced with long silences, but no start and duration information about the segments. Therefore I create supervision segments which last the full duration of my recordings and with long transcriptions.

example:
audio: ----hello---(20s silence)---yes---(10s silence)---no---
transcription: HELLO YES NO
start: 0s
end: 35s

I can calculate the forced alignment, for example using the "align_with_torchaudio" workflow or adding an external ctm file, but is there a way to then use these alignments to create a segmented version of my data?
 
The goal would be to obtain 3 supervision segments, something like:
transcription: HELLO
start: 0s
end: 1s
transcription:YES
start: 21s
end: 22s
transcription: NO
start: 32s
end: 33s

 thanks

desh2608:
I needed this function for my own work, so I went ahead and added it. It would be great if you could help test it out.

lubacien:
thanks! i'll take a look

PR #926, opened 2022-12-13
Add `trim_to_alignments()` method

desh2608:
Based on discussion in https://github.com/lhotse-speech/lhotse/issues/923

### DIFF
--- a/lhotse/cut/base.py
+++ b/lhotse/cut/base.py
@@ -558,6 +558,10 @@ class Cut:
             # `alignment` field in the new supervisions.
             merged_alignments = [(alignments[0], [0])]
             for i, item in enumerate(alignments[1:]):
+                # If alignment item is blank, skip it. Sometimes, blank alignment items
+                # are used to denote pauses in the utterance.
+                if item.symbol.strip() == "":
+                    continue
                 prev_item, prev_indices = merged_alignments[-1]
                 if item.start - prev_item.end <= max_pause:
                     new_item = AlignmentItem(
--- a/lhotse/cut/base.py
+++ b/lhotse/cut/base.py
@@ -506,6 +506,92 @@ class Cut:
             cuts.append(trimmed)
         return CutSet.from_cuts(cuts)
 
+    def trim_to_alignments(
+        self,
+        type: str,
+        max_pause: Seconds = 0.0,
+        delimiter: str = " ",
+        keep_all_channels: bool = False,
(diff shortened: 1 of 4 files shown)


lhotse/cut/base.py
```diff
@@ -506,6 +506,93 @@ def trim_to_supervisions(
...
+                continue
+            alignments = sorted(segment.alignment[type], key=lambda a: a.start)
+
+            # Merge the alignments if needed. We also keep track of the indices of the
+            # merged alignments in the original list. This is needed to create the
+            # `alignment` field in the new supervisions.
+            merged_alignments = [(alignments[0], [0])]
+            for i, item in enumerate(alignments[1:]):
+                prev_item, prev_indices = merged_alignments[-1]
+                if item.start - prev_item.end <= max_pause:
```
desh2608, line 562: A catch here is that if there are alignment items which are continuous (i.e. `prev_end == cur_start`) then they will necessarily get merged. So in order to trim such that we get #cuts = num. of alignment items, a negative value of `max_pause` will need to be passed.

desh2608:
> All tests are passing on my side (although I have to use `torch.set_num_threads(1)` to pass the test `test_compute_and_store_features_lazy(nj)` with `nj=2`, this is not related). I also tested it from command line on my data and it works as expected.

Thanks for testing this. I will verify it on my end and then ask @pzelasko to merge if he doesn't see any immediate issues.

lhotse/cut/base.py
```diff
@@ -506,6 +506,97 @@ def trim_to_supervisions(
             cuts.append(trimmed)
         return CutSet.from_cuts(cuts)
 
+    def trim_to_alignments(
+        self,
+        type: str,
+        max_pause: Seconds = 0.0,
```
pzelasko, line 512: Can we make it `Optional[Seconds] = None` and if it's `None` set max_pause = -1 so that it splits all alignment items without merging, and when it's 0.0 it merges the consecutive ones? I think it's more transparent from the usage perspective.

pzelasko (commented):
Thanks! Looks good to me, just left one comment.

lhotse/cut/base.py
desh2608: That makes sense. I'll change it.

desh2608:
> Thanks! Looks good to me, just left one comment.

Done.

PR #927, opened 2022-12-15
Fix serialization for reverb

desh2608:
The changes in https://github.com/lhotse-speech/lhotse/pull/920 introduced a bug in serialization/deserialization of `ReverbWithImpulseResponse` transform when used with an RIR generator. This PR fixes that bug and adds a test case for it.


pzelasko (approved):
Thanks!

Issue #891, opened 2022-11-15
Bug when transcribing audio with Whisper


SGombert:
Hi there, 

I am currently working on a project where we first transcribe audio recordings with Whisper and then force align them with Wav2Vec. 

When I just run the cuts generated by the method "annotate_with_whisper" into the method "align_with_torchaudio", I get the following error:

AssertionError: We don't support forced alignment of cuts with overlapping supervisions (cut ID: 'A011221._001_TR6')

When I try to trim the supervisions, I run into another issue where suddenly supervisions with a negative duration appear.

So far, I could not find a fix myself. Do you have any suggestions?

vorpaladin:
I am having the same problem with force aligning Whisper output with wav2vec using annotate_with_whisper.  I'm getting a 14.4% failure rate due to this issue.  I ran two sets of 90 speech audio files and had 13 fail in this way from both sets.  The sets are disjoint, all files contain different speech.  Durations, noise, and accents vary for the failures, from 10 minute to 60 minute durations, and accents from different UK regions.  I examined each audio file that resulted in failure and did not discern any obvious commonality between them.  I am using Whisper settings device="cuda", language="en" and model_name="small.en".

pzelasko:
Can you show an example of a cut with overlapping supervisions? Maybe that would help trace it if it's a bug in how we parse Whisper or something else.

mattzque:
I might be seeing the same or similar issues (I think) but perhaps I'm doing something wrong.

Code:

```python
from lhotse import (
    CutSet,
    Recording,
    RecordingSet,
    align_with_torchaudio,
    annotate_with_whisper,
)
from pathlib import Path
from pprint import pprint

recording = Recording.from_file(Path("./bug2.wav"))
recordings = RecordingSet.from_recordings([recording])

cuts = annotate_with_whisper(recordings, model_name="tiny", device="cuda")

cuts_aligned = align_with_torchaudio(cuts)

next(cuts_aligned)
```

Error:

```
ValueError: IntervalTree: Null Interval objects not allowed in IntervalTree: Interval(171.0, 171.0, SupervisionSegment(id='bug2-000058', recording_id='bug2', start=171.0, duration=0.0, channel=0, text='Yeah.', language='en', speaker=None, gender=None, custom=None, alignment=None))
py
```

Added logging of the cut, here is the full output: https://gist.github.com/mattzque/0e048e68ff5a0034a8c7d21511d2e2b0

[...]

PR #928, opened 2022-12-16
Remove negative duration segments from whisper

desh2608:
This PR addresses https://github.com/lhotse-speech/lhotse/issues/891.

* Remove segments with non-positive duration from the whisper output
* Segment post-processing to force non-overlapping is made optional (disabled by default)
* Allow overlapping segments in forced alignment workflow


pzelasko (approved):

pzelasko:
Thanks!

PR #941, opened 2023-01-15
Safe extract for more recipes

desh2608:
See https://github.com/lhotse-speech/lhotse/pull/939#issuecomment-1382974291.


pzelasko (approved):
Thanks!

PR #947, opened 2023-01-18
Batch extraction for kaldi features

desh2608:
This PR adds `extract_batch()` method for the kaldi compatible features implemented in Lhotse.

NOTE: We rename Torchaudio's `Spectrogram` feature to `TorchaudioSpectrogram` to be consistent with `TorchaudioMfcc` and `TorchaudioFbank`.

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -1984,6 +1984,7 @@ class CutSet(Serializable, AlgorithmMixin):
         batch_duration: Seconds = 600.0,
         num_workers: int = 4,
         collate: bool = False,
+        num_buckets: int = 20,
         augment_fn: Optional[AugmentFn] = None,
         storage_type: Type[FW] = LilcomChunkyWriter,
         overwrite: bool = False,
@@ -2029,6 +2030,9 @@ class CutSet(Serializable, AlgorithmMixin):
             padded tensor before being passed to the feature extractor. Some extractors
             can be faster this way (for e.g., see ``lhotse.features.kaldi.extractors``).
             If you are using ``kaldifeat`` extractors, you should set this to ``False``.
+        :param num_buckets: If ``collate`` is ``True``, we will use a bucketing sampler
+            to group similar-length waveforms into batches. This parameter controls
+            the number of buckets.
         :param augment_fn: an optional callable used for audio augmentation.
             Be careful with the types of augmentations used: if they modify
             the start/end/duration times of the cut and its supervisions,
@@ -2044,7 +2048,11 @@ class CutSet(Serializable, AlgorithmMixin):
         import torch
         from torch.utils.data import DataLoader
 
-        from lhotse.dataset import SingleCutSampler, UnsupervisedWaveformDataset
+        from lhotse.dataset import (
+            BucketingSampler,
+            SimpleCutSampler,
+            UnsupervisedWaveformDataset,
+        )
         from lhotse.qa import validate_features
 
         frame_shift = extractor.frame_shift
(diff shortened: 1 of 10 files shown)


pzelasko (approved):
Thanks! LGTM once the tests pass. You might need to try tweaking some settings, maybe the configurations for both extractors in the test are not identical.

test/features/test_kaldi_features.py
```diff
@@ -77,8 +99,25 @@ def test_kaldi_mfcc_extractor(recording):
...
     torch.testing.assert_allclose(feats, feats_ta)
+
+
+def test_kaldi_spectrogram_extractor(recording):
+    spec = Spectrogram()
+    feats = spec.extract(recording.load_audio(), recording.sampling_rate)
+    assert feats.shape == (1604, 257)
+
+
+def test_kaldi_spectrogram_extractor_vs_torchaudio(recording):
```
desh2608, line 115: This test is currently failing with the following message:
```python
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 3874 / 412228 (0.9%)
E       Greatest absolute difference: 22.66506841033697 at index (826, 0) (up to 1e-05 allowed)
E       Greatest relative difference: 17694.401589361405 at index (1029, 0) (up to 0.0001 allowed)
```

desh2608: On changing the assertion to `torch.testing.assert_allclose(feats, np.exp(feats_ta))`, I get:
```python
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 1604 / 412228 (0.4%)
E       Greatest absolute difference: 8.267975032700633 at index (325, 0) (up to 1e-05 allowed)
E       Greatest relative difference: 0.999999999856555 at index (826, 0) (up to 0.0001 allowed)
```
1604 is the number of frames, and it seems they only differ in the 0th dimension.

desh2608: Ahh, the first dimension is the log energy for both, so we don't need to exponentiate that for comparing.

desh2608:
> Thanks! LGTM once the tests pass. You might need to try tweaking some settings, maybe the configurations for both extractors in the test are not identical.

@pzelasko Made a small fix to the test and now it's passing. Ready to merge.

pzelasko (approved):
LGTM - please do the honors :)

desh2608:
@pzelasko I noticed some room for optimization in the batched feature extraction. Before this PR, we have 2 extractors which support batch extraction: kaldifeat and S3PRL. Both of these take a list of tensors and perform padding/collation internally, so we don't need to perform collation before-hand.

In this PR, we add batch extraction method for the `nn.Module` based Kaldi-compatible extractors (originally implemented by Jesus). These extractors take a batch of padded/collated samples as input. An easy solution is to pad and batch the list of tensors before passing them to the extractor. 

However, recall that the `extract_batch()` method is most often used from `CutSet.compute_and_store_features_batch()`, which uses PyTorch dataloaders to create batches in the background. Therefore, we add a `collate` option to this method (set as `False` by default), which specifies whether to collate the batch of samples. This can be set to `True` for the extractors which work with batched tensors, so that no additional GPU time is spent on collation.

Another small optimization is to off-load the task of saving the extracted features to a background thread.

I will write some tests for the newly added features if they seem reasonable to you.

lhotse/cut/set.py
```diff
@@ -2087,65 +2155,10 @@ def compute_and_store_features_batch(
...
-                                fastcopy(s, recording_id=cut.id)
-                                for s in cut.supervisions
-                            ],
-                            features=feat_manifest,
-                            recording=None,
-                        )
-                    cuts_writer.write(cut, flush=True)
 
+                futures.append(executor.submit(_save_worker, cuts, features))
                 progress.update(len(cuts))
```
desh2608, line 2163: Something that I find weird is that this works even though I have not called `future.result()` anywhere?

pzelasko: When the executor is called as a context manager it blocks the execution on `__exit__` until all threads have finished (calls `.join()`)

desh2608: Makes sense, thanks!

desh2608:
> Sounds good to me! Once you finish I think I will use your changes to build an example how to use Lhotse Shar (so the data will be stored sequentially in shards) for re-storing recordings and then computing the features. I am postponing writing the tutorials for quite a while now but that data format works fantastically in real world settings.

Awesome! I'm mostly done with this PR now, but I will sit on it until tomorrow perhaps and use the new batch extractors on some of my tasks to make sure everything works seamlessly. I will ping you when I think it is ready to review/merge.

desh2608:
@pzelasko I think this can be merged now.

pzelasko:
Thanks!!

pzelasko (approved):

PR #950, opened 2023-01-19
Optional quadratic duration correction for dynamic bucketing sampler

pzelasko:
I've also removed some deprecated options like specifying max frames/samples and related tests.

### DIFF
--- a/lhotse/dataset/sampling/base.py
+++ b/lhotse/dataset/sampling/base.py
@@ -295,74 +295,65 @@ class CutSampler(Sampler, Dillable):
 class TimeConstraint:
     """
     Represents a time-based constraint for sampler classes.
-    It can be defined either as maximum total batch duration (in seconds),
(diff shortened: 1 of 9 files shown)


pzelasko:
I gave it some runs, it seems to work alright, even though I still haven't calibrated it very well. 

As a piece anegdotic evidence, I was able to adjust a training which had `max_duration=250` into `max_duration=450` with `quadratic_duration=30` without OOM. So I almost doubled the duration, but the average duration in my dataset is about 5-6 seconds, which is far below the penalty that effectively doubles a cut duration. It's hard to give effective increase in throughput without measuring it carefully, but I think throwing around numbers like 20-30% is not unreasonable.

I'll merge it. If somebody uses it, it's important to remember that you need to adjust max_duration after setting quadratic_duration.

danpovey:
That's great!!

PR #952, opened 2023-01-20
Handling <end-time> of -1 in the segments file

JinZr:
Update kaldi.py to support <end-time> == -1 in the segments file. 

> \<end-time\> of -1 means the segment runs till the end of the WAV file
> https://kaldi-asr.org/doc/extract-segments_8cc.html


desh2608:
LGTM. Please fix the style check issue.

desh2608 (approved):

PR #958, opened 2023-01-26
Allow extend_by to pad arrays and features

pzelasko:
@desh2608 could you please review this change as you're well familiar with this code? The main point is to allow padding other fields than `recording` when using `extend_by` rather than just discarding the data.

### DIFF
--- a/lhotse/cut/data.py
+++ b/lhotse/cut/data.py
@@ -570,6 +570,8 @@ class DataCut(Cut, metaclass=ABCMeta):
         direction: str = "both",
         preserve_id: bool = False,
         pad_silence: bool = True,
+        pad_feat_value: float = LOG_EPSILON,
+        pad_value_dict: Optional[Dict[str, Union[int, float]]] = None,
     ) -> Cut:
         """
         Returns a new Cut (DataCut or MixedCut) that is an extended region of the current DataCut by extending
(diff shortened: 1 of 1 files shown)


desh2608 (approved):
LGTM.

Regarding the `pad_feat_value`, I was wondering if a more scalable approach would be for the `Features` class to have a `default_pad_value` property which is defined for each feature (for example, it may be log epsilon for logarithmic features or epsilon for STFT features, etc.). We can then use this default value for padding purposes instead of having users manually specify it. I faced some issues related to this recently when mixing STFT features.

pzelasko:
> Regarding the `pad_feat_value`, I was wondering if a more scalable approach would be for the `Features` class to have a `default_pad_value` property which is defined for each feature (for example, it may be log epsilon for logarithmic features or epsilon for STFT features, etc.). We can then use this default value for padding purposes instead of having users manually specify it. I faced some issues related to this recently when mixing STFT features.

I agree, but the implementation should be backward compatible (i.e. allow to correctly read manifests before this change).

desh2608:
> > Regarding the `pad_feat_value`, I was wondering if a more scalable approach would be for the `Features` class to have a `default_pad_value` property which is defined for each feature (for example, it may be log epsilon for logarithmic features or epsilon for STFT features, etc.). We can then use this default value for padding purposes instead of having users manually specify it. I faced some issues related to this recently when mixing STFT features.
> 
> I agree, but the implementation should be backward compatible (i.e. allow to correctly read manifests before this change).

I think that should be doable. I'll make a PR later.

pzelasko:
You know what, I will close this one for now. There is an edge case of:

```
|-------recording-------|
  |----features---|
  |-----cut-------|
 |---extended-cut---|
```

with `pad_silence=False` where we don't want to pad so we'd have to discard or raise error, and I don't really need these changes right now on a second thought. We can revisit if it's needed...

PR #961, opened 2023-01-30
Add option to create segments for LibriCSS

desh2608:
The LibriCSS dataset comes with official ~1-min segments obtained by splitting the 10-min recordings at long silences. We add this an option to prepare this segmented data in the recipe.

Possible use case: evaluation of multi-talker ASR (e.g. https://arxiv.org/abs/2109.08555).


pzelasko:
Thanks!

PR #967, opened 2023-02-07
Allow dash in SequentialJsonlWriter

desh2608:
This makes piping possible for several CLI binaries such as `trim-to-supervisions` etc. An example piped workflow for combining LibriSpeech train set, cutting into shorter segments, and shuffling:

```
lhotse combine data/manifests/librispeech_cuts_train* - |\
  lhotse cut trim-to-alignments --type word --max-pause 0.2 - - |\
  shuf | gzip -c > data/manifests/librispeech_cuts_train_trimmed.jsonl.gz
 ```


pzelasko:
And so we have went full cycle bash -> python -> bash 😄

Thanks LGTM!

PR #972, opened 2023-02-08
Add multiprocessing to meeting simulation workflow

desh2608:
## Changes in this PR

* Refactor meeting simulation workflow to put utterance group creation as a common sampler.
* Add option for using multiple workers to mix the sampled utterance groups. Can potentially speed up ~2-3x.
* Fix `parallel_map` method.

## Recommendations for `num_jobs`

When using the `simulate()` method in the workflow. we recommend using 1 job when the number of source utterances is small (up to 50k, for example). For larger inputs, this number can be scaled up slowly, but not more than 4-8 jobs to avoid slow-down due to multiprocessing overhead.

### DIFF
--- a/lhotse/workflows/meeting_simulation/base.py
+++ b/lhotse/workflows/meeting_simulation/base.py
@@ -145,12 +145,15 @@ class MeetingSampler:
         # sampling items from a set is O(n).
         self.samplers = {}
         for spk, spk_cuts in tqdm(
-            groupby(cuts, lambda cut: cut.supervisions[0].speaker),
+            groupby(
+                sorted(cuts, key=lambda cut: cut.supervisions[0].speaker),
+                lambda cut: cut.supervisions[0].speaker,
+            ),
             desc="Creating samplers for each speaker...",
         ):
             sampler = DynamicCutSampler(
                 CutSet.from_cuts(list(spk_cuts)).repeat(
-                    times=num_repeats, preserve_id=False
+                    times=num_repeats, preserve_id=True
                 ),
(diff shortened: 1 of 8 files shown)


pzelasko (dismissed):
Very cool!

desh2608:
> Very cool!

I'll keep this under WIP while I test it out on my actual simulation tasks.

lhotse/workflows/meeting_simulation/base.py
```diff
@@ -83,6 +89,120 @@ def reverberate(self, cuts: CutSet, *rirs: RecordingSet) -> CutSet:
...
+            len(self.samplers),
+        )
+
+        # Sample speakers.
+        this_batch_spk_ids = self.rng.sample(list(self.samplers.keys()), N)
+        utterances = CutSet.from_cuts([])
+        for spk_id in this_batch_spk_ids:
+            sampler = self.samplers[spk_id]
+            try:
+                this_batch = next(sampler)
```
desh2608, line 195: @pzelasko An issue I am facing here is that the whole sampler gets exhausted after sampling just 1 batch. I'm not sure why this is happening. Could you take a look?

desh2608: Okay very basic mistake --- forgot to sort the cuts by speaker id before groupby!

desh2608:
The simulation should be quite fast now. As an example, creating mixtures from ~3.2M source utterances (each used once) takes ~1h using 4 jobs.

pzelasko (approved):

PR #974, opened 2023-02-09
Fix bug in computing same speaker pause distribution

desh2608:
We should only consider **adjacent** same-speaker segments when computing the distribution.


pzelasko (approved):
Thanks!

PR #978, opened 2023-02-13
Fix forced alignment for multi supervisions

pkufool:


lhotse/workflows/forced_alignment.py
```diff
@@ -59,7 +59,7 @@ def align_with_torchaudio(
     for cut in cuts:
 
         for idx, subcut in enumerate(cut.trim_to_supervisions(keep_overlapping=False)):
-            sup = subcut.supervisions[0]
+            sup = subcut.supervisions[idx]
```
desh2608, line 62: Is this correct? Once we have trimmed to supervisions, each subcut will have exactly 1 supervision, so I think we should index it by 0, no?

pkufool: Yes, you are right! I read the documents again, it should be 0.  What I want to do is adding a bpe supervisions to each cut and get its alignments, as follows:
```
cut_list = []
for c in cuts:
    c.supervisions.append(copy.deepcopy(c.supervisions[0]))
    c.supervisions[1].text = " ".join(
        sp.encode(c.supervisions[0].text, out_type=str)
    )
    cut_list.append(c)
cutset = CutSet.from_cuts(cut_list)
```
I think the problem is I used the same id for the bpe supervision, so the `trim_to_supervisions` does not work as expected.

desh2608: So you want each cut to have both word-based and BPE-based supervisions? I think there are 2 may be 2 ways to go about this:
1. If you have the start/end times of each BPE token, you can add the BPE token sequence as an `alignment` of the supervision.
2. Otherwise, I think you should create a new CutSet with the same ids as the original cuts but with supervisions replaced by the BPE-based supervision. You could then use the same cut ids to index both cut sets, or use things like `CutPairsSampler` if you want to use them for training etc. Here is a one-liner to create the new cuts.
```python
new_cuts = CutSet.from_cuts(
    [
        fastcopy(
            c,
            supervisions=[
                fastcopy(s, text=" ".join(sp.encode(s.text, out_type=str)))
                for s in c.supervisions
            ],
        )
        for c in cuts
    ]
)
```

pkufool: Thanks! Actually, I have BPE token sequence and want to use the forced-alignment tool to get the start/end times of each BPE token.  If my trick won't work, I will try your suggestions, Thank you!

pzelasko: Maybe it would make sense to wrap one of Icefall's BPE CTC models in a similar way as torchaudio's Wav2Vec2 character-level CTC, and create a new workflow (e.g. `align-with-icefall`)? I don't have the time currently to do it but maybe somebody else is interested and could pick it up.

PR #980, opened 2023-02-16
Add padding direction when using transform ExtraPadding

marcoyang1998:
This PR adds the padding direction as an option when using transform `ExtraPadding`.

The original implementation only supports padding on both sides. Now users can select from "left", "right" or "both".


pzelasko:
Please fix the failing formatting check and LGTM!

marcoyang1998:
Just did it

pzelasko:
Thanks!

PR #985, opened 2023-02-21
Add RecordingChunkIterableDataset

### LINKED ISSUE #983: Large audio files processing
Piotr, 
I want to do some transcription of large audio files for some data-prep and I'm wondering the best way to do the processing.
These files may be too long to keep the audio in memory at one time, i.e. may be better to process them streaming or at least in chunks if there are efficient ways to process long compressed audios in chunks.

pzelasko:
CC @danpovey this is what I suggest to efficiently load chunks of long recordings. I just need to add support for chunk_shift to fit your requirements. 

Resolves #983


pzelasko:
It scales pretty much linearly, on a local SSD I'm loading 4h+ of audio in one second with 4 workers, not a bad result...

image

PR #996, opened 2023-03-15
Speech Commands v0.01 & v0.02 dataset

yfyeung:
Speech Commands is an audio dataset of spoken words designed to help train and evaluate keyword spotting systems.
paper: https://arxiv.org/pdf/1804.03209.pdf

### DIFF
--- a/lhotse/recipes/speechcommands.py
+++ b/lhotse/recipes/speechcommands.py
@@ -41,18 +41,7 @@ _TEST_DOWNLOAD_PATH_V2 = (
 
 _SPLITS = ["train", "valid", "test"]
 
-WORDS = [
-    "down",
-    "go",
-    "left",
-    "no",
-    "off",
-    "on",
-    "right",
-    "stop",
-    "up",
-    "yes"
-]
+WORDS = ["down", "go", "left", "no", "off", "on", "right", "stop", "up", "yes"]
 SILENCE = "_silence_"
 UNKNOWN = "_unknown_"
 BACKGROUND_NOISE = "_background_noise_"
@@ -82,8 +71,7 @@ def _download_speechcommands(
     )
 
     for part in tqdm(
-        dataset_parts,
-        desc=f"Downloading Speech Commands v0.0{speechcommands_version}"
+        dataset_parts, desc=f"Downloading Speech Commands v0.0{speechcommands_version}"
     ):
         logging.info(f"Processing split: {part}")
         part_dir = corpus_dir / part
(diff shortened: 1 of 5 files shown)


desh2608:
Perhaps the recipe should be called "speech_commands" instead of "speech_commands001", and the version should be provided as a parameter for download/prepare.

yfyeung:
> Perhaps the recipe should be called "speech_commands" instead of "speech_commands001", and the version should be provided as a parameter for download/prepare.

Ok, I will implement this.

desh2608:
Thanks! You can look at the VoxCeleb recipe for an example where we support v1 and v2.

yfyeung:
@desh2608 Hi, I have finished this recipe, supporting Speech Commands v0.01 and v0.02 dataset.
Hope for your review of this data preparation part.

lhotse/recipes/speechcommands.py
```diff
@@ -0,0 +1,465 @@
...
+    force_download: bool = False,
+) -> Path:
+    """
+    Download and unzip the Speech Commands v0.02 data.
+    :param target_dir: Pathlike, the path of the dir to store the dataset.
+    :param force_download: bool, if True, download the archive even if it already exists.
+    :return: the path to downloaded and extracted directory with data.
+    """
+
+    return _download_speechcommands(
```
pzelasko, line 129: Do we really need v1 and v2 to have separate download functions even though they share the same logic? I think it's cleaner to rename `_download_speechcommands` into `download_speechcommands` and set the default version to the latest one, and then expose the version argument in the CLI (so that we have one CLI program for this rather than two).

lhotse/recipes/speechcommands.py
```diff
@@ -0,0 +1,465 @@
...
+        speechcommands_version="2",
+        target_dir=target_dir,
+        force_download=force_download,
+    )
+
+
+def _prepare_train_valid(
```
pzelasko, line 136: could you split this function into separate prepare_train and prepare_valid? I don't think you really need a generator here if you restructure the code, and the way to use this as-is is a little confusing.

pzelasko (commented):
Thanks, I left two comments. Can you address them before we merge?

lhotse/recipes/speechcommands.py
yfyeung: OK, I will split it.

lhotse/recipes/speechcommands.py
yfyeung: OK.

pzelasko:
LGTM!

PR #998, opened 2023-03-16
Aishell3

zjwang21:
Dataset download and prepare code supported for aishell3.


pzelasko (commented):
Looks good, can you update the imports in `lhotse/recipes/__init__.py` and the table in `docs/corpus.rst` as well?

PR #999, opened 2023-03-16
Changing devices in Fbank

Tomiinek:


lhotse/features/kaldi/extractors.py
```diff
@@ -76,6 +76,10 @@ def device(self) -> Union[str, torch.device]:
     def frame_shift(self) -> Seconds:
         return self.config.frame_shift
 
+    def to(self, device: str):
+        self.config.device = device
+        self.extractor.to(device)
```
Tomiinek, line 81: Because I would like to instantiate extractors using `.from_yaml(...)` from the same config file, but use them on CPUs and GPUs.

lhotse/features/kaldi/extractors.py
```diff
@@ -403,7 +406,9 @@ def _extract_batch(
         samples = torch.nn.utils.rnn.pad_sequence(samples, batch_first=True)
 
     # Perform feature extraction
-    feats = extractor(samples.to(device)).cpu()
```
Tomiinek, line 406: Since I would expect the output to be on the same device as the input.

lhotse/features/kaldi/extractors.py
```diff
@@ -385,7 +389,6 @@ def _extract_batch(
             samples = [samples.reshape(1, -1)]
 
         if any(isinstance(x, torch.Tensor) for x in samples):
-            samples = [x.numpy() for x in samples]
```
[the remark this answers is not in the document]
Tomiinek: I think that a usecase where the user inputs a list of mixture of numpy arrays and tensors is paranoid :smile:

PR #1004, opened 2023-03-20
Add cache for KaldiReader

david20181:
Make KaldiReader faster by using cache of readers

### DIFF
--- a/lhotse/features/io.py
+++ b/lhotse/features/io.py
@@ -942,6 +942,26 @@ class LilcomURLWriter(FeaturesWriter):
 Kaldi-compatible feature reader
 """
 
+def check_kaldi_native_io_installed():
+    if not is_module_available("kaldi_native_io"):
+        raise ValueError(
+            "To read Kaldi feats.scp, please 'pip install kaldi_native_io' first."
+        )
+
+
+@lru_cache(maxsize=None)
+def lookup_reader_cache_or_open(storage_path: str):
+    """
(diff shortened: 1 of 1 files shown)


lhotse/features/io.py
```diff
@@ -942,6 +942,26 @@ def write(self, key: str, value: np.ndarray) -> str:
...
 """
 
+def check_kaldi_native_io_installed():
+    if not is_module_available("kaldi_native_io"):
+        raise ValueError(
+            "To read Kaldi feats.scp, please 'pip install kaldi_native_io' first."
+        )
+
+
+@lru_cache(maxsize=None)
```
pzelasko, line 952: Can you also update

https://github.com/lhotse-speech/lhotse/blob/faddd86d43f247f227ce581a7630c4ce74084c3e/lhotse/features/io.py#L389

to clear the objects cached by this function?

Also maybe I'm wrong but I think we once supported caching for kaldi_io/kaldi_native_io but for very large Kaldi data dirs it took a lot of memory.

david20181: Good suggestion on clearing cache. I updated accordingly.

I've tested on some fairly large data dirs and have not observed memory issue.

pzelasko: Thanks for confirming that! LGTM, please just fix the formatting.

pzelasko:
Thanks!

PR #1005, opened 2023-03-21
Fix load_kaldi_dara_dir not loading segments and feats.scp correctly

### LINKED ISSUE #987: load_kaldi_data_dir with segments
Hello, when loading a kaldi data dir including `feats.scp` and `segments` into CutSet, what would be the best approach?

I used `lhotse.kaldi.load_kaldi_data_dir` to load a kaldi data dir into `recordings`, `supervisions` and `features`, and then creating a CutSet from `CutSet.from_manifests` fails to find a supervision corresponding to each cut. 

Since each supervision inherits its start time from kaldi `segments`, `create_cut_set_eager` cannot match supervisions with their corresponding cut: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/cut/set.py#L2894-L2899

I got around this by forcing a supervision start and duration to be cut start and duration, but I'm sure there is a cleaner approach to doing this, or I'd appreciate it if you could let me know if I'm missing a better way to load features and supervisions from kaldi data dir.

I'll leave my workaround way to load a kaldi data dir.

``` python
recs, sups, feats = lhotse.kaldi.load_kaldi_data_dir(
        args.kaldi_dir,
        sampling_rate=args.sampling_rate,
        frame_shift=args.frame_shift,
    )

# here, matching supervisions with cuts fails
cutset = lhotse.CutSet.from_manifests(
        recordings=recs,
        supervisions=sups
        features=feats,
    )

sup_dict = {}
for sup in sups:
    sup_dict[sup.id] = sup

for cut in cutset:
    storage_key = cut.features.storage_key
    # match supervision with cut by features.storage_key
    supervision = sup_dict[storage_key]
    # if duration is too different between supervision cut,
    # something wrong is happening
    sup_dur = supervision.duration
    if sup_dur - cut.duration > 0.05:
        raise ValueError("supervision duration and cut duration is too different.")

    # discard supervision.start inheritated from segments
    # and assign cut.duration and cut.start to supervision
    supervision.duration = cut.duration
    supervision.start = cut.start

    cut.supervisions.append(supervision)
```

yasumori:
Fix #987

* I'd change the script to store `mat_shape` in `utt_id_to_start_and_duration` if calling `kaldi_native_io.MatrixShape.read(ark)` twice is an overhead.
* Even nicer is probably to consolidate information from kaldi data files (segments, utt2spk, feats.scp) in a sort of data structure and then create Recordings, Supervisions and Features but I didn't have time to do this. 
* The reason why Supervisions need to inherit duration computed from a feature matrix is that the validation in https://github.com/lhotse-speech/lhotse/blob/master/lhotse/qa.py#LL298C3-L298C3 would fail if duration is inherited from segments.


desh2608:
I think this is fine as a temporary solution, but as you mentioned, the `kaldi.py` script is getting a little cluttered, and too many if-else inside the functions. There are some issues we may run into in the future, such as: if the feature duration is very different from the segment duration (e.g. more than "off by 1") we may want to discard such segments, or trim either the segment or the feature, or whatever. Otherwise it could cause mismatch errors when trying to use the corresponding cuts. Such issues can be elegantly handled by creating an intermediate data structure as you mentioned, and providing some sensible defaults.

desh2608:
Please fix the style issues.

jtrmal:
looks ok but it would be add test case for the functionality

pzelasko:
@yasumori could you add a unit test as well? We have some tests for Kaldi export/import so hopefully it wouldn't be excessively difficult.

yasumori:
@pzelasko Yes, I'll push a new commit adding a test soon

pzelasko (approved):
Thanks, LGTM

jtrmal:
thanks @yasumori

PR #1007, opened 2023-03-22
Fix minor bug in conversational meeting simulation algorithm

desh2608:
* We use `add_durations` to make the start/end times of mixed in segments more precise.
* There was a minor bug earlier which was causing 3-fold overlaps even when set to `False`. This is fixed now.


pzelasko (approved):
LGTM, thanks!

PR #1012, opened 2023-03-28
Small fix for speechcommands.py

yfyeung:
In the initial version, I assigned the value `None` to the `text` parameter in SupervisionSegment for instances labeled BACKGROUND_NOISE or SILENCE, which could cause an error in https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/speech_recognition.py#L138.

The None type is not compatible with the default_collate(), so I replace the value `None` with "" in this PR.


pzelasko:
Thanks!

PR #1013, opened 2023-03-28
Specify what formats are expected when using the pipe: prefix to load cuts

srdecny:
When using the `pipe:` command to load data, the "cuts" field expects a `.jsonl` stream and the other fields expect a `.tar` stream. The example in the current documentation is a bit misleading, because it passes a`.jsonl.gz` stream, which will not work.

This PR attempts to clarify this in the documentation.


pzelasko:
Thanks!

PR #1014, opened 2023-03-30
LibriLight dataset

yfyeung:
This recipe enables preparing LibriLight manifests. 
Since LibriLight is unlabelled, no text field exists in `SupervisionSegment`.

Usage:
```shell
lhotse prepare librilight <librilight_dir> <output_dir>  -j <num_jobs>
```

### DIFF
--- a/lhotse/recipes/librilight.py
+++ b/lhotse/recipes/librilight.py
@@ -21,7 +21,7 @@ from lhotse.utils import Pathlike
 _SPLITS = ["small", "medium", "large"]
 
 
-def _prepare_split(
+def _prepare_subset(
     subset: str,
     corpus_dir: Pathlike,
 ) -> Tuple[RecordingSet, SupervisionSet]:
@@ -111,7 +111,7 @@ def prepare_librilight(
             logging.info(f"LibriLight subset: {part} already prepared - skipping.")
             continue
 
-        recording_set, supervision_set = _prepare_split(part, corpus_dir)
+        recording_set, supervision_set = _prepare_subset(part, corpus_dir)
 
         if output_dir is not None:
             supervision_set.to_file(
--- a/lhotse/recipes/librilight.py
+++ b/lhotse/recipes/librilight.py
@@ -1,0 +1,127 @@
+"""
+About the librilight corpus
+Libri-light is a benchmark for the training of automatic speech recognition (ASR) systems with limited or no supervision.
+It contains a large dataset of 60K hours of unlabelled speech from audiobooks in English and a small labelled dataset (10h, 1h, and 10 min) plus metrics, trainable baseline models, and pretrained models that use these datasets.
+It is covered in more detail at https://arxiv.org/abs/1912.07875.
+"""
+
+import logging
+import os
+from collections import defaultdict
+from pathlib import Path
+from typing import Dict, List, Optional, Sequence, Tuple, Union
+
+from tqdm.auto import tqdm
+
+from lhotse.audio import Recording, RecordingSet
+from lhotse.recipes.utils import manifests_exist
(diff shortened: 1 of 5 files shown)


lhotse/recipes/librilight.py
```diff
@@ -0,0 +1,124 @@
...
+        if not audio_path.is_file():
+            logging.warning(f"No such file: {audio_path}")
+            continue
+
+        recording = Recording.from_file(
+            path=audio_path,
+            recording_id=file_name,
+        )
+        recordings.append(recording)
+        segment = SupervisionSegment(
```
desh2608, line 61: FWIW, `Recording` class has a `to_cut()` method which creates a cut out of the recording, so it may be okay to just return a `RecordingSet` from LibriLight (like we do with MUSAN). I guess we would miss explicit speaker id information though, so this is fine too.

lhotse/recipes/librilight.py
```diff
@@ -0,0 +1,124 @@
...
+        supervisions.append(segment)
+    recording_set = RecordingSet.from_recordings(recordings)
+    supervision_set = SupervisionSet.from_segments(supervisions)
+
+    return recording_set, supervision_set
+
+
+def prepare_librilight(
+    corpus_dir: Pathlike,
+    output_dir: Optional[Pathlike] = None,
```
desh2608, line 80: Since this is a large corpus, perhaps it would be useful to add a `num_jobs` option to speed up manifest creation? Check the LibriSpeech recipe for an example.

lhotse/recipes/librilight.py
```diff
@@ -0,0 +1,124 @@
+"""
+About the librilight corpus
+Libri-light is a benchmark for the training of automatic speech recognition (ASR) systems with limited or no supervision.
+It contains a large dataset of 60K hours of unlabelled speech from audiobooks in English and a small labelled dataset (10h, 1h, and 10 min) plus metrics, trainable baseline models, and pretrained models that use these datasets.
+It is covered in more detail at https://arxiv.org/abs/1912.07875.
+"""
```
pzelasko, line 6: If we're not adding download functions, could you at least provide a link to where this dataset can be obtained from?

lhotse/recipes/librilight.py
yfyeung: OK

lhotse/recipes/librilight.py
yfyeung: Sure.

lhotse/recipes/librilight.py
yfyeung: OK, I will avoid setting it.

yfyeung:
@pzelasko I have addressed them, please re-check it.

lhotse/recipes/librilight.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Returns the RecodingSet and SupervisionSet given a dataset part.
+    :param subset: str, the name of the subset.
+    :param corpus_dir: Pathlike, the path of the data dir.
+    :return: the RecodingSet and SupervisionSet for train and valid.
+    """
+    corpus_dir = Path(corpus_dir)
+    part_path = corpus_dir / subset
+    audio_paths = []
+    for root, dirs, files in os.walk(part_path):
```
desh2608, line 82: Can this loop be turned into a one-liner with `Path(part_path).rglob("*.flac")`?

yfyeung: Sure.

yfyeung:
@pzelasko Please re-check the code, thx.

pzelasko (approved):
LGTM

Issue #966, opened 2023-02-02
Loudness normalisation


saveli:
Is there a way to do audio normalisation on-the-fly on file-level (`Recording.from_file()`) using `AudioTransform`?

Reason is, when using `annotate_with_whisper` in `SupervisionSegment` sometimes `alginment=None` despite `text` being not `None`.
After normalising the file manually (ffmpeg & loudnorm), the error was gone.

Hence, it is still loud enough for whisper to find text (speech prob > no speech threshold), but alignment fails.

desh2608:
@saveli you can consider adding https://github.com/csteinmetz1/pyloudnorm as an optional dependency and using their methods for normalization.

saveli:
@pzelasko I'm not that firm in lhotse's data processing behind the scenes. You would have to do a 2-pass (find peak & adjust volume) loudness normalisation.

Therefore, for now I will use @desh2608 suggestion, thanks.

PR #1016, opened 2023-04-05
Loudness normalization with `pyloudnorm`

desh2608:
Related to https://github.com/lhotse-speech/lhotse/issues/966.

This PR adds a method `normalize_loudness()` for recordings. This takes an argument `target` which specifies the desired loudness (usually around -23 dB is a good value). The implementation uses [pyloudnorm](https://github.com/csteinmetz1/pyloudnorm/tree/master).

Also, we move `ReverbWithImpulseResponse` out of the torchaudio.py and into its own file, since it is not a torchaudio based transform.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -36,6 +36,7 @@ from tqdm.auto import tqdm
 from lhotse.augmentation import (
     AudioTransform,
     DereverbWPE,
+    LoudnessNormalization,
     Resample,
     ReverbWithImpulseResponse,
     Speed,
@@ -704,6 +705,23 @@ class Recording:
         return fastcopy(
             self,
             id=f"{self.id}_vp{factor}" if affix_id else self.id,
+            transforms=transforms,
+        )
+
+    def normalize_loudness(self, target: float, affix_id: bool = False) -> "Recording":
+        """
+        Return a new ``Recording`` that will lazily apply WPE dereverberation.
+
+        :param target: The target loudness (in dB) to normalize to.
+        :param affix_id: When true, we will modify the ``Recording.id`` field
+            by affixing it with "_ln{factor}".
+        :return: a modified copy of the current ``Recording``.
+        """
+        transforms = self.transforms.copy() if self.transforms is not None else []
+        transforms.append(LoudnessNormalization(target=target).to_dict())
+        return fastcopy(
+            self,
+            id=f"{self.id}_ln{target}" if affix_id else self.id,
             transforms=transforms,
         )
 
--- a/lhotse/augmentation/loudness.py
+++ b/lhotse/augmentation/loudness.py
@@ -1,0 +1,72 @@
+import warnings
+from dataclasses import asdict, dataclass
+from typing import Optional, Tuple, Union
+
+import numpy as np
+import torch
+
+from lhotse.augmentation.transform import AudioTransform
+from lhotse.utils import Seconds, is_module_available
+
+
+@dataclass
+class LoudnessNormalization(AudioTransform):
+    """
+    Loudness normalization based on pyloudnorm: https://github.com/csteinmetz1/pyloudnorm.
+    """
+
+    target: float
+    sampling_rate: int = 16000
+
+    def __call__(
+        self, samples: Union[np.ndarray, torch.Tensor], *args, **kwargs
+    ) -> np.ndarray:
+        if torch.is_tensor(samples):
+            samples = samples.cpu().numpy()
+        augmented = normalize_loudness(samples, **asdict(self))
+        return augmented
+
+    def reverse_timestamps(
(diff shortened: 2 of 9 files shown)


lhotse/augmentation/loudness.py
```diff
@@ -0,0 +1,72 @@
...
+    meter = pyln.Meter(
+        sampling_rate, block_size=min(0.4, duration)
+    )  # create BS.1770 meter
+    loudness = meter.integrated_loudness(audio.T)
+
+    # loudness normalize audio to target LUFS. We will ignore the warnings related to
+    # clipping the audio.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        loudness_normalized_audio = pyln.normalize.loudness(audio.T, loudness, target)
```
pzelasko, line 70: If audio clipping is an issue here you can add a limiter as a post-processing step, e.g. https://github.com/pzelasko/cylimiter

pzelasko (approved):
Looks good to me, I left a single comment, up to you :)

lhotse/augmentation/loudness.py
desh2608: Sure, I can add it. But I don't have enough familiarity with limiters (or "loudness" for that matter) to know what to do exactly.

pzelasko: The defaults should "just work" with pretty much anything, it basically keeps track of the signal's loudness with a small lookahead and reduces the gain if it crosses some threshold. Think of it as a soft clipping that doesn't introduce as much distortion as hard clipping.

pzelasko: .. but again I don't know if that's a real problem with this approach and worth the extra dependency, so it's your call :)

desh2608: So far I have only used it to make the LibriCSS distant mic audio louder (it is at around -52 dB originally), and it sounds okay even with the clipping. I suppose we can let it be for now and add the limiter later if someone needs it?

lhotse/augmentation/loudness.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@dataclass
+class LoudnessNormalization(AudioTransform):
+    """
+    Loudness normalization based on pyloudnorm: https://github.com/csteinmetz1/pyloudnorm.
+    """
+
+    target: float
+    sampling_rate: int = 16000
```
lifeiteng, line 19: @pzelasko should we make `sampling_rate` as a member?

```
class AudioTransform:
    def __call__(self, samples: np.ndarray, sampling_rate: int) -> np.ndarray:
        """
        Apply transform.

        To be implemented in derived classes.
        """
        raise NotImplementedError
```

lhotse/audio.py
```diff
@@ -707,6 +708,23 @@ def perturb_volume(self, factor: float, affix_id: bool = True) -> "Recording":
...
+        """
+        Return a new ``Recording`` that will lazily apply WPE dereverberation.
+
+        :param target: The target loudness (in dB) to normalize to.
+        :param affix_id: When true, we will modify the ``Recording.id`` field
+            by affixing it with "_ln{factor}".
+        :return: a modified copy of the current ``Recording``.
+        """
+        transforms = self.transforms.copy() if self.transforms is not None else []
+        transforms.append(LoudnessNormalization(target=target).to_dict())
```
lifeiteng, line 721: `sampling_rate` is missed

lifeiteng:
@desh2608 @pzelasko 

I used set.normalize_loudness as blow: https://github.com/lifeiteng/vall-e/blob/main/valle/bin/tokenizer.py#L173
```
if args.prefix == "aishell":
    # NOTE: the loudness of aishell audio files is around -33
    # The best way is datamodule --on-the-fly-feats --enable-audio-aug
    cut_set = cut_set.normalize_loudness(
        target=-20.0, affix_id=True
    )

cut_set = cut_set.resample(24000)
```

But model's accuracy drops a lot.


截屏2023-04-14 23 44 31

desh2608:
@lifeiteng what metric is this curve? If it is validation accuracy, did you normalize loudness for both train and validation sets? I would think that for the training set, it may be better to "perturb" the loudness (within some range) rather than normalize it, similar to how we perturb volume.

lifeiteng:
@desh2608 Training CrossEntropy ArTop10Accuracy.
yes, normalized loudness for both train and validation sets.
It's used in text-to-speech, yes "perturb" is better, but I want to make it simple at current stage.
I don't understand why normalization leads to a significant drop in accuracy. The bug(comments above) is not triggered.

pzelasko:
@lifeiteng If it's for TTS, can you listen to the train and dev examples before and after normalization, as well to the model predictions? Maybe that could reveal if there's something funky going on.

PR #1017, opened 2023-04-06
[WIP] Add annotation workflow for faster-whisper (ctranslate2)

entn-at:
This PR adds a second Whisper annotation workflow that uses [faster-whisper](https://github.com/guillaumekln/faster-whisper/) powered by [CTranslate2](https://github.com/OpenNMT/CTranslate2)'s implementation (see https://github.com/entn-at/lhotse/tree/feature/whisper-ctranslate2). It's a lot faster and uses far less memory.

This implementation also obtains word start and end times. I'm still investigating whether they are accurate enough in general to be used as alignments.


desh2608:
Would it be possible to combine whisper and faster-whisper into a single CLI/method and then add faster-whisper as an optional flag enabled/disabled by default? The internals of the function call can be kept separate, but from a user perspective, it makes more sense since they have the same functionality. I'm thinking of it as 2 backends with the same user-facing wrapper.

entn-at:
Thanks for the quick initial review! I combined whisper and faster-whisper into a single CLI/method with a `--faster-whisper` flag. I also added additional off-by-default feature flags for faster-whisper:

- `--faster-whisper-add-alignments`: Whether to use faster-whisper's built-in method for obtaining word alignments (using cross-attention pattern and dynamic time warping; generally not as accurate as forced alignment).
- `--faster-whisper-use-vad`: Whether to apply speech activity detection (SileroVAD) before Whisper to reduce repetitions/spurious transcriptions (what is often referred to as "hallucinations").
- `--faster-whisper-num-workers`: Number of workers for parallelization across multiple GPUs.

**Quick benchmark on mini-librispeech dev-clean-2:**

_OpenAI Whisper_, RTX2080Ti:
```
$ time lhotse workflows annotate-with-whisper -n large-v2 -l en -m librispeech_recordings_dev-clean-2.jsonl.gz --device "cuda" librispeech_cuts_dev-clean-2.jsonl.gz
real    44m31.647s
user    46m5.540s
sys     0m10.869s
```
_faster-whisper/ctranslate2_, `float16` on RTX2080Ti:
```
time lhotse workflows annotate-with-whisper --faster-whisper -n large-v2 -l en -m librispeech_recordings_dev-clean-2.jsonl.gz --device "cuda" librispeech_cuts_dev-clean-2.jsonl.gz
real    18m15.743s
user    34m47.594s
sys     30m18.775s
```

`faster-whisper` allows parallelization across multiple GPUs. With `--faster-whisper-num-workers 4` across 4xRTX2080Ti:
```
$ time lhotse workflows annotate-with-whisper --faster-whisper -n large-v2 --faster-whisper-num-workers 4 -l en -m librispeech_recordings_dev-clean-2.jsonl.gz --device "cuda" librispeech_cuts_dev-clean-2.jsonl.gz
real    6m34.545s
user    35m50.779s
sys     25m48.421s
```

The only incompatibility with the current Whisper method is that faster-whisper doesn't expose a way to set the download location for the models. I submitted a [PR to faster-whisper](https://github.com/guillaumekln/faster-whisper/pull/126) and once that's merged/published in a new version, the currently [commented-out line 116 in faster_whisper.py](https://github.com/lhotse-speech/lhotse/blob/f4a28afb949497701e42af093993a0c12b2b4ff1/lhotse/workflows/faster_whisper.py#L116) can be changed to enable that.

lhotse/bin/modes/workflows.py
```diff
@@ -55,6 +56,36 @@ def workflows():
...
+@click.option(
+    "--faster-whisper",
+    is_flag=True,
+    default=False,
+    help="If True, use faster-whisper's implementation based on CTranslate2.",
+)
+@click.option(
+    "--faster-whisper-use-vad",
+    is_flag=True,
+    default=False,
```
pzelasko, line 68: Should we enable it as default / is there a reason not to?

pzelasko (commented):
Thanks for a great contribution, this is super interesting. The code looks good to me. I would love to enable it by default.

lhotse/bin/modes/workflows.py
```diff
@@ -55,6 +56,36 @@ def workflows():
 @click.option(
     "-d", "--device", default="cpu", help="Device on which to run the inference."
 )
+@click.option(
+    "--faster-whisper",
+    is_flag=True,
+    default=False,
```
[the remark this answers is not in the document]
entn-at: I think it should be fine to enable it as default, as long as it's documented.

lhotse/bin/modes/workflows.py
entn-at: It's useful for longer recordings with silence/noise portions, but can actually slow down processing of relatively short recordings (such as librispeech, which is already chunked). I assume the primary use case for this workflow is processing/annotating longer files, so it would make sense to enable it as default. If so, we should disable the `--faster-whisper-add-alignments` option when VAD is active, as it interacts badly with faster-whisper's word timestamps feature (see e.g. https://github.com/guillaumekln/faster-whisper/issues/125).

pzelasko: Sounds good to me! Can you make the relevant changes? 

Also maybe we can emit a warning that says what you just wrote if somebody turns on both VAD and alignment options.

lhotse/bin/modes/workflows.py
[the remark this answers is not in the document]
entn-at: Done

pzelasko:
I quickly compared the results between old and new whisper implementations on a 60s clip from AMI. In that clip, I noticed that faster-whisper tends to skip short, isolated, and noisy utterances such as "Okay" or "Thank you", probably due to VAD (which is OK I guess). However the time boundaries seem off when you compare it to the original implementation, please see the screenshot. Do you think it's possible to fix it? Maybe more accurate information is exposed somewhere in faster-whisper and it's just not being used here? Otherwise there's a lot of silence/non-speech included in the supervisions.

image

lhotse/bin/modes/workflows.py
```diff
@@ -55,6 +56,36 @@ def workflows():
 @click.option(
     "-d", "--device", default="cpu", help="Device on which to run the inference."
 )
+@click.option(
```
pzelasko, line 59: Please change to:

```python
@click.option(
    "--faster-whisper/--normal-whisper",
    default=True,
    help="If True, use faster-whisper's implementation based on CTranslate2.",
)
```

Otherwise it can't be turned off.

lhotse/bin/modes/workflows.py
```diff
@@ -55,6 +56,36 @@ def workflows():
...
+    "--faster-whisper-add-alignments",
+    is_flag=True,
+    default=False,
+    help="If True, add word alignments using timestamps obtained using the cross-attention"
+    "pattern and dynamic time warping (Note: Less accurate than forced alignment).",
+)
+@click.option(
+    "--faster-whisper-compute-type",
+    default="float16",
```
pzelasko, line 81:
```suggestion
    default="auto",
```

Otherwise it won't work on (some?) CPUs.

lhotse/workflows/faster_whisper.py
```diff
@@ -0,0 +1,244 @@
...
+        device_index = list(range(num_workers))
+    else:
+        device_index = 0
+    model = WhisperModel(
+        model_name,
+        device=device,
+        device_index=device_index,
+        compute_type=compute_type,
+        num_workers=num_workers,
+        download_root=download_root,
```
pzelasko, line 128: Since the change that enables this option is still not released in pip, I suggest a bit of workaround here, otherwise it cannot be ran:

```
+    opt_kwargs = {}
+    if download_root is not None:
+        opt_kwargs["download_root"] = download_root
     model = WhisperModel(
         model_name,
         device=device,
         device_index=device_index,
         compute_type=compute_type,
         num_workers=num_workers,
-        download_root=download_root,
+        **opt_kwargs,
     )
-    model.logger.setLevel(logging.WARNING)
+    if hasattr(model, "logger"):
+        model.logger.setLevel(logging.WARNING)
```

Note I also suggested a check for logger, on my installation `model` did not have the attribute `logger` defined.

pzelasko (changes requested):
I ran into a few issues running it, can you make the suggested changes that fix this?

entn-at:
Sorry for the delay, I've been quite busy. I'll pick this up shortly and address the requested changes.

desh2608:
@entn-at any updates on this?

PR #1019, opened 2023-04-06
Function to drop alignments from cut

desh2608:
This is just a helper function to drop alignments if they are not required. I have been using it in my multi-talker dataset class to reduce size of the returned batch. Might be useful until such time as we decide to store alignments in a separate archive, like features.

### DIFF
--- a/lhotse/cut/data.py
+++ b/lhotse/cut/data.py
@@ -382,6 +382,12 @@ class DataCut(Cut, metaclass=ABCMeta):
         """Return a copy of the current :class:`.DataCut`, detached from ``supervisions``."""
         return fastcopy(self, supervisions=[])
 
+    def drop_alignments(self) -> "DataCut":
+        """Return a copy of the current :class:`.DataCut`, detached from ``alignments``."""
+        return fastcopy(
+            self, supervisions=[fastcopy(s, alignment={}) for s in self.supervisions]
+        )
+
     def fill_supervision(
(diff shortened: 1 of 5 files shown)


pzelasko:
> Might be useful until such time as we decide to store alignments in a separate archive, like features.

This should be possible to do with Shar using sth like:

```python
# Writing
from lhotse.shar import JsonlShardWriter

cuts = CutSet.from_file(...)
with JsonlShardWriter("alis.jsonl.gz", shard_size=None) as w:
  for cut in cuts:
    ali = get_ali(cut)
    if ali is not None:
      w.write({"cut_id": cut.id, "word_alignment": [item.serialize() for item in ali["word"]]})
    else:
      w.write_placeholder(cut.id)

# Reading
cuts = CutSet.from_shar({"cuts": "path/to/cuts.jsonl.gz", "word_alignment": "alis.jsonl.gz"})
for cut in cuts:
  ali = cut.word_alignment
``` 

It's a bit raw, because currently we can't attach the alignment to the supervision with Shar, it's instead attached to the cut. Either:
- we'd support attaching shar fields to supervision
- we'd rewrite Lhotse a bit to support alignments being a property on cut rather than supervision (which would probably cause problems for users so naturally I don't like it; also not sure if it makes sense in the overall design)

pzelasko (approved):

pzelasko:
Anyway merging, this is useful :)

PR #1020, opened 2023-04-06
Add `padding_value` attribute to features

desh2608:
At the moment, when we are mixing features, we expect the user to specify a padding value. It would be better if the feature extractor itself came with a default value that should be used for padding.

### DIFF
--- a/docs/corpus.rst
+++ b/docs/corpus.rst
@@ -87,6 +87,12 @@ a CLI tool that create the manifests given a corpus directory.
     - :func:`lhotse.recipes.prepare_dihard3`
   * - DiPCo
     - :func:`lhotse.recipes.prepare_dipco`
+  * - Earnings'21
+    - :func:`lhotse.recipes.prepare_earnings21`
+  * - Earnings'22
+    - :func:`lhotse.recipes.prepare_earnings22`
+  * - The Edinburgh International Accents of English Corpus
+    - :func:`lhotse.recipes.prepare_edacc`
   * - English Broadcast News 1997
     - :func:`lhotse.recipes.prepare_broadcast_news`
   * - Fisher English Part 1, 2
--- a/lhotse/bin/modes/recipes/__init__.py
+++ b/lhotse/bin/modes/recipes/__init__.py
@@ -24,6 +24,7 @@ from .dihard3 import *
 from .dipco import *
 from .earnings21 import *
 from .earnings22 import *
+from .edacc import *
 from .eval2000 import *
 from .fisher_english import *
 from .fisher_spanish import *
--- a/lhotse/bin/modes/recipes/edacc.py
+++ b/lhotse/bin/modes/recipes/edacc.py
@@ -1,0 +1,24 @@
+from typing import Sequence
+
+import click
+
+from lhotse.bin.modes import download, prepare
+from lhotse.recipes.edacc import download_edacc, prepare_edacc
+from lhotse.utils import Pathlike
+
+__all__ = ["edacc"]
+
+
+@prepare.command(context_settings=dict(show_default=True))
+@click.argument("corpus_dir", type=click.Path(exists=True, dir_okay=True))
+@click.argument("output_dir", type=click.Path())
+def edacc(corpus_dir: Pathlike, output_dir: Pathlike):
+    """The Edinburgh International Accents of English Corpus (EDACC) data preparation."""
+    prepare_edacc(corpus_dir, output_dir=output_dir)
+
+
+@download.command(context_settings=dict(show_default=True))
+@click.argument("target_dir", type=click.Path())
+def edacc(target_dir: Pathlike):
+    """The Edinburgh International Accents of English Corpus (EDACC) download."""
+    download_edacc(target_dir)
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -1000,8 +1000,8 @@ class CutSet(Serializable, AlgorithmMixin):
                     f"CutSet has only {len(self)} items but last {last} required; not doing anything."
                 )
                 return self
-            cut_ids = list(self.ids)[-last:]
-            return CutSet.from_cuts(self[cid] for cid in cut_ids)
+            N = len(self)
+            return CutSet.from_cuts(islice(self, N - last, N))
 
         if supervision_ids is not None:
             # Remove cuts without supervisions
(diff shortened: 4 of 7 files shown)


pzelasko:
That's smart! 

As a side note I was also considering if we should support more nuanced padding strategies, e.g. with mean example/batch value (so that it doesn't influence things such as layer norm), but I'm not sure if there's much benefit to it.

lhotse/features/mixer.py
```diff
@@ -50,7 +50,11 @@ def __init__(
         self.num_channels = 1 if base_feats.ndim == 2 else base_feats.shape[-1]
         self.gains = []
         self.frame_shift = frame_shift
-        self.padding_value = padding_value
+        self.padding_value = (
+            feature_extractor.padding_value
+            if hasattr(feature_extractor, "padding_value")
```
pzelasko, line 55: I think since this behavior is undocumented in this PRs current shape, it may lead to some surprises / be hard to discover for implementers of other feature extractors how to handle this. What if we implemented this property on the base feature extractor class and made it return `None` there, and check for None here instead? Then we can add a proper docstring in feature extractor / documentation.

lhotse/features/base.py
```diff
@@ -90,6 +93,10 @@ def frame_shift(self) -> Seconds:
     def feature_dim(self, sampling_rate: int) -> int:
         ...
 
+    @property
+    def padding_value(self) -> float:
```
csukuangfj, line 97: The type hint for the return value does not match the actual return value.

lhotse/features/kaldifeat.py
```diff
@@ -193,6 +193,10 @@ def __init__(self, config: Optional[KaldifeatFbankConfig] = None) -> None:
     def feature_dim(self, sampling_rate: int) -> int:
         return self.config.mel_opts.num_bins
 
+    @property
+    def padding_value(self) -> float:
+        return -1000.0 if self.config.use_log_fbank else EPSILON
```
csukuangfj, line 198: Should -1000 be replaced with 
https://github.com/lhotse-speech/lhotse/blob/14b51f6e08166725c0e6b2ac29a210f61af4e684/lhotse/utils.py#L46

lhotse/features/mixer.py
```diff
@@ -41,7 +41,8 @@ def __init__(
         :param frame_shift: Required to correctly compute offset and padding during the mix.
         :param padding_value: The value used to pad the shorter features during the mix.
             This value is adequate only for log space features. For non-log space features,
-            e.g. energies, use either 0 or a small positive value like 1e-5.
+            e.g. energies, use either 0 or a small positive value like 1e-5. This value will be
+            ignore if the ``feature_extractor`` has a ``padding_value`` attribute.
```
csukuangfj, line 45:
```suggestion
            ignored if the ``feature_extractor`` has a ``padding_value`` attribute.
```

csukuangfj: Should this user-provided argument have a higher priority?

lhotse/features/kaldifeat.py
desh2608: Yes, ideally it should, but I was basing this on the default value used in the `FeatureMixer`.

lhotse/features/mixer.py
desh2608: Perhaps the default value for this option can be set to None. If it is None, we use the value from the feature extractor, otherwise we use the user-provided value. But I'm not sure what should be done if they are both None. Raise a error? @pzelasko what do you think?

pzelasko: I like your solution. Raising an error in this case is acceptable if you can add the default padding value attributes to the remaining feature extractors.

desh2608:
I have propagated the padding value related changes to all the affected parts of the codebase. There are a few subtle things which are causing some of the test cases to fail. Basically, when we add a padding cut to a mixed cut such that the resulting duration increases, there is double addition of padding that happens. This is because the original cut is first extended to the new duration with a padding value, and then the feats of the mixed-in cut are added.

One way to avoid this is to pass an additional `is_padding` option to `add_to_mix()` to avoid this double addition.

desh2608:
> I have propagated the padding value related changes to all the affected parts of the codebase. There are a few subtle things which are causing some of the test cases to fail. Basically, when we add a padding cut to a mixed cut such that the resulting duration increases, there is double addition of padding that happens. This is because the original cut is first extended to the new duration with a padding value, and then the feats of the mixed-in cut are added.
> 
> One way to avoid this is to pass an additional `is_padding` option to `add_to_mix()` to avoid this double addition.

This would still cause inconsistencies though. Consider the following case:
```python
# c is a 10s MonoCut, for example
m1  = c.mix(c)
m2 = m1.mix(c)
m1_padded = m1.pad(duration=15)
m2_padded = m2.pad(duration=15)
```

If we keep adding tracks to the mixture, the feature value of the padding region will keep increasing, whereas ideally we only want padding to happen once.

desh2608:
This was not caught in the test case before because the default padding value used was -1000, and `np.log(EPSILON + np.exp(-1000) + np.exp(-1000) + ...)` is still almost equal to `LOG_EPSILON`.

desh2608:
> This was not caught in the test case before because the default padding value used was -1000, and `np.log(EPSILON + np.exp(-1000) + np.exp(-1000) + ...)` is still almost equal to `LOG_EPSILON`.

I suppose this means that the easy solution is to use a very large negative number (like -1000) instead of LOG_EPSILON (which is -23) for padding value of log scale features.

pzelasko:
Hmm, now I remember that’s why I hardcoded -1000 in the feature mixer in the first place. I am a bit afraid that changing -23 (which is log(energy_floor) for a default value of energy floor in our feature extractors) to -1000 could somehow negatively affect the models as a „too out of distribution” value, especially if somebody uses layer norm or instance MVN. I agree the most elegant solution would be to pad once and have the feature mixer resolve that in a „smart” way to avoid the addition of energy floors, but maybe this is not really an issue? WDYT?

lhotse/features/fbank.py
```diff
@@ -54,6 +54,10 @@ def _feature_fn(self, *args, **kwargs):
     def feature_dim(self, sampling_rate: int) -> int:
         return self.config.num_mel_bins
 
+    @property
+    def padding_value(self) -> float:
+        return -1000.0
```
csukuangfj, line 59: Is there a reason to choose -1000.0 instead of `LOG_EPSILON` or just `0`?

desh2608: Check the discussion in the main thread.

PR #1023, opened 2023-04-07
Fix bug in lazy CutSet subset with last

desh2608:
This was raised in https://github.com/lhotse-speech/lhotse/discussions/1018#discussioncomment-5552948.

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -996,7 +996,8 @@ class CutSet(Serializable, AlgorithmMixin):
                     f"CutSet has only {len(self)} items but last {last} required; not doing anything."
                 )
                 return self
-            return CutSet.from_cuts(islice(self, len(self) - last, len(self)))
+            N = len(self)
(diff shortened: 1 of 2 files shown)


lhotse/cut/set.py
```diff
@@ -996,8 +996,7 @@ def subset(
                     f"CutSet has only {len(self)} items but last {last} required; not doing anything."
                 )
                 return self
-            cut_ids = list(self.ids)[-last:]
-            return CutSet.from_cuts(self[cid] for cid in cut_ids)
+            return CutSet.from_cuts(islice(self, len(self) - last, len(self)))
```
pzelasko, line 999: Can you assign len(self) to a variable to avoid double iteration for lazy cut sets?

desh2608: Yeah, good point.

desh2608: Done.

pzelasko (approved):

pzelasko:
Thanks!

PR #1032, opened 2023-04-18
API to enable/disable ffmpeg-torchaudio

desh2608:
Ref: https://github.com/lhotse-speech/lhotse/pull/1025#issuecomment-1511711351


lhotse/audio.py
```diff
@@ -116,6 +117,30 @@ def set_audio_duration_mismatch_tolerance(delta: Seconds) -> None:
...
+
+    Example::
+
+        >>> import lhotse
+        >>> lhotse.set_ffmpeg_torchaudio_info_enabled(False)  # don't use torchaudio
+
+    :param enabled: Whether to use torchaudio to compute audio file duration.
+    """
+    global FFMPEG_TORCHAUDIO_INFO_ENABLED
+    logging.info(
```
pzelasko, line 135: I would only emit the log if the new value is different from the old one

pzelasko (approved):

pzelasko:
Thanks

PR #1035, opened 2023-04-20
Add `transforms` attribute for MixedCut

desh2608:
Currently, for `MixedCut`, any transforms are always lazily applied on the underlying tracks. This is fine for transforms such as speed perturbation, but may not be ideal for other kinds of transforms. Consider the following examples:

1. We have a `MixedCut` consisting of 3 tracks: 2 `MonoCut`s and a `PaddingCut`, i.e., it represents 2 utterances from a speaker with some silence in between. Suppose we want to convolve it with an RIR. In the current method, the tracks will be individually convolved, and the silence portion will not contain any residue from the previous `MonoCut`. Instead, we should first mix the 3 tracks and then apply the RIR on the resulting audio.

2. We have a `MixedCut` comprising 2 `MonoCut`s with some overlap, and we want to perform loudness normalization. Currently, the tracks will be normalized individually and then mixed, and the resulting mixture may not have the target loudness.

To solve these problems, we introduce a `transforms` attribute for `MixedCut`. It is analogous to the `transforms` attribute in `Recording`, and contains the transforms that should be lazily applied at the time of audio loading, but after the underlying tracks have been mixed.

### DIFF
--- a/lhotse/cut/mixed.py
+++ b/lhotse/cut/mixed.py
@@ -722,11 +722,16 @@ class MixedCut(Cut):
             ],
         )
 
-    def normalize_loudness(self, target: float, affix_id: bool = False) -> "DataCut":
+    def normalize_loudness(
+        self, target: float, mix_first: bool = True, affix_id: bool = False
+    ) -> "DataCut":
         """
         Return a new ``MixedCut`` that will lazily apply loudness normalization.
 
         :param target: The target loudness in dBFS.
+        :param mix_first: If true, we will mix the underlying cuts before applying
+            loudness normalization. If false, we cannot guarantee that the resulting
+            cut will have the target loudness.
         :param affix_id: When true, we will modify the ``DataCut.id`` field
             by affixing it with "_ln{target}".
         :return: a modified copy of the current ``DataCut``.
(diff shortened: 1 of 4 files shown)


lhotse/cut/mixed.py
```diff
@@ -760,23 +821,64 @@ def reverb_rir(
             self.tracks
         ), "Invalid number of channels in `rir_channels`, must be either 1 or equal to the number of tracks."
 
+        # There are 2 ways to apply RIRs:
+        # 1. Mix the tracks first, then apply RIRs. This is same as applying the same RIR
+        #    to all tracks. It does not make sense if all tracks belong to different speakers,
+        #    but it is useful for cases when we have a mixture of MonoCut and PaddingCut,
+        #    and we want to apply the same RIR to all of them.
+        # 2. Apply RIRs to each track separately. This is useful when we want to simulate
```
pzelasko, line 829: Your description here got me thinking -- does it really matter in which order we apply RIR and downmix? Casting aside the numerical differences, RIR reverb is a convolution, which is linear, so the sum of reverbed signals should be equal to the reverbed sum of signals, shouldn't it? Am I missing something here?

pzelasko: Ohh I missed the effect on track edges -- the signals are possibly not the same length and have different offsets, and before the PR we only pad them after and before the convolution. You mentioned this in the PR description.

pzelasko:
Cool contribution! One question: what happens if we mix another track after the reverb/loudnorm is applied? It looks to me like the behaviour wouldn't be correct (it would reverb/norm everything including the track mixed after).

desh2608:
> Cool contribution! One question: what happens if we mix another track after the reverb/loudnorm is applied? It looks to me like the behaviour wouldn't be correct (it would reverb/norm everything including the track mixed after).

Yeah, I think perhaps the right approach is to not allow mixing in new tracks once any of the `mix_first` transforms have been applied.

lhotse/cut/mixed.py
desh2608: Yeah I was mainly concerned about edge effects in the case of reverb.

pzelasko:
> Yeah, I think perhaps the right approach is to not allow mixing in new tracks once any of the `mix_first` transforms have been applied.

Would it work if we added a MixedCut as a MixTrack when it has a transform defined on it? I think that technically mixing code only calls load_audio so it might work as expected..

desh2608:
Actually yeah we do allow MixedCut to be present as a MixTrack. In fact, the meeting simulation workflow explicitly uses this flexibility.

So you're right, if a new track is mixed into a MixedCut containing a transform, we can just create a nested MixedCut instead of adding to the tracks.

pzelasko (approved):
LGTM

Issue #1036, opened 2023-04-21
Unused sample rate in ReverbWithImpulseResponse


csukuangfj:
https://github.com/lhotse-speech/lhotse/blob/0f812851aefb1dc0560e76641f86bbcfcd96a47c/lhotse/augmentation/rir.py#L64

The passed sample rate is never used.

Also, the implementation does not check the sampling rate of the `rir` recording
https://github.com/lhotse-speech/lhotse/blob/0f812851aefb1dc0560e76641f86bbcfcd96a47c/lhotse/augmentation/rir.py#L28

It uses the input recording's sampling rate to generate an RIR
https://github.com/lhotse-speech/lhotse/blob/0f812851aefb1dc0560e76641f86bbcfcd96a47c/lhotse/audio.py#L818-L822

---

Shall we check that when `rir` is a recording its sampling rate should be equal to that of the input recording?

PR #1037, opened 2023-04-21
Ensure RIR has same sampling rate as audio

desh2608:
Fix for https://github.com/lhotse-speech/lhotse/issues/1036.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -795,6 +795,10 @@ class Recording:
         :param source_rng_seed: The seed to be used for the source position.
         :return: the perturbed ``Recording``.
         """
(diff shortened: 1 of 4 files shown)


desh2608:
Actually I think it may be better to check the sampling rate at the time of calling `reverb_rir()`, since at the time of `load_audio()` it may have gotten modified by a resampling operation. @pzelasko does this sound correct?

pzelasko:
Yes I think you're right. If the sampling rates are matching at the time of calling reverb_rir, it will be safe to apply it.

pzelasko (approved):

pzelasko:
Thanks!

PR #1043, opened 2023-04-26
Integrate torchaudio's 2.0 ffmpeg backend for audio loading + add some optimizations

pzelasko:
At least for in-memory data with file-like objects, I found ffmpeg FLAC decoder to be twice as fast as that of libsoundfile.


pzelasko:
Turns out when we enable ffmpeg backend it seems to slightly modify the audio when saving to WAVE (by up to 2/int16max), so I modified some tests to work around it and filed an issue in torchaudio https://github.com/pytorch/audio/issues/3281

PR #1048, opened 2023-04-28
AMI beamformed mic option

desh2608:
AMI comes with official beamformed recordings, so we also add this option.

Usage: `lhotse prepare ami --mic mdm8-bf /path/to/amicorpus data/`


pzelasko (approved):
Cool, I didn’t know that. I’m curious; are there any results for this setup vs others?

desh2608:
I think these were created using BeamformIt (most likely using the [Kaldi AMI recipe](https://github.com/kaldi-asr/kaldi/blob/19185083f4ce3f74d7b2fc7494b8ea530feeab01/egs/ami/s5b/run.sh#L55)), so the same results from that recipe should hold. On the dev set, SDM gets 33.3% and MDM gets 31.8% with the latest chain TDNN model.

In general, users can simply download the MDM recordings and run beamforming, but I thought it would save storage and effort to provide an option for already beamformed mics.

PR #1051, opened 2023-05-03
Add options to prepare data according to CHiME-7

desh2608:
Small changes in CHiME-6 and DiPCo recipes to prepare them according to the regulations of the CHiME-7 challenge.


pzelasko (approved):
Looks good! Out of curiosity, what was the reason to change the splits in CHIME7?

desh2608:
> Looks good! Out of curiosity, what was the reason to change the splits in CHIME7?

Pre-trained SSL models are allowed in CHiME-7, so we moved 2 sessions from train to eval. This should allow a better evaluation of systems.

PR #1054, opened 2023-05-04
Support audio duration mismatch tolerance in `MixedCut.load_audio()`

pzelasko:


lhotse/workflows/meeting_simulation/speaker_independent.py
```diff
@@ -218,3 +172,55 @@ def _simulate_worker(utterances: CutSet) -> MixedCut:
...
+        loc + npr.exponential(scale=scale, size=len(utterances[i]))
+        for i in range(len(utterances))
+    ]
+
+    # Create the meeting.
+    mixture = _create_mixture(utterances, silence_durations)
+    return mixture
+
+
+def _create_mixture(
```
desh2608, line 204: Is there a reason why this function is moved outside the class? The algorithm is specific to this particular simulator.

pzelasko: Makes sense, I changed this back and used the same "trick" as in the simulator.

desh2608: Thanks. Could you do the same for `_simulate_worker` (i.e. make it a function of the specific class)? The logic for those is also tied to that algorithm.

pzelasko: That would complicate the workarounds for parallelization; anyway each simulation method has a separate file, so I think it should be OK and actually consistent with other functions in lhotse that are being parallelized this way, e.g. `_X_single()` functions at the end of `lhotse/cut/set.py`.

desh2608: Okay, LGTM.

desh2608 (approved):

PR #1058, opened 2023-05-16
Enhancements and bug fixes for AMI and ICSI

desh2608:
* Add a `merge_consecutive` option for AMI. This is to prevent over-segmentation which results in too many short segments like "Yeah", "Okay", etc.
* Return word alignments in AMI supervisions (these are provided in office AMI annotations).
* Fix some bugs in ICSI recipe (e.g. `fix_manifests()` should be done before saving to file).
* Return word alignments in ICSI supervisions (provided in annotations).
* Add option for saving ICSI sph files to wav, so that they are faster to process later.


pzelasko (approved):
LGTM

Issue #1034, opened 2023-04-19
trim_to_supervisions


bloodraven66:
In trim_to_supervisions (https://lhotse.readthedocs.io/en/latest/_modules/lhotse/cut/base.html#Cut.trim_to_supervisions),
`ignore_channel=True`  should be ` keep_all_channels=False` in the assert message?

desh2608:
I think it should be `keep_all_channels=True`.

PR #1059, opened 2023-05-16
Small changes in some cut methods

desh2608:
* When trimming to supervision groups, give sensible ids instead of completely randomized.
* Progress bar for exporting to Shar.
* Minor fix in error message.
* Fix for https://github.com/lhotse-speech/lhotse/issues/1034.

### DIFF
--- a/test/cut/test_cut_mixing.py
+++ b/test/cut/test_cut_mixing.py
@@ -324,6 +324,12 @@ def test_mix_cut_snr(libri_cut):
     assert E(feats) > E(feats_snr)
 
 
+def test_mix_cut_with_other_raises_error(libri_cut):
+    libri_cut = libri_cut.drop_features()
+    with pytest.raises(ValueError):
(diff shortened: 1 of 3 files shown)


test/cut/test_cut_mixing.py
```diff
@@ -324,6 +324,12 @@ def test_mix_cut_snr(libri_cut):
     assert E(feats) > E(feats_snr)
 
 
+def test_mix_cut_with_other_raises_error(libri_cut):
+    libri_cut = libri_cut.drop_features()
+    with pytest.raises(ValueError):
+        _ = libri_cut.mix(libri_cut.recording)
```
pzelasko, line 330: why is this test supposed to fail? because of `drop features` or because the argument to `mix` is of type `Recording`?

pzelasko (approved):

pzelasko:
Thanks!

PR #1060, opened 2023-05-16
Fix `resumable_download` for fully downloaded files

flyingleafe:
When `resumable_download` is called and the file is fully downloaded, the requested range starts after the file ends, which yields HTTP error 416. It seems logical that in this case the exception should not be raised, but rather Lhotse should report that the file is already downloaded.

### DIFF
--- a/lhotse/utils.py
+++ b/lhotse/utils.py
@@ -468,16 +468,17 @@ def resumable_download(
 
     # Open the file for writing in binary mode and seek to the end
     with open(filename, "ab") as f:
-        f.seek(file_size)
 
-        try:
+        def _download(rq, size):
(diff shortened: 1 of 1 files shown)


flyingleafe:
@rouseabout Fair enough. If the server does not support resumable downloads though, we should not simply fail but rather try to download from scratch with a request without range headers then. Updated PR accordingly.

pzelasko (dismissed):
LGTM, thanks!

pzelasko:
Please just fix the CI checks

flyingleafe:
@pzelasko only black was failing, should be ok now.

pzelasko (approved):

PR #1067, opened 2023-05-22
Use torchaudio to read sph files first

desh2608:
New torchaudio with ffmpeg backend can read SPH files. It doesn't seem to be any faster than sph2pipe, but at least it does not require installing new tools.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1767,7 +1767,6 @@ def info(
     force_opus_sampling_rate: Optional[int] = None,
     force_read_audio: bool = False,
 ) -> LibsndfileCompatibleAudioInfo:
-
     if force_read_audio:
         # This is a reliable fallback for situations when the user knows that audio files do not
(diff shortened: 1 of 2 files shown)


pzelasko:
Did you check that it works with `shorten` encoded SPH? Some older LDC distributions have SPH that couldn't be opened with anything other than sph2pipe or shorten.

desh2608:
> Did you check that it works with `shorten` encoded SPH? Some older LDC distributions have SPH that couldn't be opened with anything other than sph2pipe or shorten.

Is it in any of the test cases? I only tried it on ICSI dataset where it works fine, but I'm not sure how those are encoded. I can add a test cases for shorten.

pzelasko:
No, I don't think it's in the tests. You'd have to try something old like maybe Callhome, not sure but maybe also SWBD. I don't remember which corpora this happened with unfortunately.

Issue #1064, opened 2023-05-22
Supervision and feature frame size mismatch


rouseabout:
For some specific entries in my dataset, I observse a mismatch of between supervision duration and actual audio samples passed to the feature extractor. The mismatch is very small, equating to one or two samples, but is enough to cause an additional feature frame to be allocated. During training, icefall/utils.py fails with an assertion because the audio in frame size (computed from audio samples) does not match the supervision frame size (computed from supervision duration).

```
def filter_uneven_sized_batch(batch: dict, allowed_max_frames: int):
[...]
	assert T == supervisions["num_frames"].max(), (T, supervisions["num_frames"].max())
```

This mismatch can occur in two places. I am curious what the best solution is.

**Mismatch 1: CutSet.trim_to_supervisions() with supervision durations that do not compute to a round number of samples**

My dataset has some supervision durations (in seconds) that do not compute to a round number of samples at the original recording sample rate. When trim_to_supervisions() is called, add_durations() is used to recalculate the cut duration. The new duration is calculated by converting the supervision duration(s) from seconds to samples (with ROUND_HALF_UP), adding then up, and then converting back to seconds.

Worked example using atcosim_zf1_02_120_000000_000107:

- 	supervision duration: 3.7249687 seconds (119198.9984 samples at 32000 kHz)

After trim_to_supervisions() 

- 	cut duration: 3.72496875 seconds (119199 samples at 32000 kHz)

If we resample to 16000 and calculate frame size for 10ms window:

- 	supervision frames: 372 frames  (`compute_num_frames(3.7249687, 0.01 ,16000)`)
- 	feature frames:  373 frames (`compute_num_frames(3.72496875, 0.01 ,16000)`)

Solutions?:

- 	During the prepare phase of the recipe, fix problematic segment duration values so they calculate to a round number of samples.

**Mismatch 2: MonoCut.load_audio() optimization**

commit e4c5cf97f introduces the optimization:

```
                # set duration=None when cut covers the whole recording
                # to let the audio loading code "know" to read everything
                duration=self.duration
                if not math.isclose(
                    self.duration, self.recording.duration, abs_tol=1e-3
                )
                else None,
```

Even with 10^-3 tolerance, the number of samples returned by load_audio() is sometimes sufficiently different to cause an extra feature frame to be allocated.

Worked example using “atcosim_sm4_05_025_000000_000553”:

- 	recording duration = 5.53496875 seconds (177119 samples at 32000 Hz)
[...]

PR #1068, opened 2023-05-23
cut/mono: only optimize load_audio() when returned number of feature frames is identical

rouseabout:
Candidate fix for second problem reported in #1064

This adds an optional _frame_shift_ parameter to cut/mono:load_audio(). We compute the number of frames for the cut.duration and cut.recording.duration and only apply the optimization if the resulting number of feature frames are identical.

Idea(?): The duration math.isclose() test could be removed altogether.


pzelasko:
Thank you for the contribution. I think I have a slightly different idea how to fix this issue, I'll try to push it sometime soon.

pzelasko:
Please see #1071

pzelasko:
Closing as #1071 was preferred

PR #1071, opened 2023-05-24
Fix for audio loading optimization to return the expected number of samples

pzelasko:
@rouseabout please see this alternative to your PR https://github.com/lhotse-speech/lhotse/pull/1068, and let me know if it solved your issue.


rouseabout:
Can confirm, this solves my issue.

When applied issue https://github.com/lhotse-speech/lhotse/issues/1064  can be marked as closed. Thanks.

PR #1073, opened 2023-05-25
Fix bugs in `MixedCut` logic

flyingleafe:
I attempted to simulate some conversations, then cut them into windows and save the resulting cuts on the disk. I have encountered two bugs:

- `IndexError` in `MixedCut.truncate`: happened because the cut window happened to be in the region between `MixTrack`s, so no intersecting tracks were found and such case was not handled;
- `AssertionError` in `MixedCut.load_audio`: the resulting audio length were mismatched, because the offset of the first track in the cut was not accounted for during mixing, resulting in incorrect mix length.

This PR fixes those:
- returning `PaddingCut` in the first case;
- accounting for the base audio offset in the mixer, via providing an extra optional constructor parameter.

### DIFF
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1301,6 +1301,7 @@ class AudioMixer:
         base_audio: np.ndarray,
         sampling_rate: int,
         reference_energy: Optional[float] = None,
+        base_offset: Seconds = 0.0,
     ):
         """
         AudioMixer's constructor.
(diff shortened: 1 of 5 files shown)


flyingleafe:
@desh2608 yes, added.

pzelasko (approved):
Thanks, LGTM!

PR #1077, opened 2023-05-28
Expose keep_excessive_supervision parameter

nshmyrev:
Currently when supervision is slightly longer than recording user gets very confusing message about multiple channels:

> Trimmed cut has supervisions with different channels. Either set `keep_all_channels=True` to keep original channels or `keep_overlapping=False` to retain only 1 supervision per trimmed cut.

This is because keep_overlapping is False and thus we filter out those bad supervisions. And keep_overalapping actually has nothing about the problem of wrong supervision times. One better expose keep_excessive_supervision parameter to method and provide more straightforward assert message.

### DIFF
--- a/lhotse/cut/base.py
+++ b/lhotse/cut/base.py
@@ -393,6 +393,7 @@ class Cut:
     def trim_to_supervisions(
         self,
         keep_overlapping: bool = True,
+        keep_excessive_supervisions: bool = True,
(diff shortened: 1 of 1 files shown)


desh2608:
I think it may be better to first use fix_manifests() from lhotse/qa.py, instead of this. Then the longer supervisions will be trimmed to the recording length.

PR #1083, opened 2023-06-01
Set more-default audio backend

MicKot:
Allows you to set which backend is the first to read audio.

For some reason in my case AudioreadBackend is even 2x as fast as whatever reads my files by default.


MicKot:
@pzelasko Check this out please

MicKot:
But it returns non-default backends only if you yourself change them at which point they may not be default to `lhotse` but default to you/whatever you are runnig atm. I would argue that this method is a proper one:  you set your prefered backend while keeping other options in place (in case wild audio appears), what I think was the idea with `CompositeAudioBackend` in the first place. And it doesn't require changing code in other places.

pzelasko:
I agree with your approach in spirit, I just don't want to confuse the users by modifying "default" backend which should always return the same thing. It's OK to change the code a bit more if that makes it more transparent.

pzelasko:
This PR is now superseded by #1144 which adds the ability to change the audio backend. Thanks for your contribution!

PR #1085, opened 2023-06-01
`CutSet.from_files` constructor for random order multi-file cutsets

pzelasko:
An example of typical usage is:

```python
# We start with a big cut set that refers to all of the data.
big_cutset = CutSet.from_file("100k_hour_cuts.jsonl.gz")

# It's being split into smaller parts below.
big_cutset.split_lazy("shards", chunk_size=1000)

# Now we can iterate over the shards, randomizing their order on each iteration.
# Randomization is enabled by default.
sharded_cutset = CutSet.from_files(Path("shards").glob("*.jsonl.gz"))

# Order of shards will be randomized
for cut in sharded_cutset:
    pass

# This time the order will be different
for cut in sharded_cutset:
   pass

# We can randomize the cuts across the shards as well in a streaming fashion
# (note: this happens automatically in samplers when shuffle=True)
cuts = sharded_cutset.shuffle(buffer_size=50000)

# Works as usual with samplers
sampler = DynamicBucketingSampler(sharded_cutset, ...)
```

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -271,6 +271,34 @@ class CutSet(Serializable, AlgorithmMixin):
         )
 
     @staticmethod
+    def from_files(
+        paths: List[Pathlike], shuffle_iters: bool = True, seed: int = 0
+    ) -> "CutSet":
+        """
+        Constructor that creates a single CutSet out of many manifest files.
+        We will iterate sequentially over each of the files, and by default we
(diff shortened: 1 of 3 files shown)


pzelasko:
I added one more improvement: by default we'll use Python RNG here so it's automatically different on each script instantiation. The user can specify the seed if they want reproducibility.

PR #1087, opened 2023-06-08
Loudness fix

desh2608:
* Minor fix for the meter in loudness computation (for short segments)
* Minor fix for supervision channel in mono cut created from mixed cut


pzelasko (approved):
Thanks!

PR #1089, opened 2023-06-12
[Recipe] VoxPopuli

desh2608:


pzelasko:
IMO LGTM after fixing and if somebody wants to add s2s mode later, they can add it.

desh2608:
> IMO LGTM after fixing and if somebody wants to add s2s mode later, they can add it.

Yeah, I am already using the ASR prep part for some of the projects at JSALT. I like our method more than the official data preparation since it does not duplicate all the segments. I will fix the extract and then maybe we can merge.

pzelasko (approved):

PR #1094, opened 2023-06-19
Adding log spectrogram

Tomiinek:
This PR adds `kaldi-log-spectrogram` which outputs log features (similarly to torchaudio-based `spectrogram`) and is based on Lhotse's `Wav2FFT` so it supports streaming.


pzelasko (approved):
LGTM

PR #1095, opened 2023-06-20
CutSet.mix: add random_mix_offset option

rouseabout:
Patch attempts to address this situation: given a recording of continuous background noise spanning an hour (e.g. https://doi.org/10.5281/zenodo.2660112), and a training cutset of short clean utterances, create an augmented dataset.

Run-time performance is about the same as this:

```
noise = noise.cut_into_windows(duration=20).to_eager()
cuts = cuts.mix(noise)
```

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -1818,6 +1818,7 @@ class CutSet(Serializable, AlgorithmMixin):
         preserve_id: Optional[str] = None,
         mix_prob: float = 1.0,
         seed: int = 42,
+        random_mix_offset: bool = False,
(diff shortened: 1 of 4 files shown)


pzelasko:
Thanks, looks good to me! Can you add a test case for this option?

rouseabout:
> Thanks, looks good to me! Can you add a test case for this option?

Shall do. Give me a ~week or so.

rouseabout:
Changes:

* added complementary random_mix_offset option to the CutMix transform
* test cases for everything

pzelasko (approved):
Thanks! LGTM

Issue #1096, opened 2023-06-23
Request for feature (?),  a way to sample from large jsonl manifests.


danpovey:
@pzelasko for some recent language modeling work I have been using a very simple type of data loader and I find it very refreshing to work with.  The data loader is completely stateless: it makes no attempt to sync any kind of state, but just draws completely randomly from the full data each time.   If someone resumes training, we just use the start-batch index as part of the random seed, so it won't repeat the already seen data; but it also doesn't make an attempt to see the exact same data that you would have seen if you had iterated from the start.  [There are ways you could fix this fairly easily, though.]

In my case, for language modeling the way it works (https://github.com/k2-fsa/icefall/pull/1141/files#diff-d641dd2a3ec42b15b0f36fd57011a184ad8119a1fa478d5e4f1485b47113ab73) is we write a kind of manifest containing the list of training files and their length in bytes; and all the data loaders read this; they use a nn.random.multinomial to randomly select the training file with prob proportional to its length; then choose a random chunk within the file.  I decided that it was too much of a hassle to deal with epochs, so there is no concept of epoch.  In training mode, it just loops infinitely and the training loop decides when to exit, based on the total amount of data that has been seen.   We make the Eden scheduler depend on the number of tokens seen in this case, not on the epochs; this can be done from train.py without changing the optimizer code.

In eval mode, it just loops enough to see the data about once; but in reality you'd use a different data loader for validation data.

The way I propose to do this for not-compressed jsonl manifests, is as follows.

Firstly, the user would provide to this data loader either a list of jsonl files, or a list of (jsonl, scale) where `scale` is a real-valued factor that you can use to upweight or downweight the contribution of that jsonl file.  Each time we sample, we sample the jsonl file with probability (number_of_lines_in_that_jsonl_file * scale), and then randomly choose the line number within the jsonl file.

[...]

PR #1102, opened 2023-07-18
Infinite random-file random-line stateless sampler

pzelasko:
As proposed in #1096 

It's pretty much what was described, with at most minor deviations. @danpovey let me know what you think, you can find the usage in unit tests. I couldn't think of a name that would describe what this sampler does, so I initially called it `PoveySampler` -- we should continue to build on the legacy of the Povey feature window :)


lhotse/dataset/sampling/povey.py
```diff
@@ -0,0 +1,274 @@
...
+from lhotse.dataset import DynamicBucketingSampler, DynamicCutSampler
+from lhotse.dataset.sampling.base import SamplingDiagnostics
+from lhotse.lazy import Dillable
+from lhotse.serialization import decode_json_line
+from lhotse.utils import Pathlike
+
+PathlikeAndScale = Tuple[Pathlike, float]
+
+
+class PoveySampler(torch.utils.data.Sampler, Dillable):
```
csukuangfj, line 21: There is a "povey window". 
Now we also have a "povey sampler"  😀

danpovey: LOL, but let's call it StatelessSampler.  With the Povey Window, I think people didn't get that it was basically a joke (since no-one really cares about things like windows these days), and they perhaps thought that it was a kind of self-aggrandizement.

lhotse/dataset/sampling/povey.py
```diff
@@ -0,0 +1,274 @@
...
+
+        >>> sampler = PoveySampler(
+        ...     cuts_paths=["data/cuts_a.jsonl", "data/cuts_b.jsonl"],
+        ...     index_path="data/files.idx",
+        ...     max_duration=600.0,
+        ... )
+
+
+    .. note:: This sampler works only with uncompressed jsonl manifests, as it creates extra index files with line byte offsets to quickly find and sample JSON lines.
+     This means this sampler will not work with Webdataset and Lhotse Shar data format.
```
danpovey, line 53: Perhaps we should mention that if you restart training, you should set the random seed to some function of the step you restarted at, to ensure that you don't see the exact same data that you saw at the beginning of training (or at some previous restart point).

pzelasko: Good point, I'll also add some logging that prints out the random seeds and maybe other info at initialization so it'll be possible to keep track of these things between experiments.

pzelasko:
Check it now, I changed it to use OS TRNG for sampling the base seed (if available) so that it doesn't depend on Python's global seed at all.

pzelasko:
I'll merge it, let me know later if it works as you expected.

danpovey:
I'm not sure about the decisions about the TRNG stuff.  My plan was to have the base seed depend on the python RNG state.  The idea was that in the main thread, we set the RNG state to be a function of the start-iter if you are restarting, so if you are restarting training you get a different random seed.  OTOH it would be consistent between runs, so that if you get a crash it can be debugged.  Otherwise debugging crashes would be quite hard!

pzelasko:
It's going to print the seed it chose in the logs, and if you want to reproduce the run, you can pass the argument `base_seed=<value>` to the sampler (by default it's None which means choose it at random). The same argument can be used to pass another seed of your own choice.

danpovey:
Hm, OK.  When we use this I plan to pass in something like the start iteration then.
If the secrets module gives a different number per worker, it seems to me it would be quite painful to pass
in separate numbers per worker in order to reproduce a crash.

pzelasko:
You would be passing in only the base_seed, which is then modified in each node+worker combination similarly to the LM dataset code shared by you earlier. See the snippet linked here. 

https://github.com/lhotse-speech/lhotse/pull/1102/files#diff-4b4592f9922837830d7c5a7fa206aa9c9e8b9c000f9125dc6a651d8d6e1ed446R180

Just make sure to use the IterableDatasetWrapper as described here so that the sampler is being placed into the worker subprocess (so it can correctly read out the worker ID). 

https://github.com/lhotse-speech/lhotse/pull/1102/files#diff-4b4592f9922837830d7c5a7fa206aa9c9e8b9c000f9125dc6a651d8d6e1ed446R30

danpovey:
OK.  I think it's good if we encourage people to use it with IterableDatasetWrapper and also the base_seed (set as a function of at least the start-iter for restarts), because if they don't the use base_seed, in general they'll end up with different base_seed values in different workers, and reproducing failures would be hard.  Debugging isn't something you normally plan in advance to have to do.

pzelasko:
Alright, on another thought I'll simplify this code by removing TRNG and requiring `base_seed` to be passed in by the user.

PR #1103, opened 2023-07-18
text normalization for aishell4

JinZr:


lhotse/recipes/aishell4.py
```diff
@@ -134,7 +165,7 @@ def prepare_aishell4(
                             channel=recording.channel_ids,
                             language="Chinese",
                             speaker=spk_id,
-                            text=text.strip(),
+                            text=text_normalize(text.strip()),
```
desh2608, line 168: Could you add this as an optional parameter, e.g. `--normalize-text`?

pzelasko (approved):

PR #1106, opened 2023-07-20
PR for the KeSpeech recipe

JinZr:


pzelasko (commented):
Thanks, looks good to me! Would it be possible to add a download function (since it's openly available)?

Also, please update the table in `docs/corpus.rst` with this corpus.

JinZr:
Noted!
I'll look into the document part, whereas the download function might not
be able to be implemented as the link they provided does not support direct
downloading so far (check the url in the READMD.md here:
https://github.com/KeSpeech/KeSpeech).

pzelasko (approved):

PR #1117, opened 2023-08-07
Fixing shar reader assert

Tomiinek:
This PR fixes shar reader's assert that fails unexpectedly for cut IDs with slashes. For example, it fails if the cut ID is `small/5118/told_in_a_french_garden_1010_librivox_64kb_mp3/toldinafrenchgarden_05_aldrich_64kb` and the recording path is `small/5118/told_in_a_french_garden_1010_librivox_64kb_mp3/toldinafrenchgarden_05_aldrich_64kb.flac` (a realistic cut ID produced by this recipe https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/librilight.py)


pzelasko (approved):
Good catch!

PR #1127, opened 2023-08-23
Minor bug fix for eval2000 recipe

JinZr:


desh2608:
It may be useful to provide a `transcript_dir` option to the function, which defaults to the original path. Users can provide a different path based on how their corpus is set up.

lhotse/bin/modes/recipes/eval2000.py
```diff
@@ -6,16 +6,28 @@
 from lhotse.recipes import prepare_eval2000
 from lhotse.utils import Pathlike
 
+EVAL2000_TRANSCRIPT_DIR = "LDC2002T43"
+
 
 @prepare.command(context_settings=dict(show_default=True))
 @click.argument("corpus-dir", type=click.Path(exists=True, file_okay=False))
 @click.argument("output-dir", type=click.Path())
+@click.argument(
```
desh2608, line 15: Have you tested this with and without providing `transcript-dir`? It seems you are expecting this to be a positional argument which may not be what we want.

lhotse/recipes/eval2000.py
```diff
@@ -29,6 +29,7 @@
 def prepare_eval2000(
     corpus_dir: Pathlike,
     output_dir: Pathlike,
+    transcript_path: Pathlike,
```
desh2608, line 32:
```suggestion
    transcript_path: Optional[Pathlike] = None,
```

lhotse/recipes/eval2000.py
```diff
@@ -48,7 +49,7 @@ def prepare_eval2000(
     assert (
         audio_partition_dir_path.is_dir()
     ), f"No such directory:{audio_partition_dir_path}"
-    transcript_dir_path = corpus_dir / EVAL2000_TRANSCRIPT_DIR / "reference" / "english"
+    transcript_dir_path = transcript_path
```
desh2608, line 52:
```suggestion
    default_transcript_path = corpus_dir / EVAL2000_TRANSCRIPT_DIR / "reference" / "english"
    transcript_dir_path = default_transcript_path if transcript_path is None else transcript_path
```

lhotse/bin/modes/recipes/eval2000.py
JinZr: You are right, I'll look into it.

JinZr:
Thanks Desh, the latest commit should fix this issue.

lhotse/bin/modes/recipes/eval2000.py
```diff
@@ -10,12 +10,23 @@
...
+    "transcript-dir",
+    type=click.Path(exists=True, file_okay=False),
+    default=None,
+    required=False,
+)
```
desh2608, line 18: Should this be an `option` instead of an `argument`, since it is optional?

JinZr: The latest commit should fix this issue. 

Thanks!

pzelasko (approved):

pzelasko:
Thanks!

PR #1128, opened 2023-08-23
Add `fix_manifests` in all recipes

desh2608:
In my experience, most issues in cut manipulations arise because of some issues in the source supervisions (e.g., some supervisions exceeding recording duration). This can be fixed easily using the `fix_manifests` method.


pzelasko (approved):
Thanks for going through all of the recipes!

PR #1129, opened 2023-08-23
making the kaldi import more robust

vesis84:
get_duration():
- recover if audio file cannot be loaded for get_duration(), drop such recordings...
- use chunksize for ProcessPoolExecutor::map (avoid hanging of ProcessPoolExecutor for large RecordingSets)

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -1,3 +1,4 @@
+import logging
 import math
 import warnings
 from collections import defaultdict
@@ -42,12 +43,12 @@ def floor_duration_to_milliseconds(
 
 def get_duration(
     path: Pathlike,
-) -> float:
+) -> Optional[float]:
     """
     Read a audio file, it supports pipeline style wave path and real waveform.
 
     :param path: Path to an audio file or a Kaldi-style pipe.
-    :return: float duration of the recording, in seconds.
+    :return: float duration of the recording, in seconds or `None` in case of read error.
     """
     path = str(path)
     if path.strip().endswith("|"):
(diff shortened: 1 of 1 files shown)


vesis84:
Hi Piotr, what would you think about this change ?
Let's talk...

(I found this issue while importing per-utterance flac files for Chime challenge)
Cheers,
Karel

lhotse/kaldi.py
```diff
@@ -122,10 +133,30 @@ def fix_id(t: str) -> str:
...
+        with ProcessPoolExecutor(max_workers=num_jobs) as ex:
+            dur_vals = list(
+                ex.map(get_duration, recordings.values(), chunksize=chunksize)
+            )
+
         durations = dict(zip(recordings.keys(), dur_vals))
 
+    # remove recordings with 'None' duration (i.e. there was a read error)
+    for recording_id, duration in durations.items():
+        if durations == None:
```
pzelasko, line 151:
```suggestion
        if duration is None:
```

pzelasko (commented):
Looks good to me, I think I found one bug though (see the other comment) -- can you test it?

lhotse/kaldi.py
```diff
@@ -122,10 +133,30 @@ def fix_id(t: str) -> str:
...
+
         durations = dict(zip(recordings.keys(), dur_vals))
 
+    # remove recordings with 'None' duration (i.e. there was a read error)
+    for recording_id, dur_value in durations.items():
+        if dur_value == None:
+            logging.warning(
+                f"[{recording_id}] Could not get duration. "
+                f"Failed to read audio from `{recordings[recording_id]}`. "
+                f"Dropping the recording from manifest."
```
csukuangfj, line 155:
```suggestion
                "Dropping the recording from manifest."
```

lhotse/kaldi.py
```diff
@@ -122,10 +133,30 @@ def fix_id(t: str) -> str:
...
+        with ProcessPoolExecutor(max_workers=num_jobs) as ex:
+            dur_vals = list(
+                ex.map(get_duration, recordings.values(), chunksize=chunksize)
+            )
+
         durations = dict(zip(recordings.keys(), dur_vals))
 
+    # remove recordings with 'None' duration (i.e. there was a read error)
+    for recording_id, dur_value in durations.items():
+        if dur_value == None:
```
csukuangfj, line 151: To fix the style issue, we can use
```
        if dur_value is None:
```
to replace
```
        if dur_value == None:
```

vesis84:
Ok, both suggested changes are done. I also added a new sanity check...
Cheers,
K.

PR #1131, opened 2023-08-25
optimize save_audios()

vesis84:
Hi Piotr and Fangjun, 
i have also some improvements for the `CutSet::save_audios()` function.
This is mainly to improve the cache hit ratio for long audio files with
many utterances in each audio file.

I have also a calling script for that, would you be interested in having that too ?

Cheres,
Karel

- added sorted method : CutSet::sort_by_recording_id()
- allow to disable shuffling of CutSet inside `CutSet::save_audios()`
   - both changes improve cache hit ratio
- `CutSet::save_audios()` : show in log if caching was active
- caching.py : replace Union[] by Optional[]

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -2368,6 +2368,8 @@ class CutSet(Serializable, AlgorithmMixin):
             https://lhotse.readthedocs.io/en/latest/parallelism.html
         :param progress_bar: Should a progress bar be displayed (automatically turned off
             for parallel computation).
+        :param shuffle_on_split: Shuffle the ``CutSet`` before splitting it for the parallel workers.
+            It is active only when `num_jobs > 1`. The default is True.
         :return: Returns a new ``CutSet``.
         """
         from cytoolz import identity
@@ -2386,11 +2388,6 @@ class CutSet(Serializable, AlgorithmMixin):
                 "we will ignore the executor and use non-parallel execution."
             )
             executor = None
-
-        logging.info(
-            f"Check the AudioCache state: "
-            f"is_caching_enabled = {is_caching_enabled()}"
-        )
 
         def file_storage_path(cut: Cut, storage_path: Pathlike) -> Path:
             # Introduce a sub-directory that starts with the first 3 characters of the cut's ID.
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -34,6 +34,7 @@ from typing_extensions import Literal
 
 from lhotse.audio import RecordingSet, null_result_on_audio_loading_error
 from lhotse.augmentation import AugmentFn
+from lhotse.caching import is_caching_enabled
 from lhotse.cut.base import Cut
 from lhotse.cut.data import DataCut
 from lhotse.cut.mixed import MixedCut, MixTrack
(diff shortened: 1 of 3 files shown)


vesis84:
Hi @pzelasko @csukuangfj ,  did you have a chance to "peek" into this PR ?
Thank you, Karel

csukuangfj:
It looks good to me

👍 💯

lhotse/cut/set.py
```diff
@@ -2374,6 +2387,11 @@ def save_audios(
             )
             executor = None
 
+        logging.info(
```
pzelasko, line 2390: We generally don't emit logs in lhotse methods (for better or worse), can we remove it here?

lhotse/cut/set.py
```diff
@@ -1413,6 +1414,17 @@ def combine_same_recording_channels(self) -> "CutSet":
         groups = groupby(lambda cut: (cut.recording.id, cut.start, cut.end), self)
         return CutSet.from_cuts(MultiCut.from_mono(*cuts) for cuts in groups.values())
 
+    def sort_by_recording_id(self, ascending: bool = True) -> "CutSet":
```
pzelasko, line 1417: Could you add a unit test to cover this method?

lhotse/cut/set.py
```diff
@@ -2326,6 +2338,7 @@ def save_audios(
         executor: Optional[Executor] = None,
         augment_fn: Optional[AugmentFn] = None,
         progress_bar: bool = True,
+        shuffle_on_split: bool = True,
```
pzelasko, line 2341: This option is undocumented

pzelasko (commented):
Thanks Karel! Somehow this slipped through. I left a couple of comments.

lhotse/cut/set.py
vesis84: Ok, actually, by default the `logging.info(.)` level does not print anything. 
The first "active" verbosity level by default is : `logging.warn(.)`.

So, technically, it does not print anything, unless a debug mode is started by lhotse user...
Piotr, do you still want to remove it ?

csukuangfj: If lhotse is used in icefall and if icefall uses a logging level >= info, I think this message would still be printed out.

vesis84: Yes, that is also true.

lhotse/cut/set.py
vesis84: Related to that. 

I found a similar unit-test [test_cut_set_sort_by_duration()](https://github.com/lhotse-speech/lhotse/blob/master/test/cut/test_cut_set.py#L49).
Could you point me to the place, where this test is called ?

I did not find any code calling `test_cut_set_sort_by_duration()` in the `/test` 
folder and there are some arguments necessary to be filled...
Thx, Karel

vesis84:
The code is updated, but the unit test is not yet complete (it is missing the calling side).

vesis84:
Hi @pzelasko  @csukuangfj , 
all the remarks are integrated, and the unit-tests are passing.
It seems to be ready. 
Karel

pzelasko:
Cool, thanks!

pzelasko (approved):

PR #1142, opened 2023-09-13
Add VoxConverse recipe

flyingleafe:
From [here](https://github.com/joonson/voxconverse):

"VoxConverse is an audio-visual diarisation dataset consisting of multispeaker clips of human speech, extracted from YouTube videos.
Updates and additional information about the dataset can be found at our website (https://www.robots.ox.ac.uk/~vgg/data/voxconverse/index.html)."

Note: The default dev/test split is quite weird - namely, the test set is larger than the dev set... Hence, there is an option to use "dev" set as "train", and split "test" set in half into "dev" and "test", which can be disabled in the recipe.


flyingleafe:
@pzelasko ^ quite weird that a random test for a particular python version, totally irrelevant to the changes in the PR, fails

pzelasko:
> @pzelasko ^ quite weird that a random test for a particular python version, totally irrelevant to the changes in the PR, fails

Some tests depend on RNG and once in a while they get flaky. It's happening a bit too often though, so I'm trying to fix that in https://github.com/lhotse-speech/lhotse/pull/1143

lhotse/recipes/voxconverse.py
```diff
@@ -0,0 +1,149 @@
...
+        (corpus_dir / "annotations.zip").unlink()
+        completed_detector.touch()
+
+    print("Done")
+
+
+def prepare_voxconverse(
+    corpus_dir: Pathlike,
+    output_dir: Optional[Pathlike] = None,
+    split_test: bool = True,  # test part is larger than dev part - split it into dev and test by default
```
pzelasko, line 86: Is this a standard thing to do with this dataset? If not, it would be better to return the splits as defined by the creators by default.

flyingleafe:
@pzelasko yes, you're right, probably should not do the resplit by default, changed that

pzelasko (approved):
Thanks!

PR #1148, opened 2023-09-14
Fix download in LibriCSS recipe

desh2608:


pzelasko (approved):
Out of curiosity, how is this exactly different? Anyway, LGTM!

desh2608:
With the previous, I was getting an error since `corpus_zip` is a PosixPath type so `unzip()` does not work on it.

pzelasko:
Ohhh, right. I thought it's maybe a method on `Path` 🤦🏻‍♂️

PR #1155, opened 2023-09-17
fix path types in `get_duration()`

### LINKED ISSUE #1154: cli kaldi import fails with Attribution error in `get_duration`
relevant Traceback:
```cmd
"/lhotse/audio/backend.py", line 1042, in info
    if path.suffix.lower() == ".opus":
       ^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'suffix'


```

drahnreb:
Fixes #1154.

Background:
#1144 introduced a bug of how paths are handled.
Tests of `load_kaldi_data_dir` seem not to cover that code path when no reco2dur file is provided. This should be considered to be included.

### DIFF
--- a/lhotse/kaldi.py
+++ b/lhotse/kaldi.py
@@ -50,8 +50,8 @@ def get_duration(
     :param path: Path to an audio file or a Kaldi-style pipe.
     :return: float duration of the recording, in seconds or `None` in case of read error.
     """
-    path = str(path)
-    if path.strip().endswith("|"):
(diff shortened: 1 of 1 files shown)


pzelasko:
Thanks and sorry, there was not enough test coverage. However, the real issue is that info() cannot break for file-like objects. I will fix that in a separate PR -- I think the Kaldi related code does not need any modification.

pzelasko:
Should be resolved in #1156

PR #1157, opened 2023-09-18
`MixedCut.truncate`: fix the case when only `PaddingCut`s are left

flyingleafe:
I encountered another couple of bugs after simulating long conversations with `ConversationalMeetingSimulator`, cutting them into windows and loading the windows' audios:

- If a window happened to be inside an offset region without any cuts, a `PaddingCut` was returned for which the audio (zeros array) could not be loaded, since `num_samples` property was not defined;
- If it happened so that only `PaddingCut`s remained after truncation, this case was not properly handled (in the same way as "no cuts remained" case), which resulted in the `IndexError` exception at the last lines of `truncate` method (where we attempt to set SNR of first non-padding track to `None`)

This PR fixes those.

### DIFF
--- a/lhotse/cut/mixed.py
+++ b/lhotse/cut/mixed.py
@@ -497,14 +497,15 @@ class MixedCut(Cut):
                 )
             )
 
-        # Edge case: no tracks left after truncation. This can happen if we truncated an offset region.
+        # Edge case: no tracks with data left after truncation. This can happen if we truncated an offset region.
         # In this case, return a PaddingCut of the requested duration
-        if len(new_tracks) == 0:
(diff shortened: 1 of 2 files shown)


desh2608 (approved):

desh2608:
Thanks for catching this edge case. Can you also add a test which would have failed before making this change but passes now?

flyingleafe:
@desh2608 Sorry for such a delay, added the test.
The case with `IndexError` seems to be quite exotic - it replicates only when two or more `PaddingCut`s remained after truncation. Yet, apparently, cutting the output of `ConversationalMeetingSimulator` can lead to such cases.

desh2608 (approved):

desh2608:
Thanks. LGTM!

PR #1160, opened 2023-09-24
VAD workflow with Silero

rilshok:
## Description of Changes

We have added integration of the Silero VAD into the Lhotse project. This integration enables the use of Silero VAD for voice activity detection in audio recordings within Lhotse RecordingSets. The solution analyzes each track of every recording and stores the results in the SupervisionSet.

In the current change, we have introduced the following:
- A base class called `ActivityDetector` to allow for the addition of new activity detectors in the future.
- A runner class named `ActivityDetectionProcessor` for parallel execution of activity detection on the RecordingSet.
- Two classes, `SileroVAD8k` and `SileroVAD16k`, for the integration of Silero VAD.
- A workflow named `activity-detection` for running activity detection on the RecordingSet.

You can find the Silero VAD model for this integration on the [Silero VAD GitHub project](https://github.com/snakers4/silero-vad).

## Example usage in CLI

Prepare the model for work:

```bash
lhotse workflows activity-detection \
  --model-name silero-vad-16k \
  --chore
```

Run activity detection by Silero VAD:

```bash
lhotse workflows activity-detection \
  --model-name silero-vad-16k \
  --recordings-manifest data/librispeech_recordings_train-clean-5.jsonl.gz \
  --output-supervisions-manifest librispeech_recordings_train-clean-5.jsonl.gz \
  --jobs 2 \
  --device cpu
```

```bash
Loading recordings from data/librispeech_recordings_train-clean-5.jsonl.gz...
Making activity detection processor for 'silero-vad-16k'...
Running activity detection using 'silero-vad-16k'...
Using cache found in ~/.cache/torch/hub/snakers4_silero-vad_master
...
Detecting activities: 100%|████████████████| 1519/1519 [04:50<00:00,  5.22rec/s]
Saving 'silero-vad-16k' results ...
Results saved to:
.../librispeech_recordings_train-clean-5.jsonl.gz
```

## Example usage in code

```python
from lhotse.workflows.activity_detection.silero_vad import SileroVAD16k
from lhotse.audio import RecordingSet

vad = SileroVAD16k(device="cuda")

recordings = RecordingSet.from_file("data/librispeech_recordings_train-clean-5.jsonl.gz")
record = recordings[25]

vad(record)
```

```bash
[SupervisionSegment(id='6272-70171-0025-SileroVAD_16kHz-0-00000', recording_id='6272-70171-0025', start=0.194, duration=2.396, channel=0, text=None, language=None, speaker=None, gender=None, custom=None, alignment=None),
 SupervisionSegment(id='6272-70171-0025-SileroVAD_16kHz-0-00001', recording_id='6272-70171-0025', start=3.682, duration=1.02, channel=0, text=None, language=None, speaker=None, gender=None, custom=None, alignment=None),
 SupervisionSegment(id='6272-70171-0025-SileroVAD_16kHz-0-00002', recording_id='6272-70171-0025', start=4.994, duration=0.956, channel=0, text=None, language=None, speaker=None, gender=None, custom=None, alignment=None),
 SupervisionSegment(id='6272-70171-0025-SileroVAD_16kHz-0-00003', recording_id='6272-70171-0025', start=6.146, duration=2.652, channel=0, text=None, language=None, speaker=None, gender=None, custom=None, alignment=None),
 SupervisionSegment(id='6272-70171-0025-SileroVAD_16kHz-0-00004', recording_id='6272-70171-0025', start=9.122, duration=4.316, channel=0, text=None, language=None, speaker=None, gender=None, custom=None, alignment=None),
 SupervisionSegment(id='6272-70171-0025-SileroVAD_16kHz-0-00005', recording_id='6272-70171-0025', start=13.634, duration=3.006, channel=0, text=None, language=None, speaker=None, gender=None, custom=None, alignment=None)]
 ```

## Related Issues

Related tasks and links to discussions related to this integration:

- [#1041](https://github.com/lhotse-speech/lhotse/issues/1041) - Add Silero VAD integration

### CODE REVIEW

File: lhotse/bin/modes/workflows.py [hunk 1]
```diff
@@ -416,3 +416,120 @@ def simulate_meetings(
...
+    "-j",
+    "--jobs",
+    default=1,
+    help="Number of jobs for audio scanning.",
+)
+@click.option(
+    "--chore",
+    is_flag=True,
+    help="Prepare the model for work",
+)
```
desh2608, line 456: Could you describe this option a bit more? Is it basically downloading and preparing the model if it's the first time being used? If so, would it be possible to do it internally by default (similar to how model caching works in HuggingFace when you are using a pre-trained model)? This would avoid having to call the workflow CLI twice if the user wants to perform VAD.

lhotse/workflows/activity_detection/README.md
```diff
@@ -0,0 +1,156 @@
+# Activity Detection
+
```
desh2608, line 2: Great documentation!

lhotse/workflows/activity_detection/base.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @property
+    def device(self) -> str:
+        return self._device
+
+    def __call__(self, recording: Recording) -> List[SupervisionSegment]:
+        resampled = recording.resample(self._sampling_rate)
+        audio = resampled.load_audio()  # type: ignore
+
+        uid_template = "{recording_id}-{detector_name}-{channel}-{number:05}"
```
desh2608, line 42: Perhaps we can remove the detector name from the segment ID to keep it concise? Users can always just save the manifest file with the appropriate name.

desh2608 (commented):
Thanks for this contribution! Left some minor comments.

Out of curiosity, do you have any speed benchmarks? For example, if I were to run the VAD on a 1 hour recording with 1 GPU, how long would it take?

csukuangfj:
> Thanks for this contribution! Left some minor comments.
> 
> 
> 
> Out of curiosity, do you have any speed benchmarks? For example, if I were to run the VAD on a 1 hour recording with 1 GPU, how long would it take?

I think it is faster on CPU for silero VAD. 

Silero VAD uses LSTM and you need to process the file sequentially.

lhotse/bin/modes/workflows.py
rilshok: For the Silero VAD model, the state is automatically downloaded from git-hub via the `torch.hub` interface when an instance of the `lhotse.workflows.acivity_detection:SileroVAD` class is first created on the host and cached in the user's environment. This process works automatically if the conditions are met: either the model state is already cached, or the class initialization occurs initially in **only** one process.

However, when running workflow in multiprocessor mode, a situation may arise where the first process starts downloading the cached model state via torch.hub, and the second process mistakenly assumes that the model is already cached and starts deploying. This will cause an error in the second process, which will eventually cause the first process to fail. As a result, the first process will not have time to fully download the model state from git-hub, resulting in cache invalidation. Further, the model will become unusable.

To eliminate such situations, the base class provides the function `ActivityDetector.chore` without arguments, which works as an emergency reset or hard reset. It is important to note that its use is completely optional, especially on hosts where this model is regularly used. It is recommended to use this option only in critical situations.

I suggest leaving this script option and functionality in place to potentially save time for a user who may encounter future model cache invalidations.

lhotse/workflows/activity_detection/base.py
rilshok: I suggest to leave the `detector_name` in the segment identifier to allow mapping of segments created by `SileroVAD` to segments of other detectors that may be implemented in the future. 

Suppose you want to use a detector that detects the sound of a cell phone ringing. Currently, you can easily create your own detector by inheriting it from the `lhotse.workflows.acivity_detection:ActivityDetector` class and overriding the `forward` method to make it fit your needs. You can then take advantage of all the functionality associated with the activity detection task. For example, you can use your detector to segment the dataset, converting the results into a `SupervisionSegment`. This functionality is already implemented in the `__call__` method of the ActivityDetector class. 

Thus, storing the detector name in the segment identifier provides flexibility for further expansion and integration of other detectors.

rilshok:
I haven't made precise performance measurements. Silero VAD authors claim that processing one audio fragment (30+ ms) takes less than 1 ms on a single CPU thread. My own practice confirms this performance. However, it is worth noting that this value may not be correct for a single thread. When processing in parallel on a CPU with two threads (with two model states), Torch fully loads my CPU, and adding more threads does not bring a significant performance gain. On my inexpensive GPU, however, parallel processing resulted in about a 3x performance increase when running multiple processes. I would be grateful for contributions to document the performance of this model.

lhotse/bin/modes/workflows.py
desh2608: Shouldn't you first download the model (or check if it's already cached) **before** creating multiple processes? This would ensure that all the processes don't need to download it separately.

rilshok: I think that to ensure reliability in this case you would have to delete and reload the model from `torch.hub` every time you run the workflow. Unfortunately, this is not a good solution, because after N requests torch.hub will stop giving permission to load the model state again. The user will be banned by ip due to too many requests. Do you have a suggestion on how exactly we can better implement the model caching feature?

rilshok: We can try to do a test inference before running the workflow, if it is not successful then do all the chore, and only then run the parallel processing. Do you think that would be a good solution? I can do that, even though I don't like it. I think that unfortunately without a test infierence, we can't be sure that the model in the cache is okay.

desh2608: Currently, your `chore` method deletes an existing cache of the model and downloads it every time it is used. Instead, I would suggest having a `--force-download` option which does this. If this option is not specified, simply use the existing model in the cache. (From a consistency perspective, such a `force_download` option is also what we use in most of the existing Lhotse recipes.)

To play well with multi-processing, the `ActivityDetectionProcessor` should first check and download (if needed) the model before creating the processes.

rilshok: I like this solution. i will rename the `chore` method to `force_download` and implement the corresponding option for our workflow.

lhotse/workflows/activity_detection/base.py
rilshok: Are any changes required on this issue?

lhotse/bin/modes/workflows.py
rilshok: I've made the changes

lhotse/workflows/activity_detection/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+## Example usage in command line interface (CLI)
+
+1. Prepare the model for work:
+    ```bash
+    lhotse workflows activity-detection \
+    --model-name silero_vad_16k \
+    --chore
```
pzelasko, line 17: Update this doc to be consistent with later changes in this PR

lhotse/workflows/activity_detection/base.py
```diff
@@ -0,0 +1,126 @@
...
+                    detector_name=self._detector_name,
+                    channel=channel,
+                    number=i,
+                )
+                segment = SupervisionSegment(
+                    id=uid,
+                    recording_id=recording.id,
+                    start=activity.start,
+                    duration=activity.duration,
+                    channel=channel,
```
pzelasko, line 55: FYI you could leverage a custom field here to tag which vad created the segment, i.e.:

```python
SupervisionSegment(
    ...,
    custom={"vad_tag": detector_name},
)
```

but the current solution is also OK with me

lhotse/workflows/activity_detection/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _process_recording(self, record: Recording) -> List[SupervisionSegment]:
+        pid = multiprocessing.current_process().pid
+        detector = self._detectors[pid]
+        return detector(record)
+
+    def __call__(self, recordings: RecordingSet) -> SupervisionSet:
+        pool = ProcessPoolExecutor(
+            max_workers=self._num_jobs,
+            initializer=self._init_detector,
```
pzelasko, line 97: nice trick, I didn't know that and always hacked `@lru_cache` instead :)

pzelasko (approved):
Thanks for your contribution, it looks great! Can you fix that doc comment before we merge?

lhotse/bin/modes/workflows.py
desh2608: Great, thanks!

rilshok:
The tests fail on an older version of torch because of the trust_repo argument passed to trust.hub. I'm working on it.

lhotse/workflows/activity_detection/README.md
rilshok: The documentation has been updated to clarify the purpose of the `--force_download` flag

rilshok:
The testing environment uses `torch==1.12.1` `torchaudio==0.12.1` `torchdata==0.4.1`, for the case of python 3.10. But it uses more recent versions of torch packages on older versions of python. Is this correct?

pzelasko (approved):

pzelasko (approved):
LGTM

rilshok:
Do we have a plan how to solve the test problem in py 3.11 environment? The problem seems to come from outside and is not related to my changes.

rilshok:
I tried to reproduce the problem with this test, unfortunately it does not reproduce for me.

lhotse/bin/modes/workflows.py
```diff
@@ -416,3 +419,129 @@ def simulate_meetings(
...
+@click.option(
+    "-j",
+    "--jobs",
+    default=1,
+    help="Number of jobs for audio scanning.",
+)
+@click.option(
+    "--force_download",
+    is_flag=True,
+    help="Prepare the model for work",
```
desh2608, line 458: Perhaps this help string should be updated?

lhotse/workflows/activity_detection/README.md
```diff
@@ -0,0 +1,162 @@
...
+    ```
+
+2. Please note: if you are using the activity-detection script for the first time with any detection model, be sure to add the `--force_download` flag to preload the model state on your local machine from the appropriate source. Like this:
+
+    ```bash
+    lhotse workflows activity-detection \
+    --model-name silero_vad_16k \
+    ...\
+    --force_download
+    ```
```
desh2608, line 25: This is not really what I meant about the `force_download`. I meant that the user would just run the command once, with or without this optional flag.

**Case 1:** If the flag is not set, then use existing downloaded model, or download if it does not exist.

**Case 2:** If the flag is set, clear cache and re-download model before starting the VAD processing.

rilshok: Yes, I really misinterpreted that. I've made the appropriate change.

PR #1170, opened 2023-10-04
PrecomputedFeatures: take the durations in `x` and `x_lens` from same source

vesis84:
Hi Piotr,
recently, i observed a bug that was caused by tiny inconsistency between values in `Cut.num_frames` and `Cut.supervisions[0].duration`.

The log-print of the error was this:
```
  File "/mnt/matylda5/iveselyk/CNECT_TENDER/MODEL_BUILDS/k2/k2_streaming_training-czech_v3_punct-truecase_downscaled/training/pruned_transducer_stateless7_streaming/model.py", line 118, in forward
    assert x.size(1) == x_lens.max().item(), (x.shape, x_lens, x_lens.max())
AssertionError: (torch.Size([71, 563, 80]), tensor([562, 562, 562, 562, 562, 561, 561, 560, 560, 559, 558, 558, 558, 556,
        556, 555, 554, 554, 554, 554, 554, 553, 553, 552, 552, 551, 551, 550,
        550, 548, 548, 547, 546, 544, 544, 543, 542, 542, 542, 540, 540, 540,
        538, 538, 535, 533, 532, 531, 531, 531, 529, 529, 529, 528, 528, 527,
        526, 526, 524, 524, 524, 523, 523, 522, 522, 522, 522, 521, 521, 520,
        520], device='cuda:3', dtype=torch.int32), tensor(562, device='cuda:3', dtype=torch.int32))
```

By diving into the code I realized that the feature files are loaded into `x` with lenghts taken from `Cut.num_frames` in [collate_features()](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/collation.py#L131C1-L131C1) .
But for the `x_lens` the lengths are computed from `Cut.supervisions[0].duration` in [PrecomputedFeatures::supervision_intervals()](
https://github.com/lhotse-speech/lhotse/blob/287de1dbadcb70bf7398d3d482c1f730d2c5407f/lhotse/dataset/input_strategies.py#L145) via [utils::supervision_to_frames()](https://github.com/lhotse-speech/lhotse/blob/287de1dbadcb70bf7398d3d482c1f730d2c5407f/lhotse/utils.py#L732) and via [utils::compute_num_frames()](https://github.com/lhotse-speech/lhotse/blob/287de1dbadcb70bf7398d3d482c1f730d2c5407f/lhotse/utils.py#L390) .

The problem disappears after the change proposed in this PR.
However, the question is if this might back-fire in some way or not,
and if it is a good way to solve it this way...

Other solution is to check CutSet manifest and remove the utterances where `Cut.num_frames` and `compute_num_frames(Cut.supervisions[0].duration, ...)` disagree.

Piotr, what would be your view on this ?

Thanks & cheers,
Karel

### DIFF
--- a/lhotse/dataset/input_strategies.py
+++ b/lhotse/dataset/input_strategies.py
@@ -148,7 +148,14 @@ class PrecomputedFeatures(BatchIO):
                 for sup in cut.supervisions
             )
         )
+
         sequence_idx = [i for i, c in enumerate(cuts) for s in c.supervisions]
+
+        # Take `nums_frames` from `cut.num_frames` not from `cut.supervisions`
+        # -> `cut.num_frames` is also used in `collate_features()` for loading feature matrices
+        # -> this is to avoid +/- 1 frame error for `x`, `x_lens` in `Zipformer::forward()`
+        nums_frames = [cut.num_frames for cut in cuts]
+
         return {
             "sequence_idx": torch.tensor(sequence_idx, dtype=torch.int32),
             "start_frame": torch.tensor(start_frames, dtype=torch.int32),


pzelasko:
Hi Karel! 

I don't think this is the right solution because in general you might have valid data with `cut.duration != supervision.duration` (or even multiple supervisions per cut). We have been resolving some inconsistencies with Kaldi imported data in the past so it makes sense this could be the source of the issue. Is the data speed perturbed in Kaldi or in Lhotse?

Maybe it makes sense to add an extra post-processing step that compares `cut.num_frames` with `compute_num_frames(sup)` for Kaldi imported data and adjusts the timestamps (although I'm more interested in why they were different in the first place).

desh2608:
Can this be fixed by just calling `fix_manifests()` after importing manifests from Kaldi?

vesis84:
Hi,
this PR can be cancelled. I found out the "root" of the issue.

In the `Cut` there was having a tiny inconsistency of `Cut.duration` and `Cut.supervisoions[0].duration`.
The `Cut.duration` was coming from measuring the length of the audio file (say a 3.144 value),
and the `Cut.supervisoions[0].duration` was coming from `segments` file in kaldi-data dir (say a 3.14 value).

This inconsistency can be fixed by calling `trim_to_supervisions()` over a CutSet right before doing feature extraction.
In this case the two values become equal, and the problem simply vanishes ;-)

In my local code, I added a new check to `local/validate_cutset_manifest.py` (originally `local/validate_manifest.py`)
to ensure that the two duration values match exactly.

Should I add it also somewhere to Icefall ?
Btw, would you be interested in having a "generic" recipe in Icefall for datasets imported from kaldi ?
(I have that already developed, but it is not published...)

K.

vesis84:
And the speed perturbation was done by lhotse, to answer Piotr's quesion.

pzelasko:
I think it’d make sense to add a fix to supervision durations in Kaldi import utility if it can be detected in a reliable way. Generic Kaldi imported data recipe in Icefall also sounds cool although there’s probably a few hparams that’d need to be tuned depending on the dataset size and language(s), don’t you think?

vesis84:
Concerning the kaldi-import, the `trim_to_supervisions` resolved the problem. 
The problem was in my code, not always calling it...

Concerning the generic Icefall recipe, the usual hyperparameters are "number of epochs",
"base-lr" and the decode-epoch based on the minimum in the "tensorboard" graph ?
The "default" size of the model can be the same. Any other ideas on what is to be tuned ?
(it can be mentioned in the docs)

danpovey:
We don't really put much stock in the tensorboard graph minimum (I suppose you mean validation); we normally
just test different epochs - ideally on validation data, of course.
For recent recipes you shouldn't have to tune base-lr much, and you can just set epochs to 40 to 60 and reduce
it if best WER is encountered before the end.

PR #1171, opened 2023-10-04
Move window tensor to proper device

hhatefi:
When running inference on GPU, `Wav2Win._forward_strided` fails if `x_strided` and `self._window` are on different device. Therefore, `self._window` should move into the same device as that of `x_strided`

### DIFF
--- a/lhotse/features/kaldi/layers.py
+++ b/lhotse/features/kaldi/layers.py
@@ -167,7 +167,7 @@ class Wav2Win(nn.Module):
             x_strided = x_strided - self.preemph_coeff * x_offset[:, :, :-1]
 
         # Apply window_function to each frame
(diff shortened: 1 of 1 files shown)


pzelasko:
You can move the Wav2Win (or any other module in that file) to device before running the inference. If you're using `Fbank`, there's an option in the config to place it on the gpu.

hhatefi:
The problem it not the module, it is `self._window` which is created by `create_frame_window` which does not set the device type. Even if the module is loaded into gpu, forward call will fail, if `self._window` is not in gpu.

pzelasko:
Wouldn't `module = module.to("cuda")` solve the issue by placing every registered parameter and buffer on the GPU?

PR #1174, opened 2023-10-09
[WIP] spokewoz recipe

oplatek:
Hi @desh2608,

I created a recipe for https://spokenwoz.github.io/SpokenWOZ-github.io/

I am just testing it out. The download and prepare functions seem to work, and I prepared `RecordingSet` and SupervisionSet` manifests.

However, I was surprised that the following code returns `MultiCut[s]`. 
```
dev_recs = RecordingSet.from_file(manifest_dir / "spokenwoz_recordings_dev.jsonl.gz")
dev_sups = SupervisionSet.from_file(manifest_dir / "spokenwoz_supervisions_dev.jsonl.gz")
cuts_dev = CutSet.from_manifests(dev_recs, dev_sups)
```

Is MultiCut a good option for SpokenWOZ, where the system and user side of task-oriented dialogue are separated into two channels?

Personally, I do NOT want to tie together "system" and "user" turns together.
I just want to be able to traverse the whole conversations, which is possible thanks to SupervisionSegment naming where id is `f"conversationID-{turn_number:03d}"`.

The practical aspect is that cuts_dev[0].plot_audio() takes ages for multicut of 183s. It took 10+min, before I restarted the jupyter kernel.

Th

### DIFF
--- a/lhotse/bin/modes/recipes/__init__.py
+++ b/lhotse/bin/modes/recipes/__init__.py
@@ -62,6 +62,7 @@ from .primewords import *
 from .rir_noise import *
 from .speechcommands import *
 from .spgispeech import *
+from .spokenwoz import *
 from .stcmds import *
 from .switchboard import *
 from .tal_asr import *
--- a/lhotse/bin/modes/recipes/spokenwoz.py
+++ b/lhotse/bin/modes/recipes/spokenwoz.py
@@ -1,0 +1,62 @@
+from typing import Sequence
+
+import click
+
+from lhotse.bin.modes import download, prepare
+from lhotse.recipes import download_spokenwoz, prepare_spokenwoz
+from lhotse.utils import Pathlike
+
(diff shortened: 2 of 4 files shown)


desh2608:
I am not familiar with this dataset, but I suppose the recordings contain 2 channels, which is why the resulting cut is a MultiCut. If the `SupervisionSegment`s contain the appropriate channel information, you would eventually get MonoCut objects when you run `cut.trim_to_supervisions(keep_overlapping=False, keep_all_channels=False`. You can then work with these resulting MonoCuts instead of the original MultiCut.

pzelasko:
In addition to what Desh stated: `plot/play_audio` does not actually support multi-channel data (yet). The reason it was slow for you is because matplotlib plot received a 2-d array `(num_channels, num_samples)` but it expected a 1-d array `(num_samples,)` (this is not the right way to use 2-d input in matplotlib). We should fix that at some point.

Issue #1177, opened 2023-10-12
404 not found error in `lhotse download ami --mic mdm ./` 


JinZr:
After download certain number of utterances the command ran into this error. 
```
2023-10-08 22:25:59,639 INFO [utils.py:510] Server does not support range requests - attempting downloading from scratch████████████████████████████████████████████████████████████████████████████████ | 56.6M/57.0M [00:37<00:00, 2.10MB/s]
wav_db/IB4001/audio/IB4001.Array2-04.wav: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 57.0M/57.0M [00:36<00:00, 1.56MB/s]
Downloading AMI meetings: 76it [22:12:58, 1052.35s/it]██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▍| 56.8M/57.0M [00:36<00:00, 1.53MB/s]
Traceback (most recent call last):
  File "/Users/zengruijin/miniconda3/bin/lhotse", line 8, in <module>
    sys.exit(cli())
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/lhotse/bin/modes/recipes/ami.py", line 122, in ami
    download_ami(
  File "/Users/zengruijin/miniconda3/lib/python3.10/site-packages/lhotse/recipes/ami.py", line 254, in download_ami
[...]

PR #1178, opened 2023-10-12
minor fix for the AMI recipe

JinZr:
related to issue #1177

### DIFF
--- a/lhotse/recipes/ami.py
+++ b/lhotse/recipes/ami.py
@@ -201,8 +201,12 @@ def download_audio(
             )
         elif mic == "mdm":
             for array in MDM_ARRAYS:
-                for channel in MDM_CHANNELS:
-                    wav_name = f"{item}.{array}-{channel}.wav"
+                if item == "IB4001" and array == "Array2":
+                    channel_num = 4
+                else:
+                    channel_num = 8
+                for channel_index in range(channel_num):
+                    wav_name = f"{item}.{array}-{MDM_CHANNELS[channel_index]}.wav"
                     wav_url = f"{url}/AMICorpusMirror/amicorpus/{item}/audio/{wav_name}"
                     wav_dir = target_dir / "wav_db" / item / "audio"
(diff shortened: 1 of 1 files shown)


lhotse/recipes/ami.py
```diff
@@ -201,8 +201,12 @@ def download_audio(
             )
         elif mic == "mdm":
             for array in MDM_ARRAYS:
-                for channel in MDM_CHANNELS:
-                    wav_name = f"{item}.{array}-{channel}.wav"
+                if item == "IB4001" and array == "Array2":
```
desh2608, line 204: You could also just put a try/except block here and warn the user for failing downloads.

JinZr: good call! this would save a lot of effort.

lhotse/recipes/ami.py
```diff
@@ -207,9 +207,14 @@ def download_audio(
...
                     wav_dir.mkdir(parents=True, exist_ok=True)
                     wav_path = wav_dir / wav_name
-                    resumable_download(
-                        wav_url, filename=wav_path, force_download=force_download
-                    )
+                    try:
+                        resumable_download(
+                            wav_url, filename=wav_path, force_download=force_download
+                        )
+                    except:
```
desh2608, line 214: PEP8 does not allow a bare `except` clause (see [this](https://stackoverflow.com/questions/14797375/should-i-always-specify-an-exception-type-in-except-statements/14797508#14797508) for why you should specify an exception type).

JinZr: ok thanks! the next commit should fix the issue

desh2608: Thanks. Is this ready to merge?

JinZr:
i just ran through the process again, it is ready to be merged now.

thanks Desh!

desh2608 (approved):

desh2608:
Thanks for the fix!

PR #1189, opened 2023-10-16
support far field data for icmcasr challenge

yuekaizhang:


lhotse/recipes/icmcasr.py
```diff
@@ -15,75 +15,98 @@
...
+                    speaker = dataline.split('"')[1].strip()
+                elif "xmin =" in dataline:
+                    start = float(dataline.split("=")[1].strip())
+                elif "xmax =" in dataline:
+                    end = float(dataline.split("=")[1].strip())
+                elif "text" in dataline:
+                    text = dataline.split('"')[1].strip()
+                    if len(text) > 0:
+                        if float(recordings[-1].duration) < end:
+                            duration = float(recordings[-1].duration) - start
```
desh2608, line 85: It seems this block is parsing a `textgrid` formatted record. It might be better to use standard package for this --- check out the alimeeting recipe for an example.

yuekaizhang: Thanks. Done.

desh2608 (approved):
Thanks, LGTM.

PR #1192, opened 2023-10-18
Prevent accidental renaming when using with_suffix

### LINKED ISSUE #1183: LilcomChunkyWriter replaces storage_path if it contains "."
Hi! I realize that when i use storage_path with . like "gigaspeech_overlap_0.3_TRAIN" it will end up becoming saving to the file "gigaspeech_overlap_0.lca" instead of "gigaspeech_overlap_0.3_TRAIN.lca" as expected. 
It seems to be due to this line:
https://github.com/lhotse-speech/lhotse/blob/afc8eff1d1eaeeae42e0a628e1981bb231527340/lhotse/features/io.py#L829
which replaces the words after "." with lca instead of appending. 
I saw quite a few other places using `.with_suffix` as well so will have the same issue as well.

chiiyeh:
Fixes #1183 

The use of `.with_suffix` replaces the strings after the `.`, which leads to issues when the key/path contains `.` in the name.

### DIFF
--- a/lhotse/features/io.py
+++ b/lhotse/features/io.py
@@ -281,7 +281,9 @@ class LilcomFilesWriter(FeaturesWriter):
         subdir = self.storage_path_ / key[:3]
         subdir.mkdir(exist_ok=True)
         p = subdir / key
-        output_features_path = p.with_suffix(p.suffix + ".llc" if p.suffix != ".llc" else ".llc")
+        output_features_path = p.with_suffix(
+            p.suffix + ".llc" if p.suffix != ".llc" else ".llc"
+        )
         serialized_feats = lilcom.compress(value, tick_power=self.tick_power)
         with open(output_features_path, "wb") as f:
             f.write(serialized_feats)
@@ -345,7 +347,9 @@ class NumpyFilesWriter(FeaturesWriter):
         subdir = self.storage_path_ / key[:3]
         subdir.mkdir(exist_ok=True)
         p = subdir / key
-        output_features_path = p.with_suffix(p.suffix + ".npy" if p.suffix != ".npy" else ".npy")
+        output_features_path = p.with_suffix(
+            p.suffix + ".npy" if p.suffix != ".npy" else ".npy"
+        )
         np.save(output_features_path, value, allow_pickle=False)
         # Include sub-directory in the key, e.g. "abc/abcdef.npy"
         return "/".join(output_features_path.parts[-2:])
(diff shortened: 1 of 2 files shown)


lhotse/features/io.py
```diff
@@ -280,7 +280,8 @@ def write(self, key: str, value: np.ndarray) -> str:
         # too many files in a single directory.
         subdir = self.storage_path_ / key[:3]
         subdir.mkdir(exist_ok=True)
-        output_features_path = (subdir / key).with_suffix(".llc")
+        p = subdir / key
+        output_features_path = p.with_suffix(p.suffix + ".llc" if p.suffix != ".llc" else ".llc")
```
desh2608, line 284: Is the if/else needed here? (Same for all below changes)

desh2608 (commented):
Could you also add a test for this change?

lhotse/features/io.py
chiiyeh: The if/else is to prevent adding of suffix if  `p` contains the suffix for some reason say "test.llc". So it will not become "test.llc.llc". 
Which is the expected behavior mentioned here 
https://github.com/lhotse-speech/lhotse/blob/master/lhotse/features/io.py#L443

chiiyeh:
I am not too sure what sort of test is needed, is it those similar to the one in the main testing folder like this:
https://github.com/lhotse-speech/lhotse/blob/master/test/features/test_chunky_writer.py#L31-L35
Also for the code referred in this test, not sure if there is a bug, as the `writer` variable is used at the `get_reader` outside of the `with` that it is created.

pzelasko:
A test similar to that one would do. The goal is to cover a scenario in which the test would have failed before the change, but passes after it. You might want to use either `tmpdir` pytest fixture or python's `NamedTemporaryFile` to create a tmp file that has a problematic name. 

> Also for the code referred in this test, not sure if there is a bug, as the writer variable is used at the get_reader outside of the with that it is created.

That test is correct, `writer` still exists after `with` (at which point `writer.__exit__()` is being called), and we're only fetching the `.name` property to know which type of reader to create.

chiiyeh:
Thanks for the clarification. Have added in the test. Tested it on my side, should be working.

pzelasko (dismissed):
Great work, thanks!

pzelasko (dismissed):

chiiyeh:
Gosh.. so sorry.. had trouble using the isort on my vscode.. and missed one of the changes.

pzelasko (approved):

Issue #1186, opened 2023-10-16
`trim-to-alignments` produces negative utterance duration on libriheavy


JinZr:
Trimming the CutSet `libriheavy_cuts_small_ali_proc.jsonl.gz` would produce `libriheavy_cuts_small_ali_proc_trimmed.jsonl.gz` containing utterances with negative duration, attachments can be found [here](https://drive.google.com/drive/folders/1Aq1qHtfpQszTZpGhAXo8CMJPC7PZsyQB?usp=sharing).

Command I used:
```
lhotse cut trim-to-alignments --keep-all-channels --type word --max-pause 0.2 libriheavy_cuts_small_ali_proc.jsonl.gz libriheavy_cuts_small_ali_proc_trimmed.jsonl.gz
```

pzelasko:
@desh2608 could you take a look?

PR #1193, opened 2023-10-18
Fix for `trim_to_alignments` issue

desh2608:
Ref: https://github.com/lhotse-speech/lhotse/issues/1186

The key issue was that in a couple of places, we did not account for the fact that start time of `AlignmentItem` is w.r.t. the start of the recording, whereas that for `SupervisionSegment` is w.r.t. the Cut.

* To create supervision segment from an alignment item, we need to subtract the start time of the cut.
* In `with_offset` function for the supervision, alignment start times do not change, since they are w.r.t. the recording.


test/cut/test_cut_trim_to_supervisions.py
```diff
@@ -363,15 +363,15 @@ def test_cut_trim_to_supervisions_keep_overlapping_extend(mono_cut):
     # Extended on the right side only by (4.0 - 3.37) / 2 == 0.315;
     # the left side is capped by the start of the recording.
     assert len(c1.supervisions) == 1
-    assert c1.start == 0.0
+    assert c1.start == 1.0
```
pzelasko, line 366: After changing the `mono_cut` fixture, this test case no longer represents the illustrated scenario. Also the comment above seems to be in conflict with the test outcome (looks like `c1` should have 4 seconds now that it's not capped by the start of the recording anymore)

desh2608: Good point. I'll fix the tests tomorrow.

desh2608: Sorry I lost track of this. Have made the appropriate changes now --- should not affect the older tests.

pzelasko (approved):

Issue #1203, opened 2023-11-04
Supervision frame mask not handling negative end time


ArthLeu:
https://github.com/lhotse-speech/lhotse/blob/c5f26afd100885b86e4244eeb33ca1986f3fa923/lhotse/cut/set.py#L2908-L2922

`CutSet.cut_into_windows()` produces negative start times in returned `Supervision` and `Alignment` items. But when computing the mask using supervision_frame_masks, the end time `et` (assumed obtained from start + duration) does not handle this case where alignment ends before cut start, e.g. `et <= 0`. 

Please verify if this is the case and resolve if needed.

pzelasko:
I see. I think your PR makes sense to help resolve this. You may also want to use `keep_excessive_supervisions=False` for such cases.

ArthLeu:
I tried it. But it would remove an entire Supervision altogether, regardless of whether a portion of its Alignment items are within the Cut time range.

desh2608:
I made some changes in alignment start times recently (https://github.com/lhotse-speech/lhotse/pull/1193). Not sure if it is related, but may be worth checking if it resolved your issue?

ArthLeu:
Want to understand this. With #1193, the expectation is to unify all the Supervision and Alignment timestamps to absolutes (w.r.t. Recording start), and no more relatives (w.r.t. Cut start)?

desh2608:
@ArthLeu Alignments were always supposed to have absolute start times (w.r.t. the recording), but there was a bug in the implementation of some functions where they were being modified.

PR #1204, opened 2023-11-04
#1203 Update supervision_frame_masks() to skip negative end indices

ArthLeu:
When flipping mask digits to 1.0, negative indices should never be used.

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -2912,6 +2912,7 @@ def compute_supervisions_frame_mask(
                     if ali.end < cut.duration
                     else num_frames
                 )
+                if et <= 0: continue
                 mask[st:et] = 1.0
         else:
             st = round(supervision.start / frame_shift) if supervision.start > 0 else 0
@@ -2920,6 +2921,7 @@ def compute_supervisions_frame_mask(
                 if supervision.end < cut.duration
                 else num_frames
             )
+            if et <= 0: continue
             mask[st:et] = 1.0
     return mask
 


pzelasko:
Right. I triggered the tests, if they're green, LGTM. I wonder how did you run into this edge case.

ArthLeu:
It wants some newlines. Can this automatically apply the changes or do I need to manually change them?

desh2608:
You would need to run black formatting. See development instructions here: https://github.com/lhotse-speech/lhotse#development-installation

pzelasko:
Please follow these instructions, pre-commit will auto-apply relevant changes, but you'll need to commit them yourself.

https://github.com/lhotse-speech/lhotse#development-installation

ArthLeu:
Sorry I was busy with something else. I upgraded to Lhotse 1.18, and seems like the `Cut.supervision_frame_masks()` now uses *relative* time to determine mask region. 

So this means when we need to `use_alignment_if_exists`, the alignment timestamps should also be relative? I tried absolute alignment timestamps, which worked with 1.17, and does not work with 1.18.

desh2608:
> Sorry I was busy with something else. I upgraded to Lhotse 1.18, and seems like the `Cut.supervision_frame_masks()` now uses _relative_ time to determine mask region.
> 
> So this means when we need to `use_alignment_if_exists`, the alignment timestamps should also be relative? I tried absolute alignment timestamps, which worked with 1.17, and does not work with 1.18.
> 
> Want to confirm this before making any further change.
> 
> Probably related to #1193

That PR is a bug fix, not a change in behavior. Alignments were always supposed to have absolute start times (w.r.t. the recording), whereas supervisions are supposed to be relative (w.r.t the cut).

PR #1205, opened 2023-11-06
Modify SpeechSynthesisDataset class, make it return text

yaozengwei:
This PR is required by the TTS recipe https://github.com/k2-fsa/icefall/pull/1372 in icefall. In this TTS recipe, we convert the transcript text to phonemes in training.

### DIFF
--- a/lhotse/dataset/speech_synthesis.py
+++ b/lhotse/dataset/speech_synthesis.py
@@ -19,26 +19,35 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
         {
             'audio': (B x NumSamples) float tensor
             'features': (B x NumFrames x NumFeatures) float tensor
-            'tokens': (B x NumTokens) long tensor
+            'tokens': (B x NumTokens) long tensor  # when return_tokens=True
             'audio_lens': (B, ) int tensor
             'features_lens': (B, ) int tensor
-            'tokens_lens': (B, ) int tensor
+            'tokens_lens': (B, ) int tensor  # when return_tokens=True
+            'text': List[str] of len B  # when return_tokens=False
+            'cut': List of Cuts  # when return_cuts=True
         }
     """
 
     def __init__(
         self,
-        cuts: CutSet,
+        return_cuts: bool = False,
+        cuts: CutSet = None,
         cut_transforms: List[Callable[[CutSet], CutSet]] = None,
         feature_input_strategy: BatchIO = PrecomputedFeatures(),
         feature_transforms: Union[Sequence[Callable], Callable] = None,
+        return_tokens: bool = True,
         add_eos: bool = True,
         add_bos: bool = True,
     ) -> None:
         super().__init__()
+        self.return_cuts = return_cuts
 
-        self.cuts = cuts
-        self.token_collater = TokenCollater(cuts, add_eos=add_eos, add_bos=add_bos)
+        self.return_tokens = return_tokens
+        if return_tokens:
+            assert cuts is not None, "cuts is required when return_tokens=True"
+            self.cuts = cuts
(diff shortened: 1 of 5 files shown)


lhotse/dataset/speech_synthesis.py
```diff
@@ -19,26 +19,35 @@ class SpeechSynthesisDataset(torch.utils.data.Dataset):
...
+            'tokens_lens': (B, ) int tensor  # when return_tokens=True
+            'text': List[str] of len B  # when return_tokens=False
+            'cut': List of Cuts  # when return_cuts=True
         }
     """
 
     def __init__(
         self,
-        cuts: CutSet,
+        return_cuts: bool = False,
```
pzelasko, line 33: This change is breaking, can you move `return_cuts` to some position towards the end of parameter list?

pzelasko (commented):
Looks good, but don't you prefer to have G2P inside `SpeechSynthesisDataset` to avoid spending the time running it inside the training loop?

lhotse/dataset/speech_synthesis.py
yaozengwei: OK.

yaozengwei:
> Looks good, but don't you prefer to have G2P inside `SpeechSynthesisDataset` to avoid spending the time running it inside the training loop?

Hi, @pzelasko,  I prefer to do the text normalization and tokenization in separate training recipes, since they usually depend on different packages. @csukuangfj suggests doing this in data preparation stage.

pzelasko:
> > Looks good, but don't you prefer to have G2P inside `SpeechSynthesisDataset` to avoid spending the time running it inside the training loop?
> 
> Hi, @pzelasko, I prefer to do the text normalization and tokenization in separate training recipes, since they usually depend on different packages. @csukuangfj suggests doing this in data preparation stage, with converted phonemes saved to manifests.

OK cool. It looks like there are some conflicts after merging the other PR with multi-speaker support, could you resolve them?

yaozengwei:
> > > Looks good, but don't you prefer to have G2P inside `SpeechSynthesisDataset` to avoid spending the time running it inside the training loop?
> > 
> > 
> > Hi, @pzelasko, I prefer to do the text normalization and tokenization in separate training recipes, since they usually depend on different packages. @csukuangfj suggests doing this in data preparation stage, with converted phonemes saved to manifests.
> 
> OK cool. It looks like there are some conflicts after merging the other PR with multi-speaker support, could you resolve them?

Ok. Thanks. I have some local changes and will resolve the conflicts later. 

In addition, in current implementation, it will load the whole cuts set and generate a char-based vocabulary according to the given texts. But I think TTS recipes usually use phoneme tokens instead. Also, if might cause token-id-mapping mismatch bettwen training and test, when separately given two cuts sets for this class. Shall we remove this and make this calss return raw text and optionally pre-converted (phoneme) tokens? 

https://github.com/lhotse-speech/lhotse/blob/706117511588bf736601e034dc052e0426cbb3b0/lhotse/dataset/speech_synthesis.py#L43

pzelasko:
> In addition, in current implementation, it will load the whole cuts set and generate a char-based vocabulary according to the given texts. But I think TTS recipes usually use phoneme tokens instead. Also, if might cause token-id-mapping mismatch bettwen training and test, when separately given two cuts sets for this class. Shall we remove this and make this calss return raw text and optionally pre-converted (phoneme) tokens?

I see that this class accepts `cuts` in constructor, it's an outdated design we don't use anymore, so thanks for updating it. Instead, it would make sense to pass a `tokenizer` object/callable that (optionally) converts raw text into some tokens. What you're suggesting also works.

yaozengwei:
@pzelasko Thanks. It is ready to be merged.

yaozengwei:
@pzelasko @desh2608 Would you mind giving a review if you have time?

PR #1206, opened 2023-11-06
SpeechSynthesisDataset returns `speaker_ids` 

JinZr:


lhotse/dataset/speech_synthesis.py
```diff
@@ -65,15 +69,21 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, torch.Tensor]:
...
             "audio_lens": audio_lens,
             "features_lens": features_lens,
             "tokens_lens": tokens_lens,
         }
+        if self.speaker_id_mapping is not None:
+            batch["speakers"] = torch.tensor(
+                [self.speaker_id_mapping[cut.supervisions[0].speaker] for cut in cuts],
+                dtype=torch.long,
+            )
+        else:
```
pzelasko, line 85: This `else:` is a bug, you'd skip `return batch` when `speaker_id_mapping` is available.

lhotse/dataset/speech_synthesis.py
```diff
@@ -82,3 +92,10 @@ def validate_for_tts(cuts: CutSet) -> None:
         assert (
             len(cut.supervisions) == 1
         ), "Only the Cuts with single supervision are supported."
+
+
+def get_sid_to_index_map(sid_list) -> Dict[str, np.ndarray]:
```
pzelasko, line 97:
```suggestion
def get_sid_to_index_map(sid_list) -> Dict[str, int]:
```

lhotse/dataset/speech_synthesis.py
```diff
@@ -1,5 +1,6 @@
 from typing import Callable, Dict, List, Sequence, Union
 
+import numpy as np
```
pzelasko, line 3: This import won't be needed

JinZr:
we have decided to return the speaker str directly for flexibility, users could make use of the field outside dataloader upon their needs

pzelasko (approved):

PR #1221, opened 2023-11-24
save sdm files into a single mdm file to do gss

yuekaizhang:


desh2608:
I see what you're trying to get at, but I think there may be a more elegant solution which does not need running sox commands during data preparation. You can create a `Recording` object which has multiple `AudioSource`s, possibly 1 per microphone channel. Then, if you call `load_audio` with the specified channels, it would load only those channels. This is better in 2 ways:
1. data is not duplicated
2. you can load any combination of channels from the recording

As an example, I would suggest looking at the `mdm` preparation in ICSI: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/icsi.py

yuekaizhang:
> I see what you're trying to get at, but I think there may be a more elegant solution which does not need running sox commands during data preparation. You can create a `Recording` object which has multiple `AudioSource`s, possibly 1 per microphone channel. Then, if you call `load_audio` with the specified channels, it would load only those channels. This is better in 2 ways:
> 
> 1. data is not duplicated
> 2. you can load any combination of channels from the recording
> 
> As an example, I would suggest looking at the `mdm` preparation in ICSI: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/icsi.py

Done, thanks.

lhotse/recipes/icmcasr.py
```diff
@@ -6,22 +6,25 @@
 
 import logging
 import os
+import subprocess
```
desh2608, line 9: I believe this import is not needed anymore?

lhotse/recipes/icmcasr.py
```diff
@@ -6,22 +6,25 @@
...
 from lhotse.qa import fix_manifests, validate_recordings_and_supervisions
 from lhotse.recipes.utils import manifests_exist, normalize_text_alimeeting
 from lhotse.supervision import SupervisionSegment, SupervisionSet
 from lhotse.utils import Pathlike, is_module_available
 
 ICMCASR = ("train", "dev")  # TODO: Support all subsets when released
 POSITION = ("DA01", "DA02", "DA03", "DA04")
-SDM_POSITION = ("DX01C01", "DX02C01", "DX03C01", "DX04C01", "DX05C01", "DX06C01")
+# ignore "DX05C01", "DX06C01"
+SDM_POSITION = ("DX01C01", "DX02C01", "DX03C01", "DX04C01")
```
desh2608, line 27: Could you add a comment about why some of the SDM positions have been ignored? It would be useful for users of this recipe.

desh2608 (commented):
Some minor comments, otherwise LGTM.

lhotse/recipes/icmcasr.py
yuekaizhang: Done.

desh2608 (approved):

PR #1231, opened 2023-12-06
Allow lhotse installation without torchaudio for a limited set of features

pzelasko:


pzelasko:
I'm merging - the failed tests are flaky tests of SileroVAD workflow, maybe it's a temporary issue with downloading, if it persists, we'll deal with that separately as it's not affected by this PR.

PR #1232, opened 2023-12-07
Use `attacut` module for Thai word tokenization (in MMS forced alignment)

flyingleafe:
`tltk` module is not really maintained now, and it contains deprecated dependencies (`sklearn` vs. `scikit-learn`), which complicates its usage greatly. `attacut` module is fresher and has no problems with dependencies.


pzelasko (approved):
Thanks, LGTM!

PR #1242, opened 2023-12-21
Update mgb2.py

Shymaa2611:


pzelasko:
It looks like you accidentally committed some local changes.

PR #1243, opened 2023-12-21
updating the voxpopuli recipe

vesis84:
- allow to use pre-downloaded transcripts (so data praparation can be run without Internet access)
   - transcripts .tgz is downloaded into `/manifests`, and not into tmp
- search the audio data in folder: `corpus_dir / "raw_audios" / lang`


pzelasko (approved):
Thanks, Karel!

PR #1263, opened 2024-01-17
Added handlings for negative end time (#1203)

ArthLeu:
Formatted with Black this time.


desh2608:
Alignment times are supposed to be w.r.t. the recording, so I don't see how the start/end could be negative?

pzelasko:
@desh2608 Are we not adjusting alignment times together with supervision times when creating a cut? 

@ArthLeu could you add a unit test that would have failed before but will pass correctly now to cover your changes?

PR #1275, opened 2024-01-28
Install kaldi-native-io explicitly in the kaldi doc example.

csukuangfj:
Many users have forgotten to install `kaldi-native-io` when they convert features from kaldi to lhotse.
This PR puts an explicit command in the doc to mention that.


desh2608 (approved):

pzelasko:
Thanks!

PR #1278, opened 2024-01-31
Enable seed randomization in dynamic samplers

pzelasko:
This PR enables specifying `seed="randomized"` and `seed="trng"` for `DynamicCutSampler` and `DynamicBucketingSampler`. 

Both options are intended for use with `IterableDatasetWrapper` and cause the samplers to iterate with different random seeds in each node and dataloading worker. Note that for bucketing this will have the effect of de-synchronizing batch sizes across GPUs from the start of iteration (before the change, this occurs anyway after a number of training steps as observed in https://github.com/lhotse-speech/lhotse/discussions/857).

From now on, the sampler also attaches a custom field called `dataloading_info` to each cut which is a dict containing `rank`, `world_size`, and `worker_id` keys that help diagnose the dataloading.

### DIFF
--- a/lhotse/dataset/dataloading.py
+++ b/lhotse/dataset/dataloading.py
@@ -1,6 +1,10 @@
 import os
+import random
+import secrets
 from functools import partial
-from typing import Callable, Optional
+from typing import Callable, Literal, Optional, Union
+
+import torch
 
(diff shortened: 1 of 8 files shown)


pzelasko:
The failing test is flaky - merging

PR #1289, opened 2024-02-22
fix whisper for multi-channel data

yuekaizhang:


lhotse/features/whisper_fbank.py
```diff
@@ -55,7 +55,13 @@ def log_mel_spectrogram(
 
     if device is not None:
         audio = audio.to(device)
-    audio = audio.squeeze(0)
+
+    if len(audio.shape) == 2:
+        audio = audio[0]
+    assert (
```
pzelasko, line 61: With these changes if you pass a 2-channel recording, it will silently discard the second channel. I think it’s better to raise an error instead (ie only allow 1d array or 2d array where first dim is 1)

yuekaizhang: Fbank feature extractor seems use torchaudio.compliance.kaldi.fbank, where the input has to be mono. However, I can't find the place where we convert the [8, audio_len] into mono for aishell4.

yuekaizhang: Sorry for the late response, @pzelasko  I have added the warning for  your concern. Would you mind checking it, thanks.

pzelasko:
Thanks, LGTM!

PR #1292, opened 2024-02-29
Cutconcat fixed max duration

swigls:
Adding optional `max_duration` parameter to `CutConcatenate`.
It can be utilized when the ASR model consumes fixed-length audio inputs (e.g., Whisper).


pzelasko:
Cool contribution, thanks!

PR #1295, opened 2024-03-05
Extending Lhotse dataloading to text/multimodal data

pzelasko:
This PR adds a very basic support for incorporating text-only data into Lhotse samplers to enable text and multimodal dataloading. Highlights:
* new ABC `SamplingConstraint` that generalizes `TimeConstraint`, and allows to create other types of constraints to decide when to stop sampling a mini-batch as well as how to determine the "size" of an example (e.g. for audio its duration, but for text it may be sth like num tokens)
* dynamic samplers have a new argument called `constraint` where `SamplingConstraint` instances may be passed directly
* `TokenConstraint` which is almost identical to `TimeConstraint` but uses num_tokens / max_tokens
* very basic dataclass `TextExample` that wraps text/tokens, `CutSet` can be used to yield those (just pass text iterator to CutSet like `CutSet(text_example_iter)`) (it's not super clean but it works; trying to figure out if we can make this cleaner)
* unit tests illustrating how to use this for text dataloading and even for mixed modality dataloading (text and audio data together in a mini-batch)

This is stretching the original scope of Lhotse a bit, but I feel like it's worth it: we accumulated a bunch of solid techniques here and it'd be a pity to have to use something completely different for multimodal modeling. Would love to know your thoughts @danpovey @csukuangfj @desh2608 @m-wiesner

### DIFF
--- a/lhotse/custom.py
+++ b/lhotse/custom.py
@@ -1,0 +1,131 @@
+from functools import partial
+from typing import Any, Dict, Optional
+
+import numpy as np
+
+from lhotse import Recording
+from lhotse.utils import ifnone
+
+
+class CustomFieldMixin:
+    """
+    :class:`CustomFieldMixin` is intended for classes such as Cut or SupervisionSegment
+    that support holding custom, user-defined fields.
+
+    .. caution:: Due to the way inheritance and dataclasses work before Python 3.10,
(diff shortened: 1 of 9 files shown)


pzelasko:
Note that this is an experimental feature: let us know if you're running into issues with this.

PR #1306, opened 2024-03-18
Fixing recording move to memory

Tomiinek:
`move_to_memory` seems to fail for some cuts that have other audio soruce than `file`, no offset and duration matching the whole recording. I am fixing this for the `url` type, but probably still fails with the `command` or `shar` sources.

Also, I dont understand why there are `Case #1` and `Case #2` in this method  (I suppose case 1 is here to avoid additional decoding), especially when the user specifies a format that does not match the format of the original source. It looks like the case 1 condition should check also for the formats. 🤔 So feel free to fix it properly.

### DIFF
--- a/lhotse/audio/recording.py
+++ b/lhotse/audio/recording.py
@@ -295,10 +295,10 @@ class Recording:
             return x
 
         # Case #1: no opts specified, read audio without decoding and move it in memory.
-        if all(opt is None for opt in (channels, offset, duration)) or (
+        if format is None and (all(opt is None for opt in (channels, offset, duration)) or (
             (channels is None or _aslist(channels) == self.channel_ids)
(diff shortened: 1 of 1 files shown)


lhotse/audio/recording.py
```diff
@@ -294,19 +295,21 @@ def _aslist(x):
             return x
 
         # Case #1: no opts specified, read audio without decoding and move it in memory.
-        if all(opt is None for opt in (channels, offset, duration)) or (
+        if format is None and (all(opt is None for opt in (channels, offset, duration)) or (
```
pzelasko, line 298: `if all(opt is None for opt in (channels, offset, duration, format))` would be clearer

pzelasko (commented):
makes sense, pls fix the formatting and the other comment and LGTM

PR #1309, opened 2024-03-25
More similar mean batch duration across nodes with DynamicBucketingSampler in multi-GPU training

lifeiteng:
Update #863 

Fix #857

### DIFF
--- a/lhotse/dataset/sampling/dynamic_bucketing.py
+++ b/lhotse/dataset/sampling/dynamic_bucketing.py
@@ -1,4 +1,3 @@
-import logging
 import random
 import warnings
 from bisect import bisect_right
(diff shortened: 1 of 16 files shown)


pzelasko:
Does it actually help you achieve greater GPU utilization? Would be great if you can share any benchmarks. I wasn't able to verify the improvements in the original PR.

lifeiteng:
> Does it actually help you achieve greater GPU utilization? Would be great if you can share any benchmarks. I wasn't able to verify the improvements in the original PR.

It will be around 2x faster on 8xGPU.

pzelasko:
> It will be around 2x faster on 8xGPU.

Wow!! That's fantastic! Let me know if you need any help along the way.

lhotse/dataset/sampling/dynamic_bucketing.py
```diff
@@ -427,38 +448,51 @@ def is_ready(bucket: Deque[Cut]):
...
-                    for _ in range(batch_size):
-                        sampling_bucket.popleft()
+                # Shuffling, sort indexes of yielded elements largest -> smallest and remove them
+                indexes_used.sort(reverse=True)
+                for idx in indexes_used:
+                    del sampling_bucket[idx]
+
                 # Fetch new cuts and add them to appropriate buckets.
-                self._collect_cuts_in_buckets(batch_size)
+                self._collect_cuts_in_buckets(total_batch_size * self.world_size)
```
lifeiteng, line 495: we need collect more cuts here, It is often encountered that there is insufficient data in buckets.  @pzelasko Do you have any suggestion?

lifeiteng: sample batches from the some bucket(word_size times), then get batches[rank].

PR #1315, opened 2024-04-03
Enhance `CutSet.mix()` randomness and data utilization

pzelasko:
This PR is based on #1313 and #1314: huge thanks to @osadj and @kamirdin for your contributions! While I was reviewing the PRs I noticed some subtle errors and figured it will be quicker to just combine your work and fix it in one go.

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -3464,6 +3464,9 @@ class LazyCutMixer(Dillable):
     :param random_mix_offset: an optional bool.
         When ``True`` and the duration of the to be mixed in cut in longer than the original cut,
          select a random sub-region from the to be mixed in cut.
+    :param stateful: when True, each time this object is iterated we will shuffle the noise cuts
+        using a different random seed. This is useful when you often re-start the iteration and
+        don't want to keep seeing the same noise examples. Enabled by default.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


lhotse/cut/set.py
```diff
@@ -3557,10 +3559,9 @@ def noise_gen():
             # Keep sampling until we mixed in a "duration" amount of noise.
             # Note: we subtract 0.05s (50ms) from the target duration to avoid edge cases
             #       where we mix in some noise cut that effectively has 0 frames of features.
-            while mixed_in_duration < (
-                self.duration if self.duration is not None else cut.duration - 0.05
-            ):
+            while mixed_in_duration < target_mixed_duration:
                 to_mix = next(mix_in_cuts)
+                to_mix = self._maybe_truncate_cut(to_mix, target_mixed_duration, rng)
```
osadj, line 3564: @pzelasko I just noticed this. We need to update the target duration for random sub-region selection in the `while` loop (like I had in the PR). Let's assume the target is 20s. The first noise cut is 10s long, and we enter the `while` loop. The second noise cut is 15s long so we can select a random sub-region of 10s from the 15s, however because we use the original target duration we select from 0. Do you agree?

pzelasko: I think it makes sense. Can you make the change and verify it works as expected in another PR? Ideally add a unit test that would have been broken before the change is made, it will help maintaining this bit of code in the future.

osadj: Done #1317. Thanks.

PR #1316, opened 2024-04-03
Fix randomness in CutMix transform

pzelasko:


test/dataset/test_cut_transforms.py
```diff
@@ -118,6 +118,17 @@ def test_cutmix(preserve_id: bool):
         )
 
 
+def test_cut_mix_is_stateful():
+    speech_cuts = DummyManifest(CutSet, begin_id=0, end_id=10)
+    noise_cuts = DummyManifest(CutSet, begin_id=100, end_id=102)
+
+    # called twice on the same input, expecting different results
+    tfnm = CutMix(noise_cuts, snr=None, p=1.0, seed=0, preserve_id=True)
```
osadj, line 126: nit: tnfm

test/dataset/test_cut_transforms.py
```diff
@@ -118,6 +118,17 @@ def test_cutmix(preserve_id: bool):
...
+    out1 = tfnm(speech_cuts)
+    out2 = tfnm(speech_cuts)
+    assert list(out1) != list(out2)
```
osadj, line 129: Thanks for adding this test. I ran this using v1.17 which is stateless, but I still got different outputs (i.e., `out1` and `out2` are different). This is counterintuitive because `CutMix` passes the same seed in every `__call__` to `cuts.mix`. can you please help me understand this?

pzelasko: It used CutSet.sample which was not seeded and used the global RNG. If I start any new library from scratch I’ll keep it in mind to have a methodical approach to RNG handling from the beginning.

osadj: ah, you're right. I forgot about that. Thanks again.

PR #1317, opened 2024-04-05
select a random sub-region of the noise based on the delta duration

osadj:
For cut mixing, we select a random sub-region of the noise (if the noise duration supports) based on the delta duration needed to reach the target duration.

### DIFF
--- a/lhotse/cut/set.py
+++ b/lhotse/cut/set.py
@@ -3564,7 +3564,7 @@ class LazyCutMixer(Dillable):
             #       where we mix in some noise cut that effectively has 0 frames of features.
             while mixed_in_duration < target_mixed_duration:
                 to_mix = next(mix_in_cuts)
-                to_mix = self._maybe_truncate_cut(to_mix, target_mixed_duration, rng)
+                to_mix = self._maybe_truncate_cut(to_mix, target_mixed_duration - mixed_in_duration, rng)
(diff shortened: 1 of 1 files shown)


lhotse/cut/set.py
```diff
@@ -3562,9 +3562,15 @@ def noise_gen():
             # Keep sampling until we mixed in a "duration" amount of noise.
             # Note: we subtract 0.05s (50ms) from the target duration to avoid edge cases
             #       where we mix in some noise cut that effectively has 0 frames of features.
-            while mixed_in_duration < target_mixed_duration:
+            while mixed_in_duration < (
+                target_mixed_duration - 0.05
```
pzelasko, line 3566: why is the tolerance applied conditionally? shouldn’t it be unconditional here?

osadj: Thanks. You're right. I'm not sure if we need to subtract `50ms` though (i.e., `10ms` works, too).

pzelasko: Might have been an abundance of caution for feature extractors with different frame sizes and shifts. If you find it’s negatively affecting some experiments we can modify it otherwise I’d just keep it.

osadj: makes sense. It's updated accordingly.

pzelasko:
Thanks!

PR #1336, opened 2024-05-09
Missing 'subset' parameter

daniel-dona:
Just a missing parameter

Used here:
https://github.com/k2-fsa/icefall/blob/9d570870cff167b8b963ed1f73dae0d685ec53cf/egs/voxpopuli/ASR/prepare.sh#L84


pzelasko (approved):
Thanks!

PR #1343, opened 2024-05-28
Fix librispeech manifest caching

haerski:
Before, `prepare_librispeech` would always process manifests, regardless of whether or not the manifests were previously saved on disk. This fix passes the correct prefix to `manifests_exist`, which allows skipping.


pzelasko (approved):
Nice catch, thank you @haerski !

PR #1344, opened 2024-05-29
Add new sampler: weighted sampler

marcoyang1998:
Add a weighted sampler, where each cut's sampling probability is proportional to its weight. This is useful for unbalanced dataset, where some classes have very few data. The weight for each cut should be computed by the user and passed to the sampler. It's similar to pytorch's `WeightedRandomSampler` (see [here](https://pytorch.org/docs/stable/data.html#torch.utils.data.WeightedRandomSampler))

This sampler only works with eager manifest since we need to perform sampling globally.

### DIFF
--- a/lhotse/dataset/sampling/weighted_simple.py
+++ b/lhotse/dataset/sampling/weighted_simple.py
@@ -2,7 +2,7 @@ import warnings
 from typing import Any, Dict, List, Optional
 
 from lhotse import CutSet, Seconds
-from lhotse.dataset.sampling.base import CutSampler, TimeConstraint
+from lhotse.dataset.sampling.base import TimeConstraint
 from lhotse.dataset.sampling.data_source import WeightedDataSource
 from lhotse.dataset.sampling.simple import SimpleCutSampler
 
--- a/lhotse/dataset/sampling/weighted_simple.py
+++ b/lhotse/dataset/sampling/weighted_simple.py
@@ -70,10 +70,7 @@ class WeightedSimpleCutSampler(SimpleCutSampler):
             max_cuts=max_cuts,
             seed=seed,
         )
-        assert cuts.is_lazy == False, "This sampler does not support lazy mode!"
-        assert (
-            shuffle == False
-        ), "This sampler does not need shuffle as it performs sampling"
+        assert not cuts.is_lazy, "This sampler does not support lazy mode!"
         self.data_source = WeightedDataSource(
             cuts, weights=cuts_weight, num_samples=num_samples
         )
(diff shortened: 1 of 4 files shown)


pzelasko:
Thanks @marcoyang1998, I appreciate your work. I think you could achieve a similar outcome by splitting the cutset into subset cutsets for each class, and then using mux to get a cutset to be passed to any of the existing samplers. It would also work with lazy manifests and bucketing.

```python
class_cutsets = [cuts_class0, cuts_class1, ...]
class_weights = [w_class0, w_class1, ...]
cuts = CutSet.mux(*class_cutsets, weights=class_weights)
```

marcoyang1998:
Hi Piotr,

thanks for the `Cut.mux` example. I thought about this before as well, and it should work very well if the samples under the same class share an equal weight. However, in my setup, which is a multi-class classification (AudioSet), every sample has a unique sampling weight and using `Cut.mux` for this scenario is impractical.

lhotse/dataset/sampling/weighted_simple.py
```diff
@@ -0,0 +1,243 @@
+import warnings
+from typing import Any, Dict, List, Optional
+
+from lhotse import CutSet, Seconds
+from lhotse.dataset.sampling.base import CutSampler, TimeConstraint
+from lhotse.dataset.sampling.data_source import WeightedDataSource
+
+
+class WeightedSimpleCutSampler(CutSampler):
```
pzelasko, line 9: I wonder whether this class can inherit `SimpleCutSampler` and only override the necessary parts (e.g. using `WeightedDataSource` in `__init__`); otherwise most of this code looks very similar.

pzelasko (commented):
That's a valid point, I didn't think of this use-case. In that case LGTM, I just left one comment that may help us reduce code duplication.

lhotse/dataset/sampling/weighted_simple.py
marcoyang1998: Good idea! I just changed the inheritance, please have a look

lhotse/dataset/sampling/weighted_simple.py
```diff
@@ -181,63 +152,3 @@ def __iter__(self) -> "WeightedSimpleCutSampler":
             self.data_source.shuffle(self.seed + self.epoch)
```
pzelasko, line 152: is `if self.shuffle` branch needed here?

pzelasko (approved):
I made the changes I requested myself as I'd like to release a new version and include this. Thanks @marcoyang1998!

PR #1348, opened 2024-06-03
augmentation/torchaudio: add Phone effect (mulaw, lpc10 codecs)

rouseabout:
This patch adds a audio codec transformation.

I have found that when applying K2 ASR to speech compressed with mulaw, it is advantageous to augment the training data with these codecs. The transformation resamples the input audio to 8kHz, encodes then decodes using specified codec, then restores the original sample rate (e.g. 16 kHz).

Open issues:

- [ ] The transformation is called `phone()`. But maybe a better name is needed?
- [ ] Since it significantly alters the audio, depending on codec, I am wondering how best to test the transformation?

Example use:

```
cs2 = CutSet.from_manifests(...).phone(codec="mulaw")
cs3 = CutSet.from_manifests(...).phone(codec="lpc10")
```

libspandsp is required to use the lpc10 codec. Use apt-get install libspandsp-dev on Debian/Ubuntu.


lhotse/augmentation/torchaudio.py
```diff
@@ -212,6 +212,174 @@ def reverse_timestamps(
...
+
+
+CODECS = {
+    "lpc10": LPC10Codec,
+    "mulaw": MuLawCodec,
+}
+
+
+@dataclass
+class Phone(AudioTransform):
```
pzelasko, line 319: I suggest calling it `Narrowband` and renaming the methods to `narrowband`. Also upsampling back to original SR should be optional (`restore_orig_sr=True`).

lhotse/augmentation/torchaudio.py
```diff
@@ -212,6 +212,174 @@ def reverse_timestamps(
...
+    """
+    Phone effect.
+
+    Resample input audio to 8000 Hz, apply codec (encode then immediately decode), then resample back to the original sampling rate.
+    """
+
+    source_sampling_rate: int
```
pzelasko, line 326: We shouldn't need this option at all. You can call `get_or_create_resampler` directly in `__call__` using the input example's actual sampling rate. This way this transform can work with datasets of mixed sampling rates.

lhotse/augmentation/torchaudio.py
```diff
@@ -212,6 +212,174 @@ def reverse_timestamps(
         )
 
 
+class Codec:
```
pzelasko, line 215: The other codec classes don't actually inherit this base class.

lhotse/augmentation/torchaudio.py
```diff
@@ -212,6 +212,174 @@ def reverse_timestamps(
...
+        return self.decoder(self.encoder(samples))
+
+
+from ctypes import CDLL, POINTER, c_int, c_short, c_uint8, c_void_p
+
+LPC10_FRAME_SAMPLES = 180
+LPC10_FRAME_BYTES = 7
+
+
+def libspandsp_api():
```
pzelasko, line 242: Can we gracefully handle errors here by raising an exception with a clear message such as "We can't apply LPC10 codec because we couldn't locate library libspandsp.so. In order to use this augmentation, do ... "

pzelasko (commented):
Looks pretty good! I left a few comments. Could you also add unit tests for this transform?

rouseabout:
I have addressed everything except for `restore_orig_sr=True`. I am not sure how to achieve that!

pzelasko:
> I have addressed everything except for `restore_orig_sr=True`. I am not sure how to achieve that!

You are very close! Add a parameter `restore_orig_sr=True` in `def narrowband(self, ...)` for cut and recording, and pass the provided argument to `Narrowband` constructor. Then you can extend the condition for the second resampling to `if self.restore_orig_sr and sampling_rate != 8000)`.

rouseabout:
Done, but something extra is needed, because when I apply the transformation with `use_orig_sr=False` the following exception occurs: 

AudioLoadingError: The number of declared samples in the recording diverged from the one obtained when loading audio (offset=0, duration=19.22419501133787). This could be internal Lhotse's error or a faulty transform implementation. Please report this issue in Lhotse and show the following: diff=693887, **audio.shape=(1, 153900)**, recording=Recording(id='0_nb_lpc10', sources=[AudioSource(type='file', channels=[0], source='/home/user/workspace/rtvalid/0.wav')], sampling_rate=44100, **num_samples=847787**, duration=19.22419501133787, channel_ids=[0], transforms=[{'name': 'Narrowband', 'kwargs': {'codec': 'lpc10', 'restore_orig_sr': False}}])

pzelasko:
If you don't restore orig sr, you'll have to update both sampling_rate and num_samples property on the Recording object.

pzelasko:
Thanks for the contribution, merging!

PR #1355, opened 2024-06-11
Fix Recording.to_dict() when transforms are dicts and transform pickling issues

pzelasko:


pzelasko:
Merging despite one test failing - it's numpy 2.0 related, will fix another time.

PR #1361, opened 2024-06-24
Utils for discovering attached data and dropping in-memory data

pzelasko:
This PR adds the following methods and properties:
* `cut.iter_data()` which iterates over (key, manifest) pairs of all data items attached to a given cut (e.g., `("recording", Recording(...)), ("custom_features", TemporalArray(...))`)
* `is_in_memory` property for all manifest types to indicate if it contains data that is held in memory
* `is_placeholder` for non-cut manifests to indicate if a manifest is just a placeholder (has some metadata, but can't be used to load data)
* `cut.drop_in_memory_data()` which converts manifests with in-memory data to placeholders (this is useful for manifests that live longer than just dataloading to avoid blowing up CPU memory and/or slowing down the program)

### DIFF
--- a/lhotse/array.py
+++ b/lhotse/array.py
@@ -7,7 +7,7 @@ from typing import List, Optional, Union
 
 import numpy as np
 
-from lhotse.utils import Pathlike, Seconds, fastcopy, ifnone
+from lhotse.utils import Pathlike, Seconds, fastcopy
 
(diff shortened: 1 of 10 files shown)


pzelasko:
Skipping the last test that fails due to numpy 2.0 release

PR #1383, opened 2024-08-07
Added has_custom to MixedCut

anteju:
Currently, `MixedCut` is not resolving `has_custom` correctly.
This caused problems in `collate_audio` when using `MixedCut` with `recording_field != None`, which would fail here

https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/collation.py#L175

with something like
```
AttributeError: No such attribute: 'has_custom'
```

This PR adds `has_custom` to `MixedCut` and a brief unit test.


pzelasko (approved):
Great first contribution! Thanks!

pzelasko (approved):

PR #1392, opened 2024-09-14
Fix backend to None while ffmpeg is unavailable.

pengzhendong:
Before:

Screenshot 2024-09-14 at 18 34 15

After:

Screenshot 2024-09-14 at 18 35 08

### DIFF
--- a/lhotse/audio/backend.py
+++ b/lhotse/audio/backend.py
@@ -808,7 +808,8 @@ def torchaudio_info(
 
     if torchaudio_ffmpeg_backend_available():
         # Torchaudio 2.1 with official "ffmpeg" backend should solve all the special cases below.
-        info = torchaudio.info(path_or_fileobj, backend="ffmpeg")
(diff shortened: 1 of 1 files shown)


pzelasko:
Wait, that's weird. This code is guarded by `if torchaudio_ffmpeg_backend_available():` so it shouldn't have entered this branch at all. Perhaps you could move this extra logic inside `torchaudio_ffmpeg_backend_available` instead? Looks like the check there may be insufficient.

pengzhendong:
If it is moved to `torchaudio_ffmpeg_backend_available`, Lhotse will utilize `torchaudio.StreamReader` to read the audio instead of `torchaudio.info`.

| ImportError: Failed to initialize the FFmpeg extension. Tried versions: ['6', '5', '4', '']. Enabling DEBUG logging will provide more details about the error.

pzelasko:
I'm still not quite sure what's going on but your change seems unlikely to break anything and the backward compatibility tests are passing. LGTM, thanks!

pzelasko (approved):

PR #1393, opened 2024-09-17
Fix cli for ksponspeech

yfyeung:
An error occurs when using `lhotse prepare ksponspeech` due to the missing `normalize_text` parameter. Adding this parameter will fix the issue.


pzelasko (approved):
Thanks, LGTM!

PR #1398, opened 2024-09-30
Implement conversion from CutSet to HuggingFace dataset

domklement:
This PR implements a simple conversion from a CutSet containing MonoCuts and single-source Recording to a HuggingFace dataset.

`CutSet.to_huggingface_dataset: None -> DataSet`converts the cutset into one of two formats, depending on whether all the cuts contain only one supervision or multiple of them. The formats are described in the method's docstring.


lhotse/cut/set.py
```diff
@@ -2550,6 +2551,306 @@ def prefetch(self, buffer_size: int = 10) -> "CutSet":
...
+    def contains_only_mono_cuts(self) -> bool:
+        return all(isinstance(cut, MonoCut) for cut in self)
+
+    def has_one_supervision_per_cut(self) -> bool:
+        return all(len(cut.supervisions) == 1 for cut in self)
+
+    def has_one_audio_source(self) -> bool:
+        return all(len(cut.recording.sources) == 1 for cut in self)
+
+    def _convert_cuts_info_to_hf(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
```
pzelasko, line 2569: Since this file is pretty large and the functionality introduced is fairly well isolated, I suggest to change every method into a function (except the main HF export API) and move them into a separate file `lhotse/hf.py`. Keep the HF `datasets` imports local like they are. The body of `to_huggingface_dataset(self)` should import and call `def export_cuts_to_hf(cuts)` from that file.

lhotse/cut/set.py
```diff
@@ -2550,6 +2551,306 @@ def prefetch(self, buffer_size: int = 10) -> "CutSet":
...
+        has_speaker: bool,
+        has_language: bool,
+        alignment_types: Union[List[str], Set[str]],
+        exclude_attributes: Optional[Union[List[str], Set[str]]] = None,
+    ) -> Tuple[List[List[Dict[str, Any]]], Dict[str, Any]]:
+        """
+        Converts cut supervisions into a dictionary compatible with HuggingFace datasets format.
+
+        :param has_speaker: Whether the supervisions have speaker information.
+        :param has_language: Whether the supervisions have language information.
```
pzelasko, line 2604: Shouldn't these be auto-deducted?

pzelasko (changes requested):
Nice work!

lhotse/cut/set.py
domklement: Done.

lhotse/cut/set.py
```diff
@@ -2550,6 +2550,19 @@ def prefetch(self, buffer_size: int = 10) -> "CutSet":
             )
         )
 
+    def to_huggingface_dataset(self):
+        """
+        Converts a CutSet to a HuggingFace Dataset. Currently, only MonoCut with one recording source is supported.
+        Other cut types will be supported in the future.
+
+        More detailed description is in lhotse/hf.py
```
pzelasko, line 2558: One last comment: can you copy (duplicate) the documentation from `export_to_hf` here? It will be more user friendly for people using interactive suggestions (notebooks, IDEs)

domklement: That's a good point. It's done.

pzelasko (approved):

pzelasko:
Thanks!!!

PR #1400, opened 2024-10-10
Adds radio data recipe

m-wiesner:
This is the recipe for some radio data we collected and are able to distribute if someone emails us for it. We may possibly have a way of releasing it with some restrictions in the future, but for now it is just a prepare script. The paper it corresponds to is 

https://aclanthology.org/2024.naacl-long.286.pdf

The main use is for geolocating speech, self-supervised model pretraining, or language id.

### DIFF
--- a/lhotse/recipes/radio.py
+++ b/lhotse/recipes/radio.py
@@ -1,0 +1,144 @@
+"""
+This recipe prepares data collected from radio streamed on the web. The data
+have some metadata attached to them, including the geographic location of
+broadcast, date and time of the recorded clip, as well as a unique station
+identifier.
+
+Obtaining the data
+-----------------------------------------------------------
+If you want to use this corpus please email: wiesner@jhu.edu
+
+As the data are collected from radio stream, they cannot be broadly
+disseminated or used for commercial purposes. In the email, include your
+affiliated academic institution and the intended use for the data and we will
+the data to you if it is indeed for non-commercial, academic purporses.
+
+Description
+------------------------------------------------------------
+The data consist of ∼4000 hours of speech collected between
+September 27, 2023 to October 1, 2023, in 9449 locations all over the world,
+from 17171 stations. 
+
+These data were used for Geolocation of speech in order to answer the question,
+Where are you from? in the paper 
+
+Where are you from? Geolocating Speech and Applications to Language
+Identification, presented at NAACL 2024. Please read for a full descrption
+and please cite as 
+ 
+@inproceedings{foley2024you,
+  title={Where are you from? Geolocating Speech and Applications to Language Identification},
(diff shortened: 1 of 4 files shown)


lhotse/recipes/radio.py
```diff
@@ -0,0 +1,144 @@
...
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike
+import lhotse
+import re
+
+from functools import partial
+
+from lhotse.parallel import parallel_map
+from lhotse.audio import set_ffmpeg_torchaudio_info_enabled
+import torchaudio
```
pzelasko, line 52: Make torchaudio imports local, it’s not a core dependency anymore.

pzelasko (changes requested):
Thanks! Please fix the CI and the torchaudio import.

lhotse/recipes/radio.py
m-wiesner: I think it should all be fixed now. I have an associated recipe in icefall for doing LID and geolocation that I am going to submit a pull request for. Part of that involves a custom dataset class, but it's fairly specific to this one task. I assume it would be better to just have that as part of the icefall recipe and not part of Lhotse?

pzelasko: Yeah that makes sense.

lhotse/recipes/radio.py
```diff
@@ -0,0 +1,141 @@
...
+from typing import Dict, Optional, Union
+
+from tqdm import tqdm
+
+from lhotse.audio import Recording, RecordingSet, set_ffmpeg_torchaudio_info_enabled
+from lhotse.parallel import parallel_map
+from lhotse.supervision import SupervisionSegment, SupervisionSet
+from lhotse.utils import Pathlike
+
+set_ffmpeg_torchaudio_info_enabled(False)
```
pzelasko, line 50: One last request - don’t set these things in the global scope because it’ll be executed when Lhotse is imported. Move it to local scope (even better i think this one works like a context manager)

m-wiesner: I think I fixed it now?

pzelasko: Almost - now you need to move to actual usage to inside the function that does the audio reading and use it with Python with statement

m-wiesner: I actually forget why I started adding this in to all the Lhotse recipes. I think it fixed some problem I was having at some point, but I have no recollection of what that problem was. Do I actually need it? When I actually use it locally each time it gives me a ton of horrible logging messages.

pzelasko: You shouldn’t need it. We are using libsndfile now by default rather than torchaudio. Feel free to remove it if you prefer

m-wiesner: That is what I ended up doing. I believe the current updated commit just removed this.

m-wiesner:
Is there anything left to do for this one?

pzelasko:
Looks good, thanks!

PR #1427, opened 2024-11-28
Fix .m4a support in some setups (possibly for other formats not supported by libsndfile)

racoiaws:
Requesting to merge following changes

* Explicitly set m4a as unsupported in `lhotse.audio.backend.LibsndfileBackend` because it does not support m4a and will not support it in observable future. This is mostly a cosmetic change to eliminate a confusing exception when working with m4a files.  
  * https://github.com/bastibe/python-soundfile/issues/431#issuecomment-2068550691

* Set `AudioreadBackend.supports_info` to `True` because it has an `info` method just a line below. This is a bugfix making fallback to this backend actually work.

Real world impact: fixing `Recording.from_file` for m4a files in environments where `torchaudio` is not installed.
An example of such environment would be the `nvcr.io/nvidia/nemo:24.07` container.

### DIFF
--- a/lhotse/audio/backend.py
+++ b/lhotse/audio/backend.py
@@ -516,6 +516,10 @@ class LibsndfileBackend(AudioBackend):
         return False
 
     def is_applicable(self, path_or_fd: Union[Pathlike, FileObject]) -> bool:
+        if isinstance(path_or_fd, (Path, str)) and any(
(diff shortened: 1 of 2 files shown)


pzelasko (approved):
Thanks, LGTM!

pzelasko:
Don't worry about 3.10 and 3.11 tests - they are suffering from some GH action cache issue. If the other tests pass, we can merge it.

PR #1434, opened 2024-12-11
Optimized ReazonSpeech download speed using hf datasets features

yuta0306:
**Optimized download speed for ReazonSpeech**

The previous download codebase for ReazonSpeech is too lazy because it uses only one process to download, although we use HF datasets.

So, I optimized the codebase with HF datasets features:
* Added options for `num_proc`
* Disable decoding audio to reduce RAM and computational time

I assume downloading ReazonSpeech all (35k hrs audio) takes three days by the previous script, but we can download it in at least a day.


pzelasko (approved):
Thanks!!

pzelasko (approved):

PR #1441, opened 2024-12-24
allow more than one candidate backends

lifeiteng:


pzelasko:
Can you elaborate why is this change needed?

lifeiteng:
I want to try use different backends for opus: `export LHOTSE_LEGACY_OPUS_LOADING=1` will have two backends (`ffmpeg` + `soundfile`).

pzelasko:
Could you instead remove the condition in lines 514-515 in lhotse/audio/backend.py? 

https://github.com/lhotse-speech/lhotse/blob/509f4a97ab79a9eaeb8ba377d2a8e4acc794e45b/lhotse/audio/backend.py#L514-L515

lifeiteng:
> Could you instead remove the condition in lines 514-515 in lhotse/audio/backend.py?
> 
> https://github.com/lhotse-speech/lhotse/blob/509f4a97ab79a9eaeb8ba377d2a8e4acc794e45b/lhotse/audio/backend.py#L514-L515

done

pzelasko:
Thanks, I meant to also revert the change removing the check for unique backends - I think it is useful.

PR #1451, opened 2025-02-05
Add new augmentations: lowpass filtering, lossy compression (opus/mp3/vorbis)

racoiaws:
This PR adds two operations on `Cut`s/`Recording`s and corresponding randomized `CutSet` transforms
1) Lowpass filtering
2) In-memory lossy compression (with immediate decoding back to raw waveform)

Both would be useful for training robust ASR models, speech enhancement models, etc.

TODO
- [x] tests
- [x] docstrings for new methods in `lhotse.audio.recording.Recording`

### DIFF
--- a/lhotse/augmentation/lowpass.py
+++ b/lhotse/augmentation/lowpass.py
@@ -83,7 +83,9 @@ class Lowpass(AudioTransform):
         else:
             raise Exception(f"The lowpass filter {self.filter_type} is not supported!")
 
-        return scipy.signal.sosfiltfilt(sos, samples)
+        return scipy.signal.sosfiltfilt(
+            sos, samples
+        ).copy()  # copy because torch.from_numpy complains
 
     def reverse_timestamps(self, offset, duration, sampling_rate):
         return offset, duration
--- a/test/cut/test_cut_augmentation.py
+++ b/test/cut/test_cut_augmentation.py
@@ -928,6 +928,11 @@ def test_cut_lowpass(
     assert cut.num_samples == cut_lp.num_samples
     assert cut.load_audio().shape == cut_lp.load_audio().shape
 
+    assert (
+        torch.from_numpy(cut.load_audio()).shape
+        == torch.from_numpy(cut_lp.load_audio()).shape
+    )
+
 
 @pytest.mark.parametrize("codec", ["opus", "mp3", "vorbis"])
 @pytest.mark.parametrize("compression_level", [0.01, 0.5, 0.99])


lhotse/augmentation/codec.py
```diff
@@ -0,0 +1,60 @@
...
+                samples,
+                sampling_rate,
+                closefd=False,
+                compression_level=self.compression_level,
+                **self.prepare_sf_arguments(),
+            )
+            f.seek(0)
+            samples_compressed, rate_compressed = sf.read(
+                f, always_2d=True
+            )  # TODO: handle possible sample rate change with the opus codec?
```
pzelasko, line 44: When you write OPUS files with soundfile, it adds extra information in the file header about the original sampling rate and resamples from 48k before returning the audio array (unlike most other tools).

lhotse/augmentation/lowpass.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass
+from typing import Literal
+
+import numpy as np
+import scipy.signal
```
pzelasko, line 5: Since scipy is an optional lhotse dependency, please move this import inside __call__, and add an import guard (from `lhotse.utils`) before it:

```python
if not is_module_available("scipy"):
    raise ImportError("In order to use Lowpass transforms, run 'pip install scipy'")
```

lhotse/augmentation/lowpass.py
```diff
@@ -0,0 +1,37 @@
...
+import soundfile as sf
```
pzelasko, line 6: Can you also move all soundfile imports to local function scopes? IIRC importing this globally used to silently break documentation builds.

pzelasko (commented):
Great work! Please also add unit tests for calling transform methods on Recording and Cut, and for calling `cut_transforms` on CutSet

lhotse/audio/recording.py
```diff
@@ -921,6 +923,46 @@ def resample(self, sampling_rate: int) -> "Recording":
...
+        self, codec: Literal["opus", "mp3", "vorbis"], compression_level: float = 0.99
+    ) -> "Recording":
+        """
+        Return a new ``Recording`` that will lazily apply audio compression while loading audio.
+
+        :param codec: The codec to use for compression (one of: opus, mp3, vorbis).
+        :param compression_level: The compression level between 0.0 and 1.0 (higher means more compression).
+        :return: a modified copy of the current ``Recording``.
+        """
+        if codec not in ["opus", "mp3", "vorbis"]:
```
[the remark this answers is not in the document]
pzelasko: I know 3.8 had end of life recently but I'd like to keep the support for it for a little longer.

anhnami:
I would like to see packet loss augmentation added to Lhotse. This is slightly different from SpecAug because in this case the audio signal is masked in time domain. Packet loss happens frequently, results in choppy audio signals, and can slightly reduce ASR performance. Two frequent causes are internet problem or when you record audio via browser's API and browsers have to dropped audio chunks due to various reasons (Safary in particular). Currently we have to patch Nemo's Lhotse to add it to sampler which is very hard to maintain.

PR #1456, opened 2025-02-20
Add `features_dtype` arg to `collate_features` function

t13m:
, to allow user specify custom data type instead of `torch.float` in default. It's used for collate discrete code features (like in VQ codecs), which are always be in int types.


pzelasko (approved):
Thanks!

t13m:
Do I need to reformat the code to pass the black formatter check?

pzelasko:
Yes

pzelasko (approved):

pzelasko:
Thanks!

Issue #1457, opened 2025-02-21
Question about `validate_cut_set` in qa.py


t13m:
The validator `validate_cut_set` serves to make sure no duplicated ids in a `CutSet`. If I get it right, it is because the `CutSet` maintains a dict-like interface. My question is this: what should I do to duplicate some of the samples(cuts)? Will it work well if I simply comment out the whole `validate_cut_set` function and duplicate corresponding lines in `cutset.jsonl`?

In my experiment setting, the amount of data from different speakers varies over a wide range. In order to achieve the minimum level of balance, the data from speakers with fewer data need to be duplicated. Is there any more *lhotse way* to achieve that?

pzelasko:
That used to be a constraint but at some point we dropped it. I may have missed that validation still checks for this. If you could make a PR to remove this check it would be greatly appreciated. Thanks.

t13m:
Sure, will do it.

PR #1458, opened 2025-02-22
Remove `validate_cut_set` from qa.py

t13m:
issue #1457


pzelasko (changes requested):
Please keep the function, just remove the Counter thingy and the assertion relates to unique ids.

pzelasko (approved):
Thanks!

pzelasko (approved):

PR #1472, opened 2025-04-18
fix: add LHOTSE_MSC_BACKEND_FORCED flag to only enfore MSCIOBackend for non-MSC URLs

jayya2:
- Added `LHOTSE_MSC_BACKEND_FORCED `environment variable to force MSC for all URLs only when requested
- Modified is_applicable() to be more constrainted


pzelasko:
Thanks, LGTM! One of the tests failed on unrelated GH actions cache fetching - I will check it separately.

PR #1474, opened 2025-04-28
Handle edge case in .repeat() when the input is empty

pzelasko:


pzelasko:
Merging, the failing tests are unrelated and will be fixed separately.

PR #1483, opened 2025-05-21
Add image loading capabilities via Pillow

pzelasko:
Adds a tutorial with two examples how to set up manifests and data loading for:
* audio + image + text
* video + text


lifeiteng:
https://github.com/pytorch/torchcodec demonstrated good performance.

pzelasko:
Thanks for the tip, I might add it as a new backend.

PR #1485, opened 2025-06-05
minor fix: remove execute permission for all recipes

yfyeung:
Ensured consistent Linux permissions for all recipe files.


pzelasko (approved):
Thanks.

Issue #1459, opened 2025-03-01
Issue using Lhotse in NeMo - RuntimeError: cannot join current thread


LiamLonergan:
Hi there,

I've run into trouble using the lhotse dataloader in NeMo. My training script is failing straight away when initially trying to load the first batch for step 1. 

I am continuing the pretraining of the NEST-L model with ~9,000 hours of custom Irish unlabelled speech. I am training on a single machine with 80 cores, using 3x RTX 8000 Quadro (48Gb) and 128Gb system ram. 

I'm attaching the model config file and the output of a failed training attempt. Here is a sample of the error produced:
Exception in thread Thread-3 (producer):
Traceback (most recent call last):
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/site-packages/lhotse/dataset/sampling/dynamic_bucketing.py", line 744, in producer
    cuts = next(self.cuts_iter)
TypeError: 'NoneType' object is not an iterator
Exception ignored in: <function DynamicBucketer.__del__ at 0x7fca87762ef0>
Traceback (most recent call last):
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/site-packages/lhotse/dataset/sampling/dynamic_bucketing.py", line 784, in __del__
    self._producer_thread.join()
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/threading.py", line 1093, in join
    raise RuntimeError("cannot join current thread")
RuntimeError: cannot join current thread
Exception in thread Thread-3 (producer):
Traceback (most recent call last):
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/home/liam/miniconda3/envs/nemo/lib/python3.10/site-packages/lhotse/dataset/sampling/dynamic_bucketing.py", line 744, in producer
    cuts = next(self.cuts_iter)
TypeError: 'NoneType' object is not an iterator
[...]

PR #1487, opened 2025-06-16
fix cannot join current thread when drop_last=True

hoangtran9122:
#1459 
When using drop_last with concurrent_bucketing, sometimes when the data iterator starts, none of the buckets have enough data. Hence, the StopIteration() is raised outside the producer thread, which forces cuts_iter to None and makes the producer thread error (as in the first error, TypeError: 'NoneType' object is not an iterator).

Checking for the alive thread and joining it manually at the end of the iteration will solve the problem.


pzelasko:
Very interesting, thank you for debugging. LGTM

pzelasko:
Can you fix the failing checks?

hoangtran9122:
I fixed it, pls run the test again

pzelasko:
Thanks!

PR #1490, opened 2025-07-16
Add optional params to specify target split sizes in number of hours in ReazonSpeech recipe.

kinanmartin:
Adds optional parameters to specify target split sizes by number of hours in the ReazonSpeech prepare recipe.

The user can now optionally specify target hours by command line, like so:

```
lhotse prepare reazonspeech \
  -j $nj \
  --train-hours 100 \
  --dev-hours 2 \
  --test-hours 2 \
  $dl_dir/ReazonSpeech data/manifests
```

### DIFF
--- a/lhotse/bin/modes/recipes/reazonspeech.py
+++ b/lhotse/bin/modes/recipes/reazonspeech.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List
+from typing import List, Optional
 
 import click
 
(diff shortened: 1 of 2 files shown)


pzelasko:
What’s the rationale for this? Wouldn’t it cause different recipes to have different test and dev sets?

kinanmartin:
@pzelasko Thanks for the comment! 

The rationale is to allow our icefall model training recipe to create different split sizes depending on desired sizes for each split of the dataset.

We are currently working on a bilingual (English and Japanese) icefall recipe which relies on the data prepared via lhotse in the prepare script [here](https://github.com/reazon-research/icefall/blob/4ee8b26850ef80d999c1706cf56845604066f286/egs/reazonspeech/ASR/prepare.sh#L71-L80) as well as data from an English dataset ([see here](https://huggingface.co/datasets/parler-tts/mls_eng)). For the bilingual model, we want to be able to use the icefall recipes to prepare equally sized train, dev, and test sets for both datasets, then combine them to make a balanced dataset for the bilingual model, so we would like to be able to have control over the sizes of the splits instead of hardcoding them.

The way I have written the code here, if the new optional parameters are not specified, the train, dev, and test sets should all be generated identically to the current version.

PR #1502, opened 2025-08-12
Grid recipe fix

ialmajai:
- This is mainly to fix mis-assigned alignment data of different speaker from source.
- It also catches audio errors in some mpg files  
-


pzelasko:
Is this true? How did this the mis-alignment happen?

ialmajai:
> Is this true? How did this the mis-alignment happen?

Yes, it's very strange to see that many mistakes in the labels of the Grid dataset uploaded to Zenodo. The original data on the University of Sheffield's website doesn't have this problem though!

pzelasko (approved):
Thanks for fixing this!

PR #1504, opened 2025-08-14
Fix Windows test failure in test/test_utils.py caused by NamedTemporaryFile

### LINKED ISSUE #1503: Tests fail on Windows due to PermissionError with NamedTemporaryFile
I was running the tests on my Windows machine and noticed a failure in [test/test_utils.py](https://github.com/lhotse-speech/lhotse/blob/master/test/test_utils.py) caused by `NamedTemporaryFile` being opened twice, which is not allowed on Windows because the file is already open with an exclusive handle. It happend in the fixtures `safe_tar_file` and `unsafe_tar_file`.

Here’s the error I got:

```python
E           PermissionError: [Errno 13] Permission denied: 
E           'C:\\Users\\user\\AppData\\Local\\Temp\\tmpeqruskgx'

C:\hostedtoolcache\windows\Python\3.12.8\x64\Lib\gzip.py:192: PermissionError
```

This issue happens when the tests try to open a `NamedTemporaryFile` path again using `tarfile.open()` while the temporary file is still open.

On Linux this works fine, but on Windows it fails due to file locking behavior.

I’ll try to open a PR with a fix that makes it compatible with both systems.

denini08:
Fix #1503

On Windows, NamedTemporaryFile() keeps the file handle open with exclusive access, so when the tests try to open the same file again with tarfile.open(), it raises a PermissionError. This does not happen on Linux because files can be reopened even if they are already open.

Changes made:

- Updated safe_tar_file and unsafe_tar_file fixtures to use `NamedTemporaryFile(delete=False)`. This ensures the file handle is closed before reopening it with tarfile.open().
- Added manual file cleanup with os.remove() in a finally block to ensure temporary files are removed after use.

The fix is compatible with both Linux and Windows.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -1,4 +1,5 @@
 import tarfile
+import os
 from io import BytesIO
 from pathlib import Path
 from tempfile import NamedTemporaryFile, TemporaryDirectory
@@ -31,10 +32,14 @@ from lhotse.utils import (
 
 @pytest.fixture
 def safe_tar_file():
-    with NamedTemporaryFile() as f:
-        with tarfile.open(f.name, "w:gz") as tar:
+    with NamedTemporaryFile(delete=False) as f:
+        temp_name = f.name  # pega o caminho
+    try:
+        with tarfile.open(temp_name, "w:gz") as tar:
             tar.add("test/fixtures/audio.json")
-        yield f.name
+        yield temp_name
+    finally:
+        os.remove(temp_name)
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


test/test_utils.py
```diff
@@ -31,10 +32,14 @@
 
 @pytest.fixture
 def safe_tar_file():
-    with NamedTemporaryFile() as f:
-        with tarfile.open(f.name, "w:gz") as tar:
+    with NamedTemporaryFile(delete=False) as f:
+        temp_name = f.name  # pega o caminho
```
pzelasko, line 36: clean up the comment

pzelasko:
Thanks, please fix the failing checks and address the comment.

PR #1511, opened 2025-09-17
Add new augmentation: lowpass using back-and-forth resampling via libsox

racoiaws:
This PR adds:
- Resampling using `libsox` - in-memory resampling by calling `libsox` functions via `ctypes`
- `LowpassUsingResampling` transform with randomized cutoff frequency

To replace https://github.com/lhotse-speech/lhotse/pull/1451

I moved away from IIR filters from previous PR because they weren't steep enough and we need good rejection. I found resampling back-and-forth to be a very efficient and straightforward way to limit bandwidth. Also, with a little extra work we could use all of other SoX filters, not only `rate`

The `ctypes` adaptor is written with LLM assistance

Would really appreciate a fresh look on this

### DIFF
--- a/test/cut/test_custom_attrs.py
+++ b/test/cut/test_custom_attrs.py
@@ -507,7 +507,7 @@ def test_cut_resample_custom_recording(target_sampling_rate):
     )
 
     cut_resampled_only_recording_explicit = cut.resample(
-        target_sampling_rate, recording_field="recording"
+        target_sampling_rate, recording_field=None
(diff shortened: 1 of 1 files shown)


pzelasko (changes requested):
Thanks! I left some comments. I'd like to also see some unit test coverage of switching resampling backends, and some documentation behind *why* it would be preferable to use libsox backend (higher quality resampling).

PR #1512, opened 2025-09-17
Add new augmentation: soft/hard clipping

racoiaws:
This PR adds a `.clip_amplitude` operation on Cuts/Recordings and a correspond randomized CutSet transform

Either hard clipping (clamp to [-1, 1]) or soft saturation (tahn)


pzelasko (approved):
LGTM

PR #1517, opened 2025-09-27
Add recipe for NOTSOFAR-1

domklement:
This PR adds a recipe for [NOTSOFAR-1](https://huggingface.co/datasets/microsoft/NOTSOFAR) dataset available on HuggingFace. This dataset was introduced as a [https://www.chimechallenge.org/challenges/chime8/task2/index](CHiME-8 Task2) by Microsoft.

The recipe uses `HuggingFace Hub` to download the data (it is imported in the download function not to disrupt the original Lhotse requirements). The process function extracts recordings and supervisions with word-level alignments for both, single-channel and multi-channel data.


pzelasko (approved):
Great work, LGTM!

PR #1519, opened 2025-09-27
Restore punctuation in AMI recipe

Lakoc:
This PR restores punctuation in the AMI supervision text, ensuring it is preserved in the annotations. Users can still choose to remove punctuation later if desired. Additionally, this change removes a large number of unnecessary warnings that were previously produced, making logs cleaner and more informative.

### DIFF
--- a/lhotse/recipes/ami.py
+++ b/lhotse/recipes/ami.py
@@ -348,7 +348,7 @@ def parse_ami_annotations(
                     continue
                 start_time = float(word.attrib["starttime"])
                 end_time = float(word.attrib["endtime"])
-                words[key].append((start_time, end_time, word.text))
+                words[key].append((start_time, end_time,("" if word.get("punc", False) else " ") + word.text + ("- " if word.get("trunc", False) else "")))
 
     # Now we create segment-level annotations by combining the word-level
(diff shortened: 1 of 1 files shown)


lhotse/recipes/ami.py
```diff
@@ -348,7 +348,7 @@ def parse_ami_annotations(
                     continue
                 start_time = float(word.attrib["starttime"])
                 end_time = float(word.attrib["endtime"])
-                words[key].append((start_time, end_time, word.text))
+                words[key].append((start_time, end_time,("" if word.get("punc", False) else " ") + word.text + ("- " if word.get("trunc", False) else "")))
```
KarelVesely84, line 351: more readable would something like:
```
maybe_space = "" if word.get("punc", False) else " "
maybe_hyphen = "- " if word.get("trunc", False) else ""
words[key].append((start_time, end_time, maybe_space + word.text + maybe_hyphen))
```

KarelVesely84: btw, do you consider punctuation as a standalone token ?

pzelasko (approved):
Thanks, LGTM!

PR #1520, opened 2025-09-29
extend SimpleCutSampler to work better with CutConcatenate

KarelVesely84:
- by default batch size is computed an (num_cuts x longest_dur), however when cuts are concatenated the sholud be sum(utt_durs)
- keeping the default behavior as it was before

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -16,12 +16,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - python-version: "3.8"  # note: no torchaudio
-            torch-install-cmd: "pip install torch==1.12.1 --extra-index-url https://download.pytorch.org/whl/cpu"
-            extra_deps: ""
-          - python-version: "3.9"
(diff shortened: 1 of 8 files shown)


pzelasko:
But in your implementation it's currently not `sum(utt_durs)`, it's only measuring the length of the last cut. Please add unit tests to demonstrate the desired behavior of this option before we proceed.

KarelVesely84:
Please don't be mistaken, the `TimeConstraing.current` is accumulating the `sum(utt_durs)` : [sampling/base.py#L477](https://github.com/lhotse-speech/lhotse/blob/047081590b18a72225bce5ca3d8aacd46678d59b/lhotse/dataset/sampling/base.py#L477)
(I had same issue on 1st reading, 'current' is related to the current batch)

Should I rename the `current` to `current_batch` or similar ?

pzelasko:
Hi Karel,
Sorry for such a long delay, I should be able to respond quicker now. 
80% is quite arbitrary, can we remove 80% and tune `max_duration` instead?

PR #1523, opened 2025-10-02
fix: resolve AttributeError in DynamicBucketer

### LINKED ISSUE #1522: `AttributeError` in DynamicBucketer due to uninitialized `_source_exhausted` attribute
### Error and Reproduction Case

when training with NeMo using the `use_bucketing` option, I encountered an `AttributeError`:

```python
AttributeError: 'DynamicBucketer' object has no attribute '_source_exhausted'
  File "lhotse/dataset/sampling/dynamic_bucketing.py", line 767, in _maybe_wait_for_producer
    and not self._source_exhausted
```

I think this error occurs when:
- using `DynamicBucketingSampler` with `concurrent=True`
- in multi-prosess `DataLoader` environments (pytorch `DataLoader` with `num_workers > 0`)
- especially in distributed training scenario (DDP)
- when the main thread executes faster than the producer thread start

### Cause Analysis

the erorr is an initialization ordering bug:

1. initialization missing: the `_source_exhausted` attribute is not initializd in `DynamicBucketer.__init__()`

2. thraed execution sequence:
   ```
   main thread: _start_data_producer_thread() -> _maybe_wait_for_producer() -> accesses self._source_exhausted -> AttributeError()
   producer thread: (starts later) → sets self._source_exhausted = False
   ```
   the main thread accesses `_source_exhausted` before the producer thread has initiailzed it.

3. timing dependent: the producer thread initializes `_source_exhausted = False` inside its execution , but if the main thread reaches `_maybe_wait_for_producer()` before the os schdeuler run the producer thread, the attribute doesnt exist yet.

### Solution

add initialization in `__init__()` to ensure the attribute exists before any thread accesses it.

```python
class DynamicBucketer:
    def __init__(self, ...):
...
        self._producer_thread = None
        self._source_exhausted = False  # added
        ...
```

and reset it at the start of `__iter__()` for reusability.

```python
def __iter__(self):
    self.cuts_iter = iter(self.cuts)
    
    if self.concurrent:
        self._source_exhausted = False  # added
        self._start_data_producer_thread()
        self._maybe_wait_for_producer()
...
```

and remove the redundant initialization from inside the producer thread.

```python
    def _start_data_producer_thread(self):
        """Start concurrent filling of the bucket buffer in a background thread."""

        def producer():
            try:
                # self._source_exhausted = False # removed
                while not self._source_exhausted:
                    if sum(b.qsize() for b in self.buckets) == self.buffer_size:
...
```

### Summary

the `AttributeError` occurs due to initiailzation ordering issue where the main thraed attempts to access `_source_exhausted` before the producer thread has created it. the fix is to initialize the attributte in `__init__()` rather than inside the producer thread, ensuring it always exist regardless of thread scheduling timing.

I also noticed that using a plain boolean for `_source_exhausted` might have thread-safety issues (memory visibility). Would `threading.Event()` be a better approach here?

somniumism:
Fixes #1522

- Add `_source_exhausted` initialization in `__init__()`
- Reset flag in `__iter__()` for sampler reuse
- Remove redundant initialization from producer thread

Prevents AttributeError when main thread accesses the attribute before producer thread initializes it in multi-process DataLoader environments.

### DIFF
--- a/lhotse/dataset/sampling/dynamic_bucketing.py
+++ b/lhotse/dataset/sampling/dynamic_bucketing.py
@@ -576,12 +576,14 @@ class DynamicBucketer:
         self.buckets: List[Queue] = [Queue() for _ in range(len(duration_bins) + 1)]
 
         self._producer_thread = None
+        self._source_exhausted = False
 
     def __iter__(self) -> Generator[CutSet, None, None]:
         # Init: sample `buffer_size` cuts and assign them to the right buckets.
         self.cuts_iter = iter(self.cuts)
 
         if self.concurrent:
+            self._source_exhausted = False
             self._start_data_producer_thread()
             self._maybe_wait_for_producer()
         else:
@@ -743,7 +745,6 @@ class DynamicBucketer:
 
         def producer():
             try:
-                self._source_exhausted = False
                 while not self._source_exhausted:
                     if sum(b.qsize() for b in self.buckets) == self.buffer_size:
                         time.sleep(0.1)


pzelasko (approved):

pzelasko:
Thanks! You keep finding these edge cases that somehow I never run into - what platform/env do you use for training? Maybe I can extend the test envs somehow to be able to catch that too.

PR #1527, opened 2025-10-10
avoid bug appearing with OnTheFlyFeatures and PerturbVolume cut_transform

KarelVesely84:
the error appeared when training on 4 GPUs with 6 workers per GPU

```
File "/mnt/matylda5/iveselyk/ASR_TOOLKITS/K2_SHERPA_PYTORCH24_CUDA121/K2_CONDA_ENVIRONMENT/lib/python3.11/multiprocessing/reduction.py", line 60, in dump
ForkingPickler(file, protocol).dump(obj)
TypeError: cannot pickle 'module' object
```

the `PerturbVolume.random` should not have the module `random` assigned: https://github.com/lhotse-speech/lhotse/blob/a509b4ad9e3c997c08b6f0d41086a14109d0ac81/lhotse/dataset/cut_transforms/perturb_volume.py#L31

assigning the object `random.Random()` was fine, the error disappeared.

the custom `_uniform()` method was implemented.

### DIFF
--- a/lhotse/dataset/cut_transforms/perturb_volume.py
+++ b/lhotse/dataset/cut_transforms/perturb_volume.py
@@ -28,13 +28,16 @@ class PerturbVolume:
 
     def __call__(self, cuts: CutSet) -> CutSet:
         if self.random is None:
-            self.random = random
+            self.random = random.Random()
         return CutSet.from_cuts(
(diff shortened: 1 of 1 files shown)


pzelasko:
Thanks!

PR #1530, opened 2025-10-30
fix: Add Image support to Cut.iter_data() method

gaikwadabhishek:
The iter_data() method was not yielding Image objects from cut.custom, only Recording, Features, Array, and TemporalArray. This prevented batch loaders and other utilities from discovering and processing Image manifests attached to cuts.

- Add Image to isinstance() check in DataCut.iter_data()
- Update type annotations to include Image in return type
- Add Image import to data.py and mixed.py


pzelasko (approved):
Thanks, sharp eyes!

PR #1542, opened 2026-01-13
refactor(ais): Add backward compatibility and improve error handling in AISBatchLoader

gaikwadabhishek:
- Add backward compatibility for older AIStore SDK versions (Colocation, ArchiveConfig)
- Move all aistore imports to method level (remove top-level imports)
- Replace module-level logging with logger instance for better configuration
- Fix return type annotation for _collect_manifest_urls (None -> bool)
- Add safe attribute access with getattr() in error messages
- Simplify ValueError handling with early return

Tests:
- Add version compatibility tests for Colocation and ArchiveConfig fallbacks

This improves SDK version compatibility and code robustness while maintaining
full backward compatibility with older AIStore deployments.

Signed-off-by: Abhishek Gaikwad <gaikwadabhishek1997@gmail.com>

### DIFF
--- a/lhotse/ais/batch_loader.py
+++ b/lhotse/ais/batch_loader.py
@@ -1,4 +1,10 @@
-from typing import Any, Dict, Iterator, Optional, Tuple
+import logging
+from typing import Any, Optional
+
+from urllib3.exceptions import TimeoutError
+
+# Get a logger instance for this module
+logger = logging.getLogger(__name__)
 
 from lhotse.array import Array, TemporalArray
 from lhotse.audio.recording import Recording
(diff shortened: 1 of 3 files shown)


gaikwadabhishek:
added

lhotse/ais/batch_loader.py
```diff
@@ -72,20 +116,102 @@ def __call__(self, cuts: CutSet) -> CutSet:
...
+                        ) from e
+
+                # Retry with direct API call if content is empty (from timeout or actual empty response)
+                if content == b"":
+                    logger.warning(
+                        f"Object {info.obj_name}/{info.archpath} from bucket {info.provider}://{info.bck} "
+                        f"returned empty content. Retrying with direct AIStore API call."
+                    )
+                    try:
+                        content = self._get_object_from_moss_in(info)
```
pzelasko: Why run individual GET every time content is empty? Can we read the returned status to determine if there was an issue, or it was OK (in which case we don't send GET again)?

PR #1558, opened 2026-03-11
Support object stores: `CutSet.from_shar(in_dir='s3://...')`

pzelasko:


lhotse/utils.py
```diff
@@ -138,6 +138,59 @@ def is_valid_url(value: str) -> bool:
...
+        if _is_top_level_object_store_key(key, listing_prefix)
+    )
+
+
+def _is_top_level_object_store_key(key: str, listing_prefix: str) -> bool:
+    if listing_prefix:
+        if not key.startswith(listing_prefix):
+            return False
+        key = key[len(listing_prefix) :]
+    return bool(key) and "/" not in key
```
karpnv: Why do we need this limitation `"/" not in key` ?

PR #1571, opened 2026-06-13
Add .copy_with(**kwargs) convenience method to manifests (#242)

### LINKED ISSUE #242: Refactor fastcopy() into a member method .copy_with()
Although Lhotse does not enforce immutability for manifest objects, it (almost?) always treats them as if they were immutable. In many places, we have this idiom e.g. `fastcopy(cut, features=Features(...))` to create a modified version of an object without actually modifying it. I think it makes sense to add a method `.copy_with(**kwargs)` to such classes, so it's a bit more convenient to use (doesn't require an import and works nicely with Python's comprehensions). E.g. `cut.copy_with(start=5.0)`

deekshaNVIDIA:
## Summary

Closes #242.

Lhotse treats manifest objects as (effectively) immutable and uses the `fastcopy(obj, field=...)` idiom all over the codebase to produce a modified copy. As suggested in #242, this adds a `.copy_with(**kwargs)` member method so users can do the same without importing `fastcopy`, in a way that reads naturally and composes nicely with comprehensions:

```python
sup2 = supervision.copy_with(text="new text")
cut2 = cut.copy_with(start=5.0)
rec2 = recording.copy_with(id="new-id")
```

## What changed

`copy_with(**kwargs)` was added to the core manifest types:

- **`CustomFieldMixin`** (`lhotse/custom.py`) — covers `SupervisionSegment`, `DataCut`/`MonoCut`/`MultiCut`, and the text examples.
- **`Cut`** base class (`lhotse/cut/base.py`) — covers `PaddingCut` and `MixedCut`; delegates to the existing `Cut.copy()` so cuts expose the same name as every other manifest (this is exactly the `cut.copy_with(start=5.0)` example from the issue).
- **`Recording`**, **`Features`**, **`Array`**, **`TemporalArray`** — thin wrappers around `fastcopy`.

All implementations delegate to `fastcopy`, so behavior/semantics are identical to the existing idiom.

## Tests

`test/test_copy_with.py` parametrizes over every manifest type and verifies that `copy_with`:
- overwrites the requested field,
- returns a new object and does **not** mutate the original,
- matches `fastcopy(obj, **kwargs)`,
- is a no-op-equivalent when called with no kwargs,
- and can set custom fields on `SupervisionSegment`.

`37 passed` locally. I also ran `test/test_supervision_set.py`, `test/cut/test_padding_cut.py`, and `test/features/test_temporal_array.py` with no regressions (one unrelated failure, `test_serialize_padded_cut_set`, is a Windows-only `NamedTemporaryFile` reopen quirk). `black`, `isort --profile black`, and `flake8 --select=E9,F63,F7,F82` are clean on the changed files.


pzelasko (approved):
Thanks!

PR #1578, opened 2026-08-22
Fix HiFiTTS partition association during parallel preparation

marcus-campbell:
## Description

`prepare_hifitts()` paired `as_completed(futures)` with submission-ordered partition IDs using `zip()`. Because `as_completed()` yields futures in completion order and *not* submission order, *a completed partition could be stored and written under another partition’s identity*. **Users whose HiFiTTS partitions completed in a different order may have manifests stored under the wrong partition ID**; parallel runs are the most likely to be affected.

The change maps each submitted future directly to its partition ID and recovers that ID when the future completes. `as_completed()` does not preserve submission order.

## What Changed

- Preserve each HiFiTTS partition’s identity while preparing partitions in parallel.
- Add a regression test that deliberately reverses future completion order.
- Verify both returned manifest keys and output filenames use the correct partition ID.

## Testing

- Regression test fails on `master` with `AssertionError: assert '6097_other_dev' == '92_clean_train'` and passes with this fix.
- `pytest -q test/recipes`: 2 passed
- Black, isort, and flake8 passed for the changed files.
- `git diff --check` passed.


test/recipes/test_hifitts.py
```diff
@@ -0,0 +1,69 @@
+import lhotse.recipes.hifitts as hifitts
```
pzelasko: Remove the unit test file and LGTM

pzelasko (changes requested):
Thanks, one change requested.

test/recipes/test_hifitts.py
marcus-campbell: Removed the unit test file as requested. Thank you!

pzelasko (approved):

pzelasko:
Please fix the formatter, don't worry about codecov

marcus-campbell:
> Please fix the formatter, don't worry about codecov

Thanks! I updated the branch from `master`, and `Black 22.3.0` now passes locally. The new CI runs are awaiting approval.

pzelasko (approved):

PR #1579, opened 2026-08-24
Fix heroico recipe: crash and empty devtest transcripts

ytinyui:
- Fix a bug that causes the crash
- Fix empty transcripts in devtest fold caused by incorrect transcript id


pzelasko (approved):
Thanks!

PR #1581, opened 2026-08-26
Speedup AISBatchLoader construction using new Batch.add() string API

msoham123:
`AISBatchLoader._append_moss_in` previously bypassed Pydantic validation using `MossIn.model_construct` and directly mutated `batch.request.moss_in` to avoid per-entry `Bucket`/`Object` construction overhead. This relied on private SDK internals and was fragile across SDK versions.

aistore 1.26.0 introduces `Batch.add(str, bck=, provider=)` as well as many other performance optimizations (see commits [dc0cf9b](https://github.com/NVIDIA/aistore/commit/dc0cf9bd8cd89072ea1c3d46dd436d6fbe5c4d5e) and [1391d0b](https://github.com/NVIDIA/aistore/commit/1391d0b96b716e60284405b8d03ec1865498b5fd)), accepting object names as plain strings and skipping intermediate wrapper construction at the SDK level. Benchmarking shows that the new SDK validating constructor is **8.5x faster** than model_construct (0.98 µs vs 8.32 µs/entry) and end-to-end batch construction is **2.4x faster** with the new SDK path (0.30ms vs 0.72ms median, 32-cut batch).

`_aistore_byte_range_supported` is also simplified to a plain version gate since the internal schema probes it performed were only needed to validate the bypass.

Signed-off-by: Soham Manoli <msoham123@gmail.com>

### DIFF
--- a/lhotse/ais/batch_loader.py
+++ b/lhotse/ais/batch_loader.py
@@ -111,7 +111,7 @@ class AISBatchLoader:
         if self._client is None:
             if not is_module_available("aistore"):
                 raise ImportError(
-                    "Please run 'pip install aistore>=1.17.0' to use AISBatchLoader."
+                    "Please run 'pip install aistore>=1.26.0' to use AISBatchLoader."
                 )
             self._client, _ = get_aistore_client()
         return self._client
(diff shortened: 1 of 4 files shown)


msoham123:
Fixed black formatter warnings

PR #1582, opened 2026-08-31
Fix both resumable_download calls in the CHiME-6 recipe raising TypeError

Anai-Guo:
Both `resumable_download` call sites in `lhotse/recipes/chime6.py` pass keyword
names the callee does not define, so each raises `TypeError` before a single
byte is downloaded.

`resumable_download` (`lhotse/utils.py:471`):

```python
def resumable_download(
    url: str,
    filename: Pathlike,
    force_download: bool = False,
    completed_file_size: Optional[int] = None,
    missing_ok: bool = False,
    ssl_context: Optional[SSLContext] = None,
    additional_headers: Optional[Dict[str, str]] = None,
) -> None:
```

There is no `**kwargs`, so both spellings below are hard errors:

| Call site | Passes | Result |
|---|---|---|
| `chime6.py:115` (`download_chime6`) | `request_ssl_context=` | `TypeError: got an unexpected keyword argument 'request_ssl_context'` |
| `chime6.py:427` (`check_md5`) | `desc=` | `TypeError: got an unexpected keyword argument 'desc'` |

Reproduced by binding the real upstream signature (extracted from
`lhotse/utils.py` on `master`) against each call site, with two correctly-spelled
sibling calls as a control group:

```
resumable_download (url, filename, force_download, completed_file_size, missing_ok, ssl_context, additional_headers)
  TypeError chime6.py:115 CHiME-6 tarball download -> got an unexpected keyword argument 'request_ssl_context'
  TypeError chime6.py:427 checksum-file download   -> got an unexpected keyword argument 'desc'
  OK        CONTROL librispeech.py style call
  OK        CONTROL ssl_context spelled correctly
```

Of the **70** `resumable_download` call sites in the repository, these two are the
only ones using these spellings — every other call already uses the real
parameter names.

This means `download_chime6` cannot download the CHiME-6 tarballs (the
`unverified_ssl_ctx` built immediately above is never actually applied), and
`check_md5` cannot fetch the checksum file.

The fix renames `request_ssl_context=` to `ssl_context=`, which preserves the
evident intent — the unverified SSL context now genuinely reaches the
downloader — and drops the unsupported `desc=` argument.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lhotse/recipes/chime6.py
+++ b/lhotse/recipes/chime6.py
@@ -112,7 +112,7 @@ def download_chime6(
             f"{url}/{file_name}",
             filename=tar_path,
             force_download=force_download,
-            request_ssl_context=unverified_ssl_ctx,
+            ssl_context=unverified_ssl_ctx,
         )
         with tarfile.open(tar_path) as tar:
             safe_extract(tar, path=target_dir)
@@ -424,7 +424,7 @@ def _verify_md5_checksums(
     temp_dir = Path(tempfile.mkdtemp())
     checksum_file = temp_dir / "md5sums.txt"
     resumable_download(
-        CHIME6_MD5SUM_FILE, str(checksum_file), desc="Downloading checksum file"
+        CHIME6_MD5SUM_FILE, str(checksum_file)
     )
     checksums = {}
     with open(checksum_file, "r") as f:


pzelasko (approved):
Thanks!

PR #1589, opened 2026-09-13
Add opt-in compact buffered-token checkpoints for DynamicBucketingSampler

albertz:
Indexed sampler snapshots can contain tens of thousands of small lists and nested graph tokens.
Reconstructing these objects during worker-to-parent snapshot transport creates substantial allocation and garbage-collection overhead.

This adds `DynamicBucketingSampler(compact_state=True)` to capture buffered tokens directly as immutable bytes, avoiding construction of the intermediate snapshot lists.
The implementation streams existing tokens through the C-backed standard-library pickler with binary structural counts.

- Defaults to `False`, preserving existing behavior.
- Supports variable token shapes and nested built-in values.
- Loads both legacy and compact states, regardless of the sampler's current setting.
- Expands compact tokens for JSON export, following the known `samplers` and `bucket_samplers` composition fields.
- Preserves the generic JSON serializer's behavior for unrelated bytes and byte-valued tokens, without mutating the original snapshot.
- Rejects custom reconstruction functions when encoding, and globals and extension opcodes when decoding, including cached extensions.

Older Lhotse versions cannot load compact binary states; use the default representation when backward loading is required.

Validation includes 170 passing CPU tests covering indexed continuation, legacy/compact conversion, snapshot independence, malformed payloads, concurrent capture, and real StatefulDataLoader worker checkpoint/restore.
Regression cases cover all three pickle extension opcodes with empty and populated caches, nested RoundRobinSampler/ZipSampler JSON export, and preservation of unrelated metadata.
The changed files also pass the repository's pinned Black, isort, and required Flake8 checks.

A synthetic benchmark compares ordinary snapshots, serialization after capture, and direct compact capture.
Against serialization after capture, direct capture uses approximately 4.8× less temporary Python memory and eliminates capture-time GC collections, with similar median capture time (36.8 ms versus 37.6 ms in this run) and a 49% larger payload.
These are CPU snapshot measurements; no additional training-throughput improvement is claimed.

Decoding uses Python's `pickle._Unpickler` implementation so extension lookups can be rejected before consulting the process-wide cache.
In a matched comparison using the same 30,000-item payload, decoding took about 231 ms versus 69 ms with the C decoder.
This extra cost occurs when expanding tokens for restore or JSON export; snapshot capture and transport retain C-backed pickle.

### DIFF
--- a/lhotse/checkpoint.py
+++ b/lhotse/checkpoint.py
@@ -161,9 +161,21 @@ class DataloaderCheckpoint:
 
     def save(self, path: Pathlike) -> None:
         """Serialize the checkpoint to a JSON file."""
+        data = asdict(self)
+        sampler_state = data["sampler_state"]
+        if isinstance(sampler_state, dict):
+            bucketer_state = sampler_state.get("bucketer_state")
+            if isinstance(bucketer_state, dict) and isinstance(
+                bucketer_state.get("bucket_tokens"), bytes
+            ):
+                from lhotse.dataset.sampling.token_codec import unpack_bucket_tokens
+
+                bucketer_state["bucket_tokens"] = unpack_bucket_tokens(
+                    bucketer_state["bucket_tokens"]
+                )
         path = Path(path)
         with open(path, "w") as f:
-            json.dump(asdict(self), f, indent=2, default=_json_serializer)
+            json.dump(data, f, indent=2, default=_json_serializer)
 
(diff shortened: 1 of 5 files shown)


danpovey:
Albert, you didn't let me know, did you get a job yet?  Respond to my email

pzelasko (approved):

